PR #6, opened 2018-07-21
toolbox: removed a call to the deprecated matplotlib prctile

balarsen:
Testing the pull request workflow. There was a deprecated call in toolbox anyway. 
`matplotlib.mlab.prctile` -> `numpy.percentile`
Not really sure why we used the matplotlib version in the first place. Tests all pass with numpy.


drsteve:
I'm pretty sure we used matplotlib's mlab because there was no numpy equivalent (in 2009/2010).
Can you add a tracker to check and document the version of numpy that we actually require?

Issue #12, opened 2018-07-28
dmarray breaks on masking [Migrated from SF; Bugs #65 and #71]


drsteve:
Combining two trackers from SourceForge:
[#65 dmarrays do not properly convert to masked arrays](https://sourceforge.net/p/spacepy/bugs/65/)

[#71 dmarray array behavior difference in np.median](https://sourceforge.net/p/spacepy/bugs/71/)

Basic summary of these two bug trackers is that dmarray is a subclass of numpy ndarray. A straightforward subclass of ndarray (e.g. following their RealisticInfoArray example) is that ufuncs like max, min, etc. behave inconsistently between numpy arrays and subclasses of arrays. The fix that was committed for this pushed the error further into hiding and manifested as a breakage on conversion to a MaskedArray. One major impact was that dmarrays needed to be converted to ndarrays before being passed to matplotlib's pcolormesh, but the issue was basically anywhere that masking was needed.

drsteve:
Candidate fix for this issue is in the new_dmarray branch.

The fix subclasses MaskedArray, with some modifications to preserve existing dmarray behaviour.
So far (since TravisCI isn't hooked up yet) the new branch passes all of our tests, and tests of some downstream packages that rely heavily on dmarray also work. Several of the @spacepy/councilofelrond have also tested this branch on some dmarray-dependent analyses with no noted issues.

spacepy.poppy tests currently don't use dmarrays (need to check internals to see whether there's any use inside the module). Assigning myself this issue with task list:
- Test PoPPy with dmarray inputs.

Assuming spacepy.poppy test suite passes then I'll do a pull request. If anyone has any objections or additional testing requirements then please leave them here and we can address before merge.

drsteve:
To update and clarify my previous comment:
- I have modified a few lines in the pycdf testing that test the output of xx.flat against yy.flat, where xx and yy may have different types. Because flat (ndarray/maskedarray) returns an iterator it depends on the subclass. Changing these to flatten() then compares objects and not iterators, so the tests compare the values and pass.
- One CDF test fails (in 2.7 and 3.6, seems to pass in 3.5??). The root cause appears to be in the "flat" method again. Specifically, the test for well-formed input data (staticmethod: check_well_formed) tests a slice from the flat iterator. Changing this to use flatten() passes all tests. My guess is that we should revisit check_well_formed as it's a bit klugey. I'm committing that change in the new_dmarray branch so that everything passes the tests.
[...]

PR #13, opened 2018-07-30
Reimplement dmarray and add continuous integration

drsteve:
Major changes:
- Added .travis.yml file to set up Travis CI and build/run test suite
- Reimplemented dmarray so it's a subclass of numpy.ma.MaskedArray, not of numpy.ndarray
- pycdf update to not compare or slice iterators in testing and check_well_formed method
- Updated included leapsecond file
- fix to test_time.py: test_randomDate regression test as expected result differs between 2.7 and 3.6

dmarray modifications fix issue #12  

[Test results for my branch](https://travis-ci.com/drsteve/spacepy)
Only 3.5 and 3.6 are currently tested here as the 2.7 build was freezing (on Travis) on compiling C sources. All tests on 2.7 pass on my local machine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,5 +18,5 @@ install:
  - wget https://spdf.sci.gsfc.nasa.gov/pub/software/cdf/dist/cdf37_0/linux/cdf37_0-dist-cdf.tar.gz; tar xzvf cdf37_0-dist-cdf.tar.gz; cd cdf37_0-dist; make OS=linux ENV=gnu all; make install; cd ..
 
 script:
- - sudo python setup.py install
+ - python setup.py install
  - cd tests; python test_spacepy.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ install:
  - pip install --upgrade h5py
  - pip install --upgrade networkx
  - pip install --upgrade ffnet
- - wget https://spdf.sci.gsfc.nasa.gov/pub/software/cdf/dist/cdf37_0/linux/cdf37_0-dist-cdf.tar.gz; tar xzvf cdf37_0-dist-cdf.tar.gz; cd cdf37_0-dist-cdf; make OS=linux ENV=gnu all; make install
+ - wget https://spdf.sci.gsfc.nasa.gov/pub/software/cdf/dist/cdf37_0/linux/cdf37_0-dist-cdf.tar.gz; tar xzvf cdf37_0-dist-cdf.tar.gz; cd cdf37_0-dist; make OS=linux ENV=gnu all; make install
 
 script:
  - sudo python setup.py install
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+language: python
+python:
+ - "2.7"
+ - "3.5"
+ - "3.6"
+
+before_install:
+ - sudo apt-get update -qq
+ - sudo apt-get install libhdf5-serial-dev gcc gfortran
+
+install:
+ - pip install --upgrade numpy
+ - pip install --upgrade scipy
+ - pip install --upgrade matplotlib
+ - pip install --upgrade h5py
+ - pip install --upgrade networkx
+ - pip install --upgrade ffnet
+ - wget https://spdf.sci.gsfc.nasa.gov/pub/software/cdf/dist/cdf37_0/linux/cdf37_0-dist-cdf.tar.gz; tar xzvf cdf37_0-dist-cdf.tar.gz; cd cdf37_0-dist-cdf; make OS=linux ENV=gnu all; make install
+
+script:
+ - sudo python setup.py install
+ - cd tests; python test_spacepy.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 python:
- - "2.7"
- - "3.5"
+ - "3.5" #temporarily removed 2.7 as the build was stalling when compiling sources...
  - "3.6"
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,4 @@ install:
 
 script:
  - python setup.py install
- - cd tests; export CDF_LIB=$HOME; python test_spacepy.py
+ - cd tests; . /home/travis/bin/definitions.B; python test_spacepy.py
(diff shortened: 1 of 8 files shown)


drsteve:
I'll leave this for review, but I just hit something that might want modifying...
When I was borrowing from the MaskedArray I was using up-to-date code as a raference. So the __repr__ in dmarray uses numpy.core.arrayprint.dtype_is_implied, which wasn't implemented until numpy 1.14. That puts a pretty hard version dependency on the new dmarray unless I rewrite the __repr__.

jtniehof:
Sorry to leave this hanging; I need to look over the pycdf stuff and double-check it. check_well_formed is scuzzy both because it's fairly early code and because it's just a wicked complicated problem. And I think flat may have changed; I remember encountering the "wait it's an iterator now" thing a couple of years ago. Either it used to be, or I thought it was, a view.

jtniehof:
There's something weird with this PR now. Since you've been updating your master as things are merged onto the mainline master, we now have some commits on both side (e.g. "benchmark: add pycdf entry typing.") While the PR is waiting, you need to rebase your master against spacepy/master. Incorporating this is going to be "fun."

jtniehof:
I'm also not thrilled about "from distutils.version import LooseVersion as Version". Losing the namespacing is bad enough although I know it's common in the codebase and I've probably lost that war. Also renaming it make the readability really bad.

jtniehof:
Okay, I think this is how we have to handle it:
1) Rebase your master to 144901c115b9cae95585db4df8c455675b4ac003 (i.e. throw away c3fa81368e9c6d441a139dbd57bb3a83859c5472) and force push. That will update this PR to not have the duplicated commit.
2) I'm going to then submit a PR against YOUR master with a few changes. Once you accept that on your branch, this will be updated.
3) Then I'll accept this.

jtniehof:
What was wrong with flat[0], incidentally? https://stackoverflow.com/a/36392089/336412 and https://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.ndarray.flat.html#numpy.ndarray.flat show that it can be sliced. Was this causing a problem?

jtniehof:
Okay, I see that switching to a masked array base changes the behavior of flat, so that it returns a different object that doesn't behave like the flat iterator. That's annoying and would potentially affect any other code that treats dmarrays like ndarrays in this way.

jtniehof:
>>> q = numpy.ma.array([1, 2, 3])
    >>> q.flat[0]
    1
    >>> z = numpy.ma.array([datetime.datetime(2010, 1, 1), datetime.datetime(2010, 1, 2)])
    >>> z.flat[0]
      File "<stdin>", line 1, in <module>
      File "/usr/lib/python2.7/dist-packages/numpy/ma/core.py", line 2589, in __getitem__
        result = self.dataiter.__getitem__(indx).view(type(self.ma))
    AttributeError: 'datetime.datetime' object has no attribute 'view'
    >>> foo = numpy.array([datetime.datetime(2010, 1, 1), datetime.datetime(2010, 1, 2)])
    >>> foo.flat[0]
    datetime.datetime(2010, 1, 1, 0, 0)

That's not high on my fun list. I know we had definite breakage with dmarray subclassing ndarray but it looks like we're going to be chasing MaskedArray bugs now. I also wonder if our issues with converting dmarray to MaskedArray wouldn't be better solved by trying to fix MaskedArray, now that we're into having to fix MaskedArray bugs.

drsteve:
The different behavior in slicing flat iterators is indeed the issue I was seeing, and hence the changes to check_well_formed and a couple of tests. My feeling is that yes, it leaves us with some incompatibilities between ndarray and dmarray... but the incompatibilities are now buried much deeper and are an upstream bug. The problems we have with the old dmarray not masking well are much greater in magnitude, are far more visible, and then the problem is firmly ours. Maybe the ultimate fix is to work out a slick way of wrapping an ndarray such that it *quacks* like an ndarray, but I think that's a lot of work relative to the change to inheriting from MaskedArray. 

I would like to point out that the numpy datetime64 support doesn't have the "view" issue:
`  y = numpy.ma.array([datetime.datetime(2010, 1, 1), datetime.datetime(2010, 1, 2)]).astype(numpy.datetime64)`
`y.flat[0]`
`numpy.datetime64('2010-01-01')`



So what to do?

The rebase and merge should bring this in, even with the additional commit I made while this PR was open. I may be wrong, but I don't think it would require cleanup. So we can do one of several things - let me know which you want:
1. Hit the "rebase and merge" button
2. Rebase my fork as you recommended, merge a new PR into my fork (with what changes?), then come back to this PR and rebase/merge
3. Something else, details needed.

Feel free to hit me up offline to talk me through your preferred option.

drsteve:
Okay, MaskedArray might be a non-starter...
My guess is that we should probably push this as a bug report upstream...

In [1]: import numpy as np
In [2]: import datetime as dt
In [3]: a = np.ma.MaskedArray([dt.datetime.now()])
In [4]: a.any()
AttributeError                            Traceback (most recent call last)
<ipython-input-4-237d43746c7e> in <module>()
----> 1 a.any()

/home/smorley/.local/lib/python2.7/site-packages/numpy/ma/core.pyc in any(self, axis, out, keepdims)
   4792         mask = _check_mask_axis(self._mask, axis, **kwargs)
   4793         if out is None:
-> 4794             d = self.filled(False).any(axis=axis, **kwargs).view(type(self))
   4795             if d.ndim:
   4796                 d.__setmask__(mask)

AttributeError: 'datetime.datetime' object has no attribute 'view'


Of course, the MaskedArray-derived dmarray suffers the same problem. This is a killer for me in its current state.

jtniehof:
The encouraging thing is, I think that's the same place where flat subscripting fails, some sort of bug on touching the masks with object arrays. So we might be able to fix it upstreams and if we're very lucky we can monkey-patch it in to dmarray.

We really do want your CI stuff. I wonder about accepting this PR, then putting together another commit that reverts dmarray back to ndarray for now, and re-opening a branch off of master that reverts the reversion and lets us work on it for awhile. Thoughts?

drsteve:
That seems messy. How about I deal with the mess at my end?
I think I can probably make two new branches, one with the CI and one with the dmarray updates. Then I can send a new PR for the CI stuff (to master) and I'll work out how to send a PR for the dmarray stuff to a new_dmarray branch. Hopefully I'll get to that tomorrow and once that's all done we can just close this PR?

I know that separates the discussion from the dmarray PR, but really the core of the discussion isn't about *our* problem so much as it is about a numpy.ma.MaskedArray problem.

PR #16, opened 2018-08-12
pycdf: handle reading empty NRV variables

jtniehof:
We had some issues with accidentally reading NRV variables that had never been written to; basically this returned default CDF pad data (not desired). I got this to work more like RV variables with the exception of scalars, which I don't even know what should be the right behavior. I'm going to open an issue for that; in the meantime, this should be a substantial improvement.


drsteve:
Closing. Will reopen immediately to trigger CI. Assuming it passes then I'll merge.

Issue #7, opened 2018-07-21
CI needed, python 2 and python 3


balarsen:
We should get CI setup, all the cool kids are doing it. 

What versions?
I use a bit of python2.7 and a lot of 3.6 (3.7 is also just out).

Are we still supporting 2.6 I forget?

drsteve:
I've got Travis hooked in to a couple of my other repos - shouldn't be too hard to get it going for SpacePy. I'd say we should do 2.7, 3.5 and 3.6 at a minimum. I also don't think we need to keep supporting 2.6 since end-of-life for Python 2 is planned. Actually, I'm not sure that we still really do support 2.6, but if we do I'm happy to drop it. I guess we should document that.

hugovk:
No need for Python 2.6, and some projects are planning or already dropping Python 2.7.

Here's the pip installs for SpacePy from PyPI for June 2018, showing no Python 2.6:

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 3.6            |  44.44% |             24 |
| 2.7            |  40.74% |             22 |
| 3.5            |  12.96% |              7 |
| 3.7            |   1.85% |              1 |
| Total          |         |             54 |

Source: `pypinfo --start-date 2018-06-01 --end-date 2018-06-30 --percent --markdown SpacePy pyversion`

balarsen:
For the CI we will want to install packages in a couple of ways to assess into the future the compatible versions. Will mean a kinda involved CI but its all in the background. Relates to #8.

For example:
numpy==1.7.0. #  make sure a version still works but also...
numpy>=1.7.0  #  to make sure it works on the most recent also

drsteve:
Thanks. I know there's another ticket to cover documenting required versions, etc.

It's interesting to see the PyPI install stats. Certainly we still have some users who probably aren't installing from PyPI and @balarsen and I can probably check with known users to see whether 2.6 is still required by them.

I've assigned myself to this and I'll try to find time to set up Travis on my fork in the near future.

drsteve:
Current pull request from my fork includes an initial Travis CI setup (3.5 and 3.6 only) and I've turned on the Travis GitHub app. Not closing this issue just yet as I was having some trouble with the Python 2.7 build... not test failures - it was freezing on trying to compile sources. The CI could/should also be expanded to test different versions of dependencies as @balarsen noted above.

[...]

PR #19, opened 2018-08-16
Add continuous integration using TravisCI

drsteve:
This PR:
- adds .travis.yml to set up the Travis CI runs. It uses Python 3.5 and 3.6.
- updates the default leapsecond data so that some tests now pass without needing to pull updated leapsecond data using `spacepy.toolbox.update` (tests would pass for local installs with an up-to-date set of leapseconds in the `.spacepy` folder, but not on a clean container)
- `test_time` was updated so that the regression test for `randomDate` passes on Python 2.x and 3.x, which appear to round the microseconds differently (issue is upstream of us, probably in the rounding used by `date2num`)
- resolves issue #7 with the currently listed caveat that the 2.7 build seemed to stall on Travis, so CI only covers 3.5 and 3.6. 

TravisCI results for this branch can be found at [https://travis-ci.com/drsteve/spacepy/builds/81975789](https://travis-ci.com/drsteve/spacepy/builds/81975789)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,5 +18,5 @@ install:
  - wget https://spdf.sci.gsfc.nasa.gov/pub/software/cdf/dist/cdf37_0/linux/cdf37_0-dist-cdf.tar.gz; tar xzvf cdf37_0-dist-cdf.tar.gz; cd cdf37_0-dist; make OS=linux ENV=gnu all; make install; cd ..
 
 script:
- - sudo python setup.py install
+ - python setup.py install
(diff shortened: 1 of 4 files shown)


drsteve:
I just closed my previous, somewhat messy, PR.
This one separates out the continuous integration commits: I checked out upstream/master, created a new branch from that, cherry-picked the CI-related commits, then pushed the branch to my github fork. So it should be a clean set of commits that implement the CI and update tests where required, and nothing else.

PR #21, opened 2018-09-07
test_lib: direct tests of libspacepy C functions

jtniehof:
One of the big issues I'm having in the Windows port is memory access violations when loading libspacepy or irbem. We don't have any tests that just hit libspacepy; most of them pull in the full Python machinery, so keeping libspacepy stuff separate from irbem is a bit unpleasant. This PR gives a few tests that just push on the C library.


balarsen:
Good idea to expand out a bit into this realm. When we get to LANLGeoMag backend we will want to think on some of this.

PR #39, opened 2018-09-27
docs: Replace pngmath with imgmath (closes #35)

### LINKED ISSUE #35: Sphinx >=1.8.0 incompatible with spacepy docs build
The new sphinx build deprecates and removes sphinx.ext.pngmath extension so the docs won't build. We will need to play with what versions work and backwards compatibility etc. One solution would be setting a max version on sphinx at 1.7.9 for the time being. 

http://www.sphinx-doc.org/en/master/contents.html#

jtniehof:
Duelling PR here, but I see no reason to pick up a version dependency when we can work around it.

### DIFF
--- a/Doc/source/conf.py
+++ b/Doc/source/conf.py
@@ -26,11 +26,16 @@ sys.path.insert(0, os.path.abspath(os.path.join('..', '..')))
 # If your documentation needs a minimal Sphinx version, state it here.
 #needs_sphinx = '1.0'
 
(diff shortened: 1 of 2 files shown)


jtniehof:
I reject your pull request and substitute my own! :)

jtniehof:
It does get included in the docs. That section renders like garbage, but I updated using the style that was in place before. Fixing that should be a different issue.

PR #44, opened 2018-10-05
setup: create a requirements.txt, closes #38. This is an initial vers…

### LINKED ISSUE #38: Need requirements.txt for pip
We need a `requirements.txt` file for pip.   We currently cannot take a bare python install and do a `pip install spacepy` even can't do `pip install https://github.com/spacepy/spacepy`

This file is a good way to document version requirements and could ever interface with the CI having several requirements.txt files in the repo and calling them in the CI rather than a hard code.

balarsen:
…ion and we may well need to tune. But this worked great on my system on a fresh virtualenv.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,5 @@
+numpy>=1.6,!=1.15.0
+scipy>=0.10
+matplotlib>=1.5
+h5py
+ffnet


balarsen:
Before this is merged full in we will also want to update the CI. The `pip install` in the CI should then call `pip install -r requirements.txt`.

jtniehof:
I'd like to wait to merge this until I have a chance to spend more time on the installer in general. Incidentally, if you want ffnet to be pulled in by default (probably good), also need to have networkx...networkx is a dependency of ffnet but it doesn't declare it.

balarsen:
Interesting it did work and did install networkx. FFNET setup.py does declare it this way:

```
requires          = ['numpy', 'scipy', 'networkx'],
install_requires = ['numpy>=1.4', 'scipy>=0.8', 'networkx>=1.3']
```
So why does the spacepy setup.py not work...?

balarsen:
Maybe spacepy `setup.py` just needs `install_requires`.  I am happy enough for this pull request to no go anywhere if it points to issues.

drsteve:
@jtniehof ffnet used to not have networkx declared as a dependency, but they do now. Trawling through the commit history for ffnet it looks like it was changed in Feb. 2016. The current release on PyPI does it properly.
@balarsen setup.py has an ```install_requires``` set behind an ```if use_setuptools:```. That's been there for at least 4 years, probably longer.

jtniehof:
Glad ffnet got that straightened out so we can pull that out (maybe requiring at least ffnet 0.8something). I do wish we could retrain lanlstar on scikit-learn and get rid of one f2py....

It's possible newer versions of pip don't force-import setuptools so that's not being caught. A quick debugging print at the top of setup.py should do it.

Having a few requirements.txt set up may be handy anyhow; it would be good to have "pip install without Fortran" command lines, for instance.

drsteve:
Since this only affects installation with an explicit ```-r``` flag, I move that we accept this and close the related issue. Fixing installation through pip3 is a different (albeit somewhat related) problem and should be in its own issue, and it'll need a different PR to resolve it.

jtniehof:
I agree, it's useful to have this in so we might as well.

Issue #24, opened 2018-09-18
pip install stalls for python 2.7


jslavin:
I was using spacepy back in July with no problem and installed it via pip.  I recently ran into a problem with my anaconda install and decided to do a fresh install.  Now installing with pip to a python 2.7 distribution doesn't work (it does work for python 3.6).  It gets to the point:
Building wheels for collected packages: spacepy
  Running setup.py bdist_wheel for spacepy ...
And then just sits there.  Anyone have suggestions on this?

Jon

drsteve:
Thanks @jslavin 
Assuming that you're just doing a straight pip install spacepy, so pulling from PyPI instead of a custom source (like github), then I'm really not sure. The version on PyPI hasn't been updated for a while, so my guess is that the issue is probably something to do with either pip, numpy.distutils or setuptools.

I'm also unable to reproduce. On a Red Hat Enterprise linux system (with Python 2.7) I seem to hit an ```ffnet``` build error, but installing spacepy with ```pip install spacepy --user --no-deps --ignore-installed``` works. That's true for both my regular environment and a clean virtualenv.

On my standard machine (Kubuntu 16.04LTS) I can install from pip (for Python 2.7) using either ```pip install spacepy --user --no-deps``` (which would avoid any ffnet issues, and uses whatever random versions of dependencies I had installed) or ```pip install spacepy --user --ignore-installed``` (which basically forces everything to be reinstalled, even if the dependencies are already met).

Can you  let us know the output from
```
pip --version
python -c "import numpy; print(numpy.version.version)"
python -c "import setuptools; print(setuptools.__version__)"
```
along with your OS version?

Thanks,
Steve

jslavin:
Yes I just did pip install spacepy.  That was with conda in a 2.7 environment. I'm running CentOS 7.
> pip --version
pip 18.0 from /export/slavin/anaconda/envs/py2/lib/python2.7/site-packages/pip (python 2.7)
> python -c "import numpy; print(numpy.version.version)"
1.15.0
> python -c "import setuptools; print(setuptools.__version__)"
40.2.0

I actually solved my problem by just copying over my spacepy directory from my old anaconda install to the new one.  Still it would be nice to understand why pip install didn't work.  I believe that's how I installed it before.  I also tried downloading the code from github and doing "python setup.py build". I used --fcompiler=gnu95.  I ended up getting a bunch of errors related to implied do-lists and in the end it hung.  After killing it I got the message:
[...]

PR #46, opened 2018-10-10
Test against specific numpy versions, plus latest release

### LINKED ISSUE #30: Python 2.7 CI
CI hangs on 2.7 with @drsteve 's latest trials. This might be related to #24 since both seem to be failures in fortan build. But we should ideally have travis running on 2.7, at least for now.

### LINKED ISSUE #31: Pinned dependencies on CI
Continuous integration builds on pull requests shouldn't fail just because a dependency has been updated, so we should probably pin versions there.

drsteve:
Updates to the Travis CI configuration.

Specifically, this adds a testing matrix with specified versions of Python and numpy.
 - Python 2.7 is now tested. Closes #30 
 - Specific versions of numpy are tested. Closes #31 
 - The latest release of numpy will also be tested as one matrix member uses constraints, e.g., >=1.6,!=1.15.0

All testing passes. Numpy 1.1.5.0 is specifically excluded from testing due to the issue described in #24 and #30 

For now I've kept the testing matrix relatively small. It can always be expanded as we develop a concrete plan for what needs testing (which I see as a future enhancement, not required to close the issues at hand). This PR provides a blueprint for expanding the testing in the future..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,17 @@ matrix:
  include:
   - python: 2.7
     env: NUMPY_VERSION="==1.15.2"
+  - python: 2.7
     env: NUMPY_VERSION="==1.13.3"
+  - python: 2.7
     env: NUMPY_VERSION="==1.11.3"
   - python: 3.5
     env: NUMPY_VERSION=">=1.6,!=1.15.0"
+  - python: 2.7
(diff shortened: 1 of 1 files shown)


balarsen:
Thanks Steve!

PR #48, opened 2018-10-24
Fix windows build of libspacepy

jtniehof:
With this PR, the Windows build works: setup.py build/install work, bdist_wininst work, and pip install from the git commit.

I haven't run all the unit tests (those will result in separate bugs as necessary) and I haven't tested that the bdist_wininst output actually installs (that's actually pretty hard with several parallel Anaconda environments, as it requires registry hacking, but it will be tested in the future). Also this doesn't close the problem of pip install without numpy, that's unrelated to the Windows build setup.


drsteve:
I truly want to believe that there is an easier way... I haven't been able to test a Windows build (I should be able to in the relatively near future), but the changes look straightforward. I'll merge it in.

PR #49, opened 2018-11-08
Scripts to do actual spacepy build in Windows

jtniehof:
Fixes the win_build_system_setup batch script and adds two new ones: one which builds all the versions of SpacePy, and one which tears down the Windows build system. This allows a build of SpacePy Windows installers for all versions (Python 2.7/3.6 and 32-bit/64-bit) without touching any other installations of Python.

I hate batch.

### DIFF
--- a/developer/scripts/build_win.cmd
+++ b/developer/scripts/build_win.cmd
@@ -1,0 +1,36 @@
+:: Build spacepy on windows
+:: Assume win_build_system_setup.cmd has already been run
+@ECHO OFF
+SETLOCAL EnableDelayedExpansion
+
+FOR %%B in (32 64) DO (FOR %%P in (2 3) DO CALL :build %%B %%P)
+
+GOTO :EOF
+
+:build
+IF "%1"=="32" (
+    set CONDA_PKGS_DIRS=%USERPROFILE%\Miniconda3\PKGS32
+    set CONDA_SUBDIR=win-32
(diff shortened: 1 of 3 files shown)


drsteve:
Thanks!
I'll merge, but note that it looks like it's hardwired for a miniconda setup. I have a new windows box at work that I can test on in a couple of weeks, so I'll give it a go then and see if there's a neat way of making that a bit more flexible. Looks fine so far though, as long as Anaconda doesn't stop supporting that flag to force 32 bit!

jtniehof:
It's hardcoded to Miniconda because the point isn't to support an arbitrary Windows build environment (the straight up setup.py bdist still works, and is used by these scripts); it's to make it possible to set up all the necessary Python environments for all installers in one go with minimal manual intervention. The point was "Anaconda does it all for you" and two years later it really doesn't, but I've managed to mostly beat it into shape.

I do intend to test the results against a non-Anaconda install. But that's another day.

drsteve:
I actually meant Miniconda rather than Anaconda. The small install has
different directory names than the full Anaconda, I guess?

Anyway, no biggie. I just wondered. I'll try it when I'm back.

jtniehof:
Hmmm. I was using Miniconda just to minimize any interaction with anything somebody would have otherwise (and having the actual build-in-me environment independent of even the miniconda setup.) Didn't realize the paths were separate.

PR #71, opened 2018-12-05
DOC: definitions of coordinates 

dkillick:
We're exploring using SpacePy to provide the basis for some coordinate transform operations using the Geomagnetic Coordinate System. Being not an expert in the field, I found it hard to work out whether SpacePy could provide such functionality based on the docs, as the supported coordinate systems are provided as abbreviations only.

As such, to lower the future bar to entry for people such as I, I've added a short definition of each supported coordinate. I've also updated the link to the IRBEM docs, as the current link seems to have stopped working.

Note: my sources for providing the definitions of the coordinates were:
* https://www.spenvis.oma.be/help/background/coortran/coortran.html
* https://www.vdl.afrl.af.mil/programs/ae9ap9/files/package/Ae9_Ap9_SPM_Users_Guide.pdf#page=16 (PDF; at time of writing also has an incorrectly configured security certificate)

As I'm not an expert in the field I'm not _certain_ the definitions are correct (especially as the two sources used slightly different definitions in a few cases), although the contents of the sources seems to relate directly to the subject matter.

### DIFF
--- a/spacepy/coordinates.py
+++ b/spacepy/coordinates.py
@@ -27,7 +27,7 @@ class Coords(object):
     in units of Re and degrees
 
     Coordinate transforms are based on the IRBEM library; `its manual
-    <http://irbem.svn.sourceforge.net/viewvc/irbem/trunk/manual/user_guide.html>`_
+    <http://svn.code.sf.net/p/irbem/code/tags/IRBEM-4.4.0/manual/user_guide.html>`_
     may prove useful. For a good reference on heliospheric and magnetospheric
     coordinate systems, see Franz & Harper, "Heliospheric Coordinate Systems",
     Planet. Space Sci., 50, pp 217-233, 2002.
--- a/spacepy/coordinates.py
+++ b/spacepy/coordinates.py
@@ -27,7 +27,7 @@ class Coords(object):
     in units of Re and degrees
 
     Coordinate transforms are based on the IRBEM library; `its manual
-    <http://svn.code.sf.net/p/irbem/code/tags/IRBEM-4.4.0/manual/user_guide.html>`_
+    <http://svn.code.sf.net/p/irbem/code/trunk/manual/user_guide.html>`_
     may prove useful. For a good reference on heliospheric and magnetospheric
     coordinate systems, see Franz & Harper, "Heliospheric Coordinate Systems",
     Planet. Space Sci., 50, pp 217-233, 2002.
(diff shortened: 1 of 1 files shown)


jtniehof:
Thanks! I've assigned myself and Steve to take a look at this...Steve is more familiar with the coordinates module than I am, and I'll check the rendering across a few different output formats. The usual song that AGU is coming up so might be slow, but we should be able to get to it.

drsteve:
Thanks for this. It's good to have a fuller, and clearer, description of this.

Hopefully I can clear up the various definitions that IRBEM uses...

GDZ is definitely geodetic and double-checking the IRBEM code (as bundled in SpacePy) the datum used is WGS84. The documentation reports numbers for an older datum, maybe GRS-67? Those numbers aren't used in the code though.

The definition of the Earth radius used for conversion between km and RE is 6371.2, which is the mean radius. This is slightly different to most conversions in space physics which use 6371.0 (the volumetric radius; i.e., the radius of a sphere of equal volume to the WGS84 ellipsoid).

There are a lot of unusual spellings an abbreviations throughout the IRBEM code and documentation as the development has largely been led by a French group at ONERA. Much of the in-code commenting is actually in French, as are some of the function names. My guess is that this is the reason behind the 'z' in geodetic.

Since the WGS84 datum is a de facto standard it probably isn't critical to have in the documentation, although it's probably useful to add that back in. It's probably worth noting in this context that their RLL coordinate system uses the radius, but geodetic latitude and longitude (not geographic). Maybe that description should be
`**RLL** (Radius, Latitude, Longitude; Geodetic)`

I will note here that the reference for the GEI coordinate system would ideally be given in the documentation. The GEI coordinate system used by IRBEM is, I believe, what's generally known as True-Of-Date (rather than, as I have seen people assume, J2000). It's probably good to describe that as
`**GEI** (Geocentric Equatorial Inertial; True-of-Date),`
That way anyone who needs to know will have the information there and it's easily ignored by folks who don't know what the reference means.

I'm also not against making references to the background on this stuff easier to find, so adding in the DOI for the reference already there would be a good idea. I'm loathe to link the SPENVIS docs as they provide a simplification over our cited reference which omits relevant detail about *how* it's simplified. The Franz and Harper paper is more complete.

How does that all sound?

jtniehof:
Thanks for the check, Steve. I'm good for reintroducing WGS84, then, and let's link the DOI but not SPENVIS. I also think Steve's suggestions are good for a little more clarity.

Peter, you've been doing the work here, so if you want to make those changes then we'll merge and keep your fingerprint on...or we can just merge here and Steve or I can make another PR with those changes.

Thanks for the contribution!

Also, not relevant to the PR but your larger goal: if there's something in the coordinates setup that doesn't work for you, please file an issue and we'll see if we can make it work. Steve is also looking at a pretty major overhaul of coordinates which is likely to be into the new year at this point, but may also improve the situation.

dkillick:
Brilliant, thanks! 🎉

PR #72, opened 2018-12-07
Fix imgmath import check in sphinx conf

jtniehof:
Somehow I managed to screw up the fix for #35 so hopefully this fixes the fix...


drsteve:
And somehow it didn't get picked up in review. This is definitely the right exception to use, so I'll merge now.

Thanks for picking this up.

PR #78, opened 2018-12-13
Toolbox norm

balarsen:
OK live and learn. The way of doing business for pull request means make a branch. 

I needed the ability to normalize to an arbitrary range and some thinking and research led to an elegant solution using numpy.interp.

This is useful for 0-255 scaling, -1-1, or whatever a user may want. Tests updated and all pass.

### DIFF
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -1931,19 +1931,23 @@ def quaternionConjugate(Qin, scalarPos='last'):
 
 # -----------------------------------------------
 
-def normalize(vec):
+def normalize(vec, minmax=(0.0, 1.0)):
     """
-    Given an input vector normalize the vector
+    Given an input vector normalize the vector to a given range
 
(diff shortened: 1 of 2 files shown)


balarsen:
Come up with a name you like for min max. Numpy linspace uses start, stop.

drsteve:
As hinted at in the last comment, I was thinking that `minmax` wasn't a great fit in kwarg naming.

My suggestions would be to either break it into two keyword arguments (`low` and `high`), or simply change it to something that's not a compound word (like `limits`).

I think either of those would be fine.

drsteve:
Can I also request a change on the testing?

Multiple tests in a single test case should be kept to a minimum really. So maybe break the tests up into:
- Test for rescaling to smaller range
- Test for rescaling to larger range
- Test for rescaling including NaN
Then add:
- Test for rescaling to same range

I think I'd be ready to merge after that.

PR #94, opened 2019-02-08
Pycdf attrs speedup (closes #89)

### LINKED ISSUE #89: pycdf attribute writing speed
Writing an entry to an attribute can be awfully slow. Direct assignment is slow from type guessing, but even calling .new() is not fast. I think part of the issue is checking to see if it exists several times..."new" explicitly checks for existence so it can raise a KeyError, and then it calls _get_or_create which tries again to see if it's in the file. I think _get_or_create can be replaced with a direct call to the AttrType constructor to avoid at least one of these. There's also lots of calls to global_scope that might be overkill (and honestly the SCOPE attribute should ensure that always works; those should maybe be class attributes anyhow).

This isn't an issue when just writing one attribute, but we've got some files with hundreds of variables and dozens of attributes each, and all these checks require passing through the ctypes barrier into the CDF library.

I probably need to implement a reasonable benchmark that's thinner than the examples I've been running into and work against that.

jtniehof:
This speeds up code that writes a lot of attributes by about a factor of 3, and would also substantially help out anything that's just hitting a lot of variables. It switches from checking the CDF for a variable/attribute number every time a variable/attribute is referenced to cacheing the number (and expiring the cache when things happen that might change it.)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -12,9 +12,8 @@ pycdf
  - Reading empty NRV variables will now return empty array (not padding)
  - Fix more cases of BAD_ENTRY_NUM when reading attribute entries
  - Fix UTC-to-TT2000 conversions on ARM (thanks Rodrigue Piberne for the bug)
- - Speed up many operations by referencing variables by number instead of
-   name. This is the biggest change to pycdf internals in several years; please
-   report any bugs.
+ - Speed up many operations by cacheing variable numbers. This is the biggest
+   change to pycdf internals in several years; please report any bugs.
 
 Changes in Version 0.1.6 (2016-09-08)
 =====================================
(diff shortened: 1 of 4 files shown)


jtniehof:
The gotchas have sadly been there all along.

I wouldn't mind a PEP-8 scrub (although it seems people do have some differing opinions on implementation). My experience with pylint has been that it gets cranky to the point of uselessness quickly, and it's very bad with code designed for both Python 2 and 3. I hope the changes aren't introducing new style violations, although the old code definitely has issues (and the changes roughly match). I'd rather start with a docs rework to get away from numpydoc (which would also provide a chance to scrub the last epydoc) and then start in on PEP8 and related.

PR #95, opened 2019-02-10
pybats: trap Window MemoryError reading ascii IDL file with binary assumption

### LINKED ISSUE #86: MagGridFile raises MemoryError on read (Windows)
Reading a "mag_grid..." file from BATSRUS using `spacepy.pybats.bats.MagGridFile` raises a MemoryError on Windows (Windows 10, Python 3.7 (32-bit)).

```
(spacepy) C:\Users\morle\Downloads\dBn_images>ipython
Python 3.7.1 (default, Dec 10 2018, 22:09:34) [MSC v.1915 32 bit (Intel)]
Type 'copyright', 'credits' or 'license' for more information
IPython 7.2.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import spacepy.pybats.bats
Qin-Denton/OMNI2 data not found in current format. This module has limited functionality.
Run spacepy.toolbox.update(QDomni=True) to download data

In [2]: spacepy.pybats.bats.MagGridFile('run_fullmag_030/mag_grid_e20100405-070050.out')
MemoryError                               Traceback (most recent call last)
<ipython-input-2-646999d70899> in <module>
----> 1 spacepy.pybats.bats.MagGridFile('run_fullmag_030/mag_grid_e20100405-070050.out')

~\Anaconda3\envs\spacepy\lib\site-packages\spacepy\pybats\bats.py in __init__(self, *args, **kwargs)
   2867
   2868         # Initialize as an IdlFile.
-> 2869         super(MagGridFile, self).__init__(header=None, *args, **kwargs)
   2870
   2871         # Additional header parsing:

~\Anaconda3\envs\spacepy\lib\site-packages\spacepy\pybats\__init__.py in __init__(self, filename, format, header, keep_case, *args, **kwargs)
    866         self.attrs['file']   = filename   # Save file name.
    867         self.attrs['format'] = format     # Save file format.
--> 868         self.read(header, keep_case)   # Read file.
    869
    870     def __repr__(self):

~\Anaconda3\envs\spacepy\lib\site-packages\spacepy\pybats\__init__.py in read(self, header, keep_case)
    880         if self.attrs['format'] is None:
    881             try:
--> 882                 _read_idl_bin(self, header=header, keep_case=keep_case)
    883             except (ValueError,EOFError):
    884                 _read_idl_ascii(self, header=header, keep_case=keep_case)

~\Anaconda3\envs\spacepy\lib\site-packages\spacepy\pybats\__init__.py in _read_idl_bin(pbdat, header, keep_case, headeronly)
    599
    600     try:
--> 601         headline=readarray(infile,str,np.int32)
    602     except (ValueError,EOFError):
    603         endian='big'

~\Anaconda3\envs\spacepy\lib\site-packages\spacepy\pybats\__init__.py in readarray(f, dtype, inttype)
    528     # Read the data
    529     if dtype is str:
--> 530         A=f.read(rec_len[0])
    531     else:
    532         A=np.fromfile(f,dtype=dtype,count=int(rec_len[0]/dtype_size_bytes))

MemoryError:
```

Suggested fix: As `rec_len` has already been checked for inconsistencies, and reading a string type looks like it's reading the full file, then line 530 should probably be changed to
```
A = f.read()
```
This fixes the issue on Windows. Not tested on other systems yet. There should also be tests for this in test_pybats.

(Data file attached, remove `.txt` extension as this was added so GitHub would allow upload.)
[mag_grid_e20100405-070050.out.txt](https://github.com/spacepy/spacepy/files/2717479/mag_grid_e20100405-070050.out.txt)

drsteve:
Type guessing on Windows leads to a MemoryError when the ASCII file is read using _read_idl_bin.

Summary of changes:
- `spacepy/pybats/__init__.py`: This PR traps for the error, and also wraps the whole `_read_idl_bin` function in a context manager so that the file is closed on raising any exception. *It looks like massive changes, but it's really just the diff not catching the indent for the `with` block*
- `tests/test_pybats.py`: A test is added to make sure the typeguessing doesn't fail.  This previously failed on Windows 10 (Python 3.7.1, 32 bit), but now passes (local tests). Spacepy's Travis suite passes with the new code and tests, but only tests on linux.

This PR resolves #86

### DIFF
--- a/spacepy/pybats/__init__.py
+++ b/spacepy/pybats/__init__.py
@@ -525,6 +525,26 @@ def readarray(f,dtype=np.float32,inttype=np.int32):
         # Zero-length record...file may be truncated
         raise EOFError('Zero-length read at start marker')
 
+    try:
+        startpos = f.tell()
+    except IOError:
+        seekable = False
+    else:
+        seekable = True
+    if seekable:
+        f.seek(0, 2)
+        endpos = f.tell()
+        if endpos - startpos < (rec_len[0] + np.dtype(inttype).itemsize):
+            raise EOFError('File is shorter than expected data')
+        f.seek(startpos + rec_len[0], 0)
+        rec_len_end = np.fromfile(f,dtype=inttype,count=1)
+        f.seek(startpos, 0)
+        if rec_len_end != rec_len:
+            raise ValueError((
+                'Read error: End marker length ({0:d}) does not match start '
+                'marker length ({1:d}).').format(rec_len[0], rec_len_end[0]) + 
+                'This indicates incorrect endiannes, wrong file type, '
+                'or file is corrupt.')
     # Read the data
     if dtype is str:
         A=f.read(rec_len[0])
--- a/spacepy/pybats/__init__.py
+++ b/spacepy/pybats/__init__.py
@@ -586,171 +586,165 @@ def _read_idl_bin(pbdat, header='units', keep_case=True, headeronly=False):
     # Open, read, and parse the file into numpy arrays.
     # Note that Fortran writes integer buffers around records, so
     # we must parse those as well.
-    infile = open(pbdat.attrs['file'], 'rb')
-    
-    # On the first try, we may fail because of wrong-endianess.
-    # If that is the case, swap that endian and try again.
+    with open(pbdat.attrs['file'], 'rb') as infile:
+        # On the first try, we may fail because of wrong-endianess.
+        # If that is the case, swap that endian and try again.
+        endian='little'
 
-    endian='little'
-
-    inttype=np.dtype(np.int32)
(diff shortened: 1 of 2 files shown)


jtniehof:
I can take a crack at extending the test.

Is there any requirement that the `f` input to `readarray` be handled properly if it isn't seekable? Because if seeking is acceptable, the record-length at the end can be checked without allocating all the memory. That would mean you can't read from stdin (although having a fallback to the current approach if seeking fails would be easy.)

jtniehof:
Submitted amendment which doesn't allocate the memory (so never actually triggers a MemoryError) to drsteve/spacepy#1.

I still think it's worth treating MemoryError as a "probably the wrong format" condition.

Issue #14, opened 2018-08-03
License.md is required


drsteve:
Currently the license is in the source code, and can be queried through spacepy.license, but the license text should be in a LICENSE.md.

drsteve:
Closed by #105

PR #105, opened 2019-02-26
Add/update license and copyright info

drsteve:
This PR closes issue #14 

License/copyright information was previously only in the LICENSE attribute in the code. Further LANS, LLC asserted copyright prior to providing open source approval for SpacePy, and that IP was transferred to Triad National Security, LLC with the recent M&O contract change.

This addresses all of these by:
- Providing explicit COPYRIGHT.md and LICENSE.md files
- Abbreviating the LICENSE attribute for maintainability
- Adding a METADATA.in file that should carry the license info with installed wheels
- Updating the URL in the setup to no longer point to the old website, this now points to the github repo.

### DIFF
--- a/COPYRIGHT.md
+++ b/COPYRIGHT.md
@@ -1,0 +1,19 @@
+Copyright (c) 2010, Triad National Security, LLC
+All rights reserved.
+
+This software was produced under U.S. Government contract 89233218CNA000001 for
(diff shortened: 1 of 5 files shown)


jtniehof:
Looks good; particularly being explicit with the portion that LANS claimed before community development. Perhaps we can add "2011-2019 Contributors" to the copyright line at some point, making the break explicit. But it's definitely much more accurate than it was.

PR #115, opened 2019-04-10
Pycdf slow ro again

jtniehof:
I found a case where opening a CDF from master meant it was opened read-only and then read-write, which was needlessly slow.

Also put in a documentation note of the ups and downs of readonly; some benchmarking indicates the cache does help things out in some cases, so always opening read-write doesn't seem like a good idea.


drsteve:
Due to other changelog updates this needs a conflict resolution.
Can you maybe rebase against master and fix the conflicts locally, then push that back here?

jtniehof:
Done.

PR #118, opened 2019-04-11
pycdf: Fix crash in tt2000_to_epoch16 (Closes #60)

### LINKED ISSUE #60: Crash in test_pycdf
Python 2.7, 32-bit Windows, in the middle of test_pycdf crashes with a popup "python.exe has stopped working". `python test_pycdf.py NoCDF` fails. Running with any one test (e.g. `python test_pycdf.py NoCDF.testTypeGuessing`) does not fail. Running `python test_pycdf.py --verbose NoCDF` and it ends with:
```
testTypeGuessing (__main__.NoCDF)
Guess CDF types based on input data ...
```

so looks like the crash is in there but is preconditioned by some other test. They appear to be running alphabetically and TT2000ToEpoch16 is the one right before.

jtniehof:
This fixes a crash caused by tt2000_to_epoch16; the essence is to fix the return value (misdocumented in the CDF C manual, which has been sent upstream.) There's some additional cleanup while I was at it.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -14,6 +14,7 @@ pycdf
  - Fix UTC-to-TT2000 conversions on ARM (thanks Rodrigue Piberne for the bug)
  - Speed up many operations by cacheing variable numbers. This is the biggest
    change to pycdf internals in several years; please report any bugs.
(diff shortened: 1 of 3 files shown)


drsteve:
Thanks @jtniehof

PR #124, opened 2019-04-16
Explicitly preserve dtypes converting SpaceData to numpy recarray

### LINKED ISSUE #62: Test failure in test_datamodel test_toRecArray
Python 2.7, Windows 32-bit, dmarrayTests.test_toRecArray fails on line 457 (assertTrue ra.type in blah blah)

drsteve:
Closes #62 

A test failure on Windows highlighted that dtypes were not guaranteed to be consistent across platforms (e.g., numpy might assign `numpy.array([4])` a dtype of int32 or int64). Also, dtypes were not guaranteed to be preserved on creating a recarray from a SpaceData (the default dtype-guessing was being used on creating the recarray, so specified dtypes may not be honoured).

This PR:
- Changes the failing test so that it doesn't check types, just equality of contents
- Adds new tests to check for type-preservation
- Updates `toRecArray` so that it uses the names/dtypes from the SpaceData contents to set the names/dtypes in the numpy record array
- Notes the change in CHANGELOG

Tested on Python 2.7 (64-bit) on Windows, plus everything in the CI.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,8 @@ Changes in Version 0.1.7 (2017-xx-xx)
 - Fix writing of configuration file on Windows
 ae9ap9
  - Fix handling of gzipped files under Windows/Python2
+datamodel
+ - Datatypes are now explicitly preserved by function "toRecArray"
 pybats
  - Many small changes and bugfixes (thanks John Haiducek)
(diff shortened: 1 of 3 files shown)


jtniehof:
Elegant.

PR #125, opened 2019-04-25
Batsupdate

spacecataz:
This PR updates capabilities in the PyBats submodules.  Amongst small quality of life updates for plotting, there are a few substantial changes:

- The quad and oct tree module now properly handles cases where BATS 2D slices passes between two blocks of different resolution.  This fixes issues with loading files, stream traces, and pcolor plotting.
- The Bats2d.add_b_magsphere_new method now replaces Bats2d.add_b_magsphere as the default method for adding magnetic field lines to 2D plots at Earth.  
- The legacy method for Bats2d.add_b_magsphere has been moved to Bats2d.add_b_magsphere_legacy
- Calls to Bats2d.add_b_magsphere_new cause a depreciation warning.
- rim.Iono files have a new method for calculating the integrated FAC strength: Iono.calc_I

### DIFF
--- a/spacepy/pybats/bats.py
+++ b/spacepy/pybats/bats.py
@@ -1509,8 +1509,9 @@ class Bats2d(IdlFile):
             isNig  = s1.x.mean()<0
             
         # Use last line to get southern hemisphere theta:
-        npts = s1.x.size/2
-        r = sqrt(s1.x**2+s1.y**2)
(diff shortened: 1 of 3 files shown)


drsteve:
Thanks for the updates. Ideally we'd keep the change log updated, but I think we need to add stuff from a number of PRs since the move to GitHub, so I'll self-assign that.

PR #130, opened 2019-05-10
Pycdf more type guessing

jtniehof:
I found an instance where passing a numpy array of dtype object but that  contained entirely integers resulted in a float variable being created. This PR improves the handling of object arrays, by trying to convert them to numeric and explicitly raising an error if that is not possible.

This doesn't affect the handling of object arrays containing datetimes, as those are handled before this code is hit.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,7 +9,8 @@ pybats
  - Many small changes and bugfixes (thanks John Haiducek)
  - Fix handling of gzipped files under Windows/Python2
 pycdf
- - Proper handling of type guessing for negative integers and numpy scalars
+ - Proper handling of type guessing for negative integers, numpy scalars, and
+   object arrays.
  - Allow ... to get all gEntries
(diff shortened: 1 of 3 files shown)


drsteve:
I've read through the code here to try to make sure that this won't have unexpected side-effects. It looks fine to me, but I have this nagging fear that maybe there'll be some datatype that should be properly dealt with earlier, and then falls through into the exception here... Tell me I'm worrying about nothing and I'll merge!

jtniehof:
The diff doesn't give a lot of context. If a numpy array is actually passed in, this code is only hit if it doesn't contain datetimes and is an object array. If something else is passed in, this code is only hit if casting it to a numpy array results in an object array, and again not containing datetimes. You can also check the coverage of the unittests to satisfy further paranoia :)

drsteve:
I have temporarily sated my paranoia.

PR #131, opened 2019-05-10
pycdf: explicitly test that cloning maintains compression

jtniehof:
I had a case come up where it looked like cloning a variable didn't maintain the compression. Wrote a test to track it down and looks like it does...this is just including that test to avoid a regression.


drsteve:
Good idea.

Issue #68, opened 2018-11-20
Test failure: test_irbempy.IRBEMBigTests.test_getLstar_T05


jtniehof:
Error on 64-bit Windows, python 2.7, in getUTC, line 1162 of spacepy time (so this may be a time bug); ultimately "cannot convert float NaN to integer".

Works on 32-bit Windows, python 2.7; 32-bit Windows, Python 3.6; and 64-bit Windows, Python 3.6.

jtniehof:
Notes from #139:
`python test_irbempy.py IRBEMBigTests` fails.
`python test_irbempy.py IRBEMBigTests.test_get_Lstar_T05` succeeds. So it's another preconditioning issue.

The previous one is test_get_Lstar_T01 and `python test_irbempy.py IRBEMBigTests.test_get_Lstar_T01 IRBEMBigTests.test_get_Lstar_T05` fails in T05. `python test_irbempy.py IRBEMBigTests.test_get_Lstar_T05 IRBEMBigTests.test_get_Lstar_T01` succeeds.

So it's not just any L*, it's specifically the T01 seems to set up the T05 for failure.

TB on this is get_omni in setUp, to getUTC in the Ticktock constructor, ultimately to num2date. It's the same off current master (32652a) and in PR #139, which are different IRBEM libraries.

I'm actually encouraged by the fact this is failing in the same place with a different IRBEM library. If it were a truly nasty Heisenbug due to the compiler screwing up or something I'd expect the failure mode to be subtly different.

jtniehof:
MBE (as contrasted with MWE)
```
import spacepy
import spacepy.omni
import spacepy.time
import spacepy.coordinates
import spacepy.irbempy as ib

ticks = spacepy.time.Ticktock(['2001-02-02T12:00:00', '2001-02-02T12:10:00'], 'ISO')
loci = spacepy.coordinates.Coords([[3,0,0],[2,0,0]], 'GEO', 'car')
omnivals = spacepy.omni.get_omni(ticks, dbase='Test')

actual = ib.get_Lstar(ticks, loci, [90], omnivals=omnivals)

ticks = spacepy.time.Ticktock(['2001-02-02T12:00:00', '2001-02-02T12:10:00'], 'ISO')
loci = spacepy.coordinates.Coords([[3,0,0],[2,0,0]], 'GEO', 'car')
omnivals = spacepy.omni.get_omni(ticks, dbase='Test')
```
[...]

PR #139, opened 2019-05-28
Update IRBEM to r616 and add GCC8 support

### LINKED ISSUE #134: cannot import name 'irbempylib' from 'spacepy.irbempy'
Spacepy doesn't work.

```
from spacepy.coordinates import Coords                                  
y=Coords([[1,2,4],[1,2,2]],'GEO','car')  
```

### Error message/Traceback:

```
ImportError                               Traceback (most recent call last)
<ipython-input-2-da18c6fd86b6> in <module>
----> 1 y=Coords([[1,2,4],[1,2,2]],'GEO','car')

~/anaconda3/lib/python3.7/site-packages/spacepy/coordinates.py in __init__(self, data, dtype, carsph, units, ticks)
     87     def __init__(self, data, dtype, carsph, units=None, ticks=None):
     88 
---> 89         from . import irbempy as op
     90         from spacepy.irbempy import SYSAXES_TYPES as typedict
     91 

~/anaconda3/lib/python3.7/site-packages/spacepy/irbempy/__init__.py in <module>
     69 
     70 
---> 71 from .irbempy import *
     72 

~/anaconda3/lib/python3.7/site-packages/spacepy/irbempy/irbempy.py in <module>
     29 import spacepy.datamodel as dm
     30 try:
---> 31     from . import irbempylib as oplib
     32 except ImportError:
     33     if 'sphinx' in sys.argv[0]:

ImportError: cannot import name 'irbempylib' from 'spacepy.irbempy' (/Users/jjung11/anaconda3/lib/python3.7/site-packages/spacepy/irbempy/__init__.py)
```

```
Darwin-18.2.0-x86_64-i386-64bit
sys.version_info(major=3, minor=7, micro=1, releaselevel='final', serial=0)
numpy=1.16.3
scipy=1.3.0
matplotlib=3.1.0
```

Uninstalling and reinstalling didn't work.
I also tried 
pip install https://github.com/spacepy/spacepy/zipball/master
but that version didn't work either.
When I downgraded the Spacepy from 0.1.6 to 0.1.5, it worked.

Can someone help me to use 0.1.6 properly?

drsteve:
The `IRBEMlib` Fortran library underpinning `irbempy` is out-of-date.
This PR:
- Updates to rev616 (dated 4 April 2019)
- Updates regression tests (as Fortran codebase has changed and gives slightly different results)
- Removes old, unused versions of the Fortran library from the codebase
- Adds GCC8 support through invocation of the `std=legacy` flag
- Updates the URL for retrieval of TS07 dynamic coefficients
- Adds docstring to `updateTS07Coeffs`
- Updates dictionary of supported coordinate systems to reflect change of system 13 from TEME to J2000.
- Updates CHANGELOG

This PR resolves #134 

In addition to the CI, this was tested under a new conda environment with GCC 8.1.0, Python 3.7.3, and numpy 1.16.3. It was also tested on a well-loved environment using GCC 5.4.0, Python 2.7.12, and numpy 1.15.1.


drsteve:
I didn't test this on Windows. I can't really do that from the office, but I might be able to find time to do it at home. Can I ask you to? That way if I get to it we've just had two sets of eyes on.

Good catch on the .so being checked in. I'm fairly confident that's something that was inadvertently checked in with a previous version, and so the update was checked in (because I'm not adding all of the changed files manually!) We can probably delete? If you're happy with that then I'll push another commit removing the .so

Line counts... it's really 1 deleted version and one modified version. The source code itself is about 97k lines of Fortran, so the deletion (which includes docs, etc.) looks about right. Similarly, the fact is that there were a few changes and a couple of additions (notably an updated implementation of TS07 using a faster implementation of the Bessel functions, as well as some geopack stuff used for common blocks). So the low number of added lines seems reasonable to me too.

jtniehof:
Ah, right, there are going to be a bunch of lines that don't change and git will track the files as "moved" not "deleted/added". So if it's 10% changed, that reads as 110% deletions and 10% additions. Makes sense.

I will fire up Windows tests.

Another commit w/o the so would be good.

jtniehof:
Works on Python 3, 32 and 64. Works on Python 2 32. Fails on Python 2, 64, with test_get_Lstar_T05 actually failing in the get_omni call of setUp, cannot convert float NaN to integer. test_omni works. I checked and I don't think I have an open Windows bug against this. It might just be an update to my Windows environment but I'll go digging.

jtniehof:
`python test_irbempy.py IRBEMBigTests` fails.
`python test_irbempy.py IRBEMBigTests.test_get_Lstar_T05` succeeds. So it's another preconditioning issue.

The previous one is test_get_Lstar_T01 and `python test_irbempy.py IRBEMBigTests.test_get_Lstar_T01 IRBEMBigTests.test_get_Lstar_T05` fails in T05. `python test_irbempy.py IRBEMBigTests.test_get_Lstar_T05 IRBEMBigTests.test_get_Lstar_T01` succeeds.

So it's not just any L*, it's specifically the T01 seems to set up the T05 for failure. This is going to take more time than I have right now to digest.

drsteve:
It's a good thing that Python 2 is almost EOL!! ;-)

I'll try to take a look at see whether I can reproduce or understand the failures. I'll also make another commit that removes the .so.

I think it's worth thinking about merging this (after next commit) despite a Windows failure on 64-bit Python 2 (because the test itself passes). I believe (commenting from phone) that we already have a similar issue in the tracker. I'm not sure I see how the IRBEM changes could cause a preconditioning failure, so I'd be inclined to say that this PR is fine and we need to fix the underlying issue.

jtniehof:
Let me check on master before we merge. It's entirely possible there's a memory usage bug in irbempy or in IRBEM itself that would be the cause. If it fails on master, I agree, we merge; otherwise we should track it down before the merge.

drsteve:
Is this issue #68 again/still?

jtniehof:
Could have sworn I checked all the windows-tagged bugs before declaring this new...

TB on this is get_omni in setUp, to getUTC in the Ticktock constructor, ultimately to num2date. It's the same when run off master, so this definitely is #68. I'll move my notes from here over to there.

I'm actually encouraged by the fact this is failing in the same place with a different IRBEM library. If it were a truly nasty Heisenbug due to the compiler screwing up or something I'd expect the failure mode to be subtly different.

Given that, I'm okay with the merge here once the .so is gone.

jtniehof:
Just waiting on the CI; any chance it got hung up somehow?

drsteve:
I don't know why the second build got queued... the first one was successful and was prompted by the new commit. The queued one is weird. I'll go hit the restart button in Travis, but the tests already passed...

jtniehof:
It seems we always have two going for a PR now.

drsteve:
I don't think Travis knows that something is supposed to be queued. (https://travis-ci.com/spacepy/spacepy/pull_requests)

Tests have now passed on this PR several times. I think Github's Travis integration is confused. I'd be fine with merging. Hopefully that'll end Github's confusion.

jtniehof:
Hmmm, you're right, most of them are getting only one. Bizarre.

PR #147, opened 2019-06-09
Poppy test updates for Windows (Closes #67)

### LINKED ISSUE #67: Test failure: test_poppy.AssocTests.testAssocBootstrapRegress
Error on 64-bit Windows, python 2.7 and Python 3.6, line 507 of poppy, on initializing lag_seeds: `ValueError: low is out of bounds for int32`.

Works on 32-bit Windows, python 2.7 and 32-bit Windows, Python 3.6.

jtniehof:
This makes the regression tests for PoPPy work on Windows and adds 32-bit "expected" values to catch any future regressions.

Because 64-bit Windows has a 32-bit long, and everything about the RNG is built around the size of long, Windows will always act basically like a 32-bit OS in this regard. If anybody wants to do the work to rewirite randomkit or otherwise work around it, that would be a fine enhancement, but this fixes the basic test failure.

0b460c1 is not strictly necessary here; it was an attempt to force 64-bit Windows to use 64-bit seeds. Enough work went into it that I thought it worth preserving in the repo (in particular if somebody decides to do the 64-bit random number rework.)


jtniehof:
I'm too used to single-commit PRs. This closes #67 (just to get the cross-linking in.)

drsteve:
That's a much shorter/cleaner resolution than I was anticipating. Looks good. Passes on my Windows envs.

PR #148, opened 2019-06-09
test_datamodel: account for windows line breaks and unicode differences

### LINKED ISSUE #63: Test failure test_datamodel.JSONTests.test_toHTML
Fails on the file size check, line 721.
Fails: 32-bit Windows, Python 3.6; 64-bit Windows, Python 3.6
Succeeds: 32-bit Windows, Python 2.7; 64-bit Windows, Python 2.7.

drsteve:
Closes #63 

In `test_datamodel.py`, the test `JSONTests.test_toHTML` checks for a size. The expected size accounted for the bytes->unicode change from Python 2 to Python 3, but not on Windows systems.

This PR adds a check for the type of the string in Windows, and sets the size accordingly.

This *_only_* changes the test, not the functionality. The previous test was failing on Windows under Python 3 because this case wasn't tested correctly.


drsteve:
Forgot to say, tested on Windows 10 in the developer environments (Python 2.7 and 3.6, 32-bit and 64-bit).

PR #151, opened 2019-06-11
Fixed implicit int division in EventClicker

### LINKED ISSUE #4: Problems with plot.utils.EventClicker
I'm using SpacePy with Python 3.6.4 (Anaconda 64-bit). When attempting to use EventClicker from plot.utils, I receive the following error:

  File "/Users/jwoodroffe/anaconda/lib/python3.6/site-packages/spacepy/plot/utils.py", line 358, in _add_event_phase
    ls=self._styles[self._curr_phase / len(self._colors) % len(self._styles)])
TypeError: list indices must be integers or slices, not float

This is due to a change in how division is handled between Python 2.x and 3.x. When fixed by changing "/" to "//", another error occurs nearby:

File "/Users/jwoodroffe/anaconda/lib/python3.6/site-packages/spacepy/plot/utils.py", line 359, in _add_event_phase
    if not self._xydata is None:
AttributeError: 'EventClicker' object has no attribute '_xydata'

If I fix this by initializing _xydata to None in __init__, I get:

  File "/Users/jwoodroffe/anaconda/lib/python3.6/site-packages/spacepy/plot/utils.py", line 394, in _add_event_phase
    self._data_events.resize((self._data_events.shape[0] + 1,
AttributeError: 'NoneType' object has no attribute 'resize'

And so on. This particular code may be in need of some TLC.

drsteve:
Closes #4.

- Changed a divide on an integer to an explicit `//` so it won't fail in Python 3
- Corrected a typo in the docstring example so that it works with a cpaste into the interpreter

As discussed in #4 the only way I could reproduce the remaining issues in the description was to use non-line-based data. Changing that should be a new enhancement request. (Sorry about the ancient commit, but I forgot I had already committed a fix locally!)

### DIFF
--- a/spacepy/plot/utils.py
+++ b/spacepy/plot/utils.py
@@ -129,7 +129,7 @@ class EventClicker(object):
     True
     >>> troughs = e[:, 1, 0] #x value of event ends
     >>> troughs -= 2 * numpy.pi * numpy.floor(troughs / (2 * numpy.pi))
-    >>> max(numpy.abs(peaks - 3 * numpy.pi / 2)) < 0.2 #troughs near 3pi/2
+    >>> max(numpy.abs(troughs - 3 * numpy.pi / 2)) < 0.2 #troughs near 3pi/2
     True
     >>> d = clicker.get_events_data() #snap-to-data of events
     >>> peakvals = d[:, 0, 1] #y value, snapped near peaks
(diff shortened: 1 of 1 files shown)


drsteve:
Thanks to TravisBuddy for the notification.
I went and checked the Travis CI build. The `pip` install of `numpy` had timed out for one of our builds. I restarted the job and everything ran/passed.

PR #152, opened 2019-06-12
Update testWriteWrongSizeData (Closes #66)

### LINKED ISSUE #66: Test failure: test_pycdf.ChangezVar.testWriteWrongSizeData
Fails on Python 2.7, 64-bit Windows, checking the message on line 2619, because there's a "3L" in there instead of just 3.

Works on Python 2.7, 32-bit Windows; Python 3.6, 32-bit Windows; and Python 3.6, 64-bit Windows.

jtniehof:
numpy shapes are "long" on 64-bit python 2 (probably because of the 32-bit long issue that's been biting us.) So the string representation has a bunch of "L". This PR makes the check for the error message match whatever representation numpy gives us.

This isn't a functionality issue, just a test.

### DIFF
--- a/tests/test_pycdf.py
+++ b/tests/test_pycdf.py
@@ -2770,8 +2770,10 @@ class ChangezVar(ChangeCDFBase):
 
     def testWriteWrongSizeData(self):
         """Write with data sized or shaped differently from expected"""
-        message = 'attempt to assign data of dimensions (3,) ' + \
-                  'to slice of dimensions (5,)'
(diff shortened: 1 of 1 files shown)


drsteve:
Reading the diffs for this one was not illuminating. Context is important!

Anyway, it looks reasonable to me, and I'm glad that you nailed down the issue. Hopefully the future will make supporting SpacePy less about platforms and versions, and more about functionality!

PR #153, opened 2019-06-13
irbem: moved inline comments on statement lines (for f2py)

drsteve:
f2py doesn't like exclamation marks on lines with a declaration e.g. INTEGER or a function call (CALL ...).
These _in-line_ comments have now been moved to adjacent lines, so it quiets down a number of the f2py complaints on compilation. There are still a bunch of warnings from the implied do-loops when constructing the data statements in AE8/AP8, but I don't think that should afect functionality. (It'd still be nice to restructure the Fortran to quiet that down too). I've already pushed this change back up to the IRBEM repo on SF. Most of the changes were in functions we don't use (IDL wrapper functions).

This doesn't touch any code _per se_, but may make f2py's job a little easier. In that sense, it might affect test failures like #68 (although it seems unlikely).


drsteve:
Intentional. Mostly just to avoid adding a new comment below a blank comment line that was already there.

PR #154, opened 2019-06-14
setup: Copy mingw pthread lib to Windows installer.

jtniehof:
This changes the Windows installer creation script to also capture the pthread dll, which seems to be required now.

Originally this machinery was meant to get just the Fortran runtimes from mingw32, so there's also a fair bit of renaming to make clear this is now a general dll-grabber.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -802,29 +802,30 @@ class bdist_wininst(_bdist_wininst):
         _bdist_wininst.finalize_options(self)
         finalize_compiler_options(self)
 
-    def copy_fortran_libs(self):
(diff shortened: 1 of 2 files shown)


drsteve:
Looks fine, but I want to actually run on a Windows box before I merge. It'll have to wait until this evening.

jtniehof:
I'm working on verifying it will fix #155 . It'll definitely fix the issue you were seeing (pthread), but verifying there isn't anything else hanging.

jtniehof:
Confirmed; this will fix #155

jtniehof:
Merge away!

PR #159, opened 2019-06-15
Support for building binary wheels

jtniehof:
This PR adds support for buliding binary wheels (bdist_wheel). The installation of the wheel has only been tested on Windows but that's the main goal at this point. It's been tested with pip install <name of wheel> not via pypi (testing for that will be later.)

There's a false start of trying to separate out the arch-specific and Python-only bits of the build. This was reverted but not rebased away because I think it's something we should do in the future. Ultimately this is treating everything (including Python code) as arch-specific but that beats treating the compiled stuff as non-arch specific. This is basically the most direct route to getting everything included in the wheel and there definitely should be some more intelligent sorting out later.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,8 +14,10 @@ Copyright 2010 - 2014 Los Alamos National Security, LLC.
 #pip force-imports setuptools, on INSTALL, so then need to use its versions
 #but on reading the egg info, it DOESN'T force-import, assumes you are using
 import sys
-if 'pip-egg-info' in sys.argv:
+if any([a in sys.argv for a in ('pip-egg-info', 'bdist_wheel')]):
     import setuptools
+if 'bdist_wheel' in sys.argv:
+    import wheel
 use_setuptools = "setuptools" in globals()
 
 import copy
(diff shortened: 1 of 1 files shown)


jtniehof:
Just pushed an update; was copying the mingw32 dlls twice over (was a bit rushed when wrapped up earlier this afternoon.)

drsteve:
Looks good. I tested on the conda dev environments and everything was great. Wheel builds fine, pip install from wheel works.

I do have a question. I tried using the py3_64 wheel to install on my vanilla python - failed with building ffnet (wanted Visual C++). Setting FC_VENDOR seems to fix (or maybe bypass) that (while I do have MINGW somewhere on my system, it didn't seem to go through the build process.

I imagine that we'll want to update the README with notes for Windows users (recommend conda, set FC_VENDOR?, installer for vanilla python and pip install for conda?)

drsteve:
Update: Installed Visual Studio, pip uninstalled spacepy and ffnet from my vanilla install, restarted. pip install from the wheel worked. I'll merge this one, but I do think we need to have notes for Windows users that are _very_ easy to find.

jtniehof:
I was definitely planning on updating Windows install notes with reference to my install scripts (including the FC_VENDOR). I had forgotten that shipping our own binary wheel still means needing to build ffnet. I'll see if I can't quickly do the pip "dependency for certain functionality" thing to make that not necessary.

PR #161, opened 2019-06-17
Do not require ffnet on Windows

### LINKED ISSUE #160: Handle ffnet dependency with binary install
Right now we hard-require ffnet in our setup.py declarations. That means even with a binary wheel built for Windows, doing a pip install on Windows will require a Fortran compiler for ffnet (and appropriate flags on the install.)

Long-term maybe we can contribute a wheel to ffnet. Short-term we can make LANLstar an "extra" (in setuptools terms). This means:
`pip install spacepy`

will not grab ffnet. (Of course, installing ffnet later would still allow LANLstar to work.) To make sure ffnet works you'd have to do:
`pip install spacepy[LANLstar]`

Now, long-term we also are talking about splitting LANLstar into a "kits" or something so this'll be a bit weird (how-to-get-LANLstar will change) but again, we're shooting for good short-term fixes right now.

Assuming we're okay with this, it should be possible to still hard-require ffnet on other platforms. (For that matter, we can also skip the "extras" thing and just tell people to install ffnet themselves if they want LANLstar on Windows.)

This is turning into multiple-choice, so:

1. Require a Fortran compiler to install a binary wheel on Windows (current)
2. Declare LANLstar an "extra" so you have to install SpacePy in a funny way to get ffnet and LANLstar working
    1. Do this on all platforms
    2. Do this just on Windows while having ffnet still base-required on other platforms
3. Do not require ffnet on Windows so people have to install ffnet explicitly on Windows, but will get it automatically on other platforms.

@drsteve ? @balarsen ?

jtniehof:
This PR removes the requirement for ffnet on Windows so people can install from binary wheel without having a Fortran compiler (for ffnet). This will keep LANLstar from working by default, but explicitly installing ffnet will make that work (without having to reinstall SpacePy.)

This is option 3 on #160 . It it not the only option (or necessarily the one I'm voting for), but it's certainly the _easiest_.

Note this only affects `install_requires` so the metadata still declares an ffnet dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ Our latest release version is available through PyPI and can be installed using
 pip install spacepy --user
 ```
 
+This will also automatically install most dependencies. To permit binary installations without a compiler, this will not install ffnet on Windows. Users needing the LANLstar module can install ffnet separately (requires Fortran compiler); this can be done before or after the SpacePy install.
+
 The latest "bleeding-edge" source code is available from our github repository at [https://github.com/spacepy/spacepy](https://github.com/spacepy/spacepy) and can be installed using the standard
 
 ```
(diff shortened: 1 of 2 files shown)


drsteve:
Okay, on the right PR now... I'm fine merging this so we have something that _just works_ for more users, but I think we need to leave the issue open.

Would it be worth adding a word or two in the README.md (where we describe the dependencies) to say that on Windows it won't install ffnet, so users need to do that themselves for using LANLstar?

jtniehof:
If other solutions to #160 seem better, let's go for those; but if this is the short-term solution, I'd like to close it. We have splitting out into "kits" as the long-term idea anyhow.

Let me take a quick look at docs updates and I'll update this PR.

jtniehof:
I'm okay with leaving #160 open if we can think of something other than splitting LANLstar out that we want to do (e.g. providing a wheel of ffnet ourselves somehow.)

Issue #112, opened 2019-04-04
Use numpy distutils to build irbempy


jtniehof:
We're building the irbempy extension "by hand" with explicit calls to build irbem and then explicit f2py calls. @asreimer in #108 suggested using the numpy distutils for the build, with the extension class. Suggested example at https://github.com/aburrell/apexpy/blob/master/setup.py

I think I've looked at this in the past and it didn't play terribly well with some other stuff we were doing, but that was a long time ago and it's worth another look. We were also considering getting away from numpy distutils since it's been the cause of some problems (#45, 9da472340d23d0892ab0b828861372f8dcb3b908). It's also fixed some problems (3fcb9fdfe15834d7b27d90731f14bdd8dfccce65).

jtniehof:
If we work on a new build system, there are probably other options to consider, e.g. https://mesonbuild.com/

balarsen:
Doesn't #296 make this unneeded?

jtniehof:
Even with #296 we still need IRBEM for L*, etc. (stuf that's in irbempy but not Coordinates).

jtniehof:
Since I'm throwing general build system notes in here, [thread on projects that can't build into wheels](https://discuss.python.org/t/pep-517-and-projects-that-cant-install-via-wheels/791/6). We should be good with this given that, uh, we can build and install from wheels :) but in case something weird shows up there might be hints in there.

jtniehof:
numpy distutils is no more, so this will not happen.

PR #162, opened 2019-06-17
setup: fix pip when numpy not installed (Closes #45)

### LINKED ISSUE #45: pip install fails to resolve dependencies under python 3 (PyPI) or when installing from git
As discussed in #38, installing in a clean environment using ```pip``` only works under Python 2, installing from PyPI. It fails under Python 3, and under both 2 and 3 when pointing to a git repo, e.g., 
```
pip install git+https://github.com/spacepy/spacepy
```

To reproduce
```
~/.local/bin/virtualenv -p python3 py3-clean
cd py3-clean/
source bin/activate
pip3 install spacepy
```

or
```
~/.local/bin/virtualenv py2-clean
cd py2-clean/
source bin/activate
pip install git+https://github.com/spacepy/spacepy
```

This should yield
```
Collecting git+https://github.com/spacepy/spacepy
  Cloning https://github.com/spacepy/spacepy to /tmp/pip-req-build-FI27ko
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-FI27ko/setup.py", line 24, in <module>
        from numpy.distutils.core import setup
    ImportError: No module named numpy.distutils.core
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-req-build-FI27ko/
```

It seems likely that the issue is that we import numpy at the top of the ```setup.py``` which means that an ImportError is raised before we even get to running the setup function. It's unclear why it works under pip2 with PyPI as the source, but it does...

jtniehof:
This PR detects if pip is calling setup.py to get egg info, and if so, allows all the numpy distutils related code to fail relatively gracefully, downloading numpy before the spacepy install. It still seems to throw an error message trying to build the spacepy wheel, but successfully installs by calling setup.py. It's probably trying to build the wheel before finishing the install of everything else.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,8 @@ if any([a in sys.argv for a in ('pip-egg-info', 'bdist_wheel')]):
 if 'bdist_wheel' in sys.argv:
     import wheel
 use_setuptools = "setuptools" in globals()
+#Calling egg info, so a lot of stuff doesn't have to work
+egginfo_only = ('pip-egg-info' in sys.argv)
 
 import copy
 import os, shutil, getopt, glob, re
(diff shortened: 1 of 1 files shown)


drsteve:
I'm happy to merge, but do we actually want to close the issue? (#45)

This clearly provides the short-term fix, but I'd be inclined to leave the issue open as a place to decide on the longer-term solution.

jtniehof:
We do also have #112 which would likely help as a better long-term fix, but this addresses the immediate issue.

PR #163, opened 2019-06-18
Updates to install documentation

drsteve:
Install instructions in documentation is old. This PR:

- changes the homepage link in the Sphinx template to point at the github repo (not the defunct LANL page)
- does a light update of the linux instructions
  - changes EPD and Python(x,y) suggestions to Anaconda and Canopy
  - adds pip to the instructions
  - points to source code at github
- leads the generic install instructions with a pip install
- clarifies suggested installation target in and out of virtual envs
  - i.e., no `sudo` or `--user` in virtual envs, save those for non-virtual env

It's not a comprehensive overhaul, by any means, but hopefully it takes out the worst anachronisms.

### DIFF
--- a/Doc/source/_templates/layout.html
+++ b/Doc/source/_templates/layout.html
@@ -3,7 +3,7 @@
 {% set script_files = script_files + ["_static/copybutton.js"] %}
 
 {% block rootrellink %}
-        <li><a href="http://spacepy.lanl.gov/index.shtml">homepage</a>|&nbsp;</li>
+        <li><a href="https://github.com/spacepy/spacepy">homepage</a>|&nbsp;</li>
         <li><a href="{{ pathto('search') }}">search</a>|&nbsp;</li>
        <li><a href="{{ pathto('index') }}">documentation </a> &raquo;</li>
 {% endblock %}
--- a/Doc/source/install.rst
+++ b/Doc/source/install.rst
@@ -2,7 +2,7 @@
 Installing SpacePy
 ******************
 
-SpacePy uses the standard Python distutils to compile and install.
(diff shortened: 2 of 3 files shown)


drsteve:
I won't be offended if there's a different docs effort going on that some of these changes can be pulled into. If that's the case then I'm happy enough to see the PR rejected - I mostly want to get these on the board so we don't miss them.

drsteve:
Yeah, I wasn't sure about how to phrase the setuptools part. I mean, I don't think it's too misleading for the reasons you mention.

For the package manager... outside of conda I tend to use pip. And there are other ways to do envs (for example, virtualenv and pipenv (in which I would also want to use pip). The package manager is useful for some things, but you are stuck with old versions. Having said that, I agree that discussion of this should go deeper into the linux install page. The question is, do we need that now? Or is this _good enough_ for this release, given that we still have a list of things to do and should push another release in the near future anyway?

Inside conda, I guess using conda install is recommended, but I've had occasional environment resolution issues. And the mix-and-match of conda to pip (when some things aren't _in_ conda) is definitely a pain, especially for installing prereqs, as you pointed out.

I know this PR doesn't get us to having install docs they way they should be. It was more of a quick tidy to stop them being quite so out-of-date (and wrong, due to age). Feel free to do as you will... if you want to use this as a base for modifications from your end, go ahead. If you want me to change something, let me know and I'll try to get to it.

jtniehof:
I think we're on a similar enough page that this should get merged because I would be making modifications from _this_ rather than from our old stuff. Expect another PR from me in a bit; it should be pretty small.

PR #165, opened 2019-06-19
docs: add Fortran options to Linux install instructions

jtniehof:
Small docs update to list the Fortran options in Linux (as it is on Mac and used to be in the generic instructions.)


drsteve:
:+1:

PR #167, opened 2019-06-19
Release prep

jtniehof:
A bit of a mishmash PR but it's all pertaining to release prep.

- Update the DISTRIBUTE document which describes making a release
- PyPI no longer hosts Windows standalone installers, so reference GitHub for those
- pythonhosted is dead so move documentation to github as well. We can always put more of "home page" there later and put the docs in a subdirectory.

### DIFF
--- a/DISTRIBUTE
+++ b/DISTRIBUTE
@@ -57,6 +57,8 @@ Test PyPI
 
 https://packaging.python.org/guides/using-testpypi/
 
+Consider https://www.python.org/dev/peps/pep-0440/#pre-releases
+
 Make a .pypirc file, see https://docs.python.org/3.3/distutils/packageindex.html
 
 [distutils]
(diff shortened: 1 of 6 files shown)


jtniehof:
Few more notes on DISTRIBUTE.

pip doesn't fall back to older versions of dependencies if there's something conflicting. So for instance latest matplotlib requires Python 3.6 so pip install spacepy on Python 3.5 will fail unless you install matplotlib some other way. Notes added on that (as a "try this").

drsteve:
Nice work on getting the docs up on github.io so quickly!
All looks good. I just found a typo _I just couldn't live with_ and fixed it. (face --> fact)

I've annoyed Travis once already today by merging before it finished its tests, so I'll merge this in 20 minutes or so.

drsteve:
Sorry, forgot I hadn't merged this and went to sleep last night!!

PR #168, opened 2019-06-20
Mac install docs

balarsen:
This is requested updates the the mac install docs. Not perfect but I verified these to work which is more important. 

Thoughts? Edits? I don't have a ton of time to edit so others feel free.

### DIFF
--- a/Doc/source/install_mac.rst
+++ b/Doc/source/install_mac.rst
@@ -2,11 +2,11 @@
 MacOS Installation
 ******************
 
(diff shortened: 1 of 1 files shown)


drsteve:
Editing other people's commits sucks...

Anyway, Travis is running, I'm going to merge anyway because if it passed a moment ago, then it'll pass now with minor changes to the docs. Thanks for the update. We can/should flesh this out a bit later, but this is good for now. It gets the compilers installed and the rest is straightforward.

PR #176, opened 2019-06-24
datamodel.py: cleanup in the py2 vs py3 string handling, was handled …

balarsen:
datamodel.py: cleanup in the py2 vs py3 string handling, was handled well is some places but not uniformly.
    - updated the test suite and made JSONTests.setUp() find the data files no matter where the test was run from

These are minor updates prepping for work on #9 but will need to come back to that.

### DIFF
--- a/spacepy/datamodel.py
+++ b/spacepy/datamodel.py
@@ -181,6 +181,7 @@ from . import time as spt
 
 __contact__ = 'Steve Morley, smorley@lanl.gov'
 
+# python2 python3 string wrangling
 try:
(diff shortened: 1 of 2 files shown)


drsteve:
That `unicode` assignment makes me queasy. (line 39 in `test_datamodel.py`)
It's a typo away from overloading a standard library function...

It's reasonable enough, and while I'd rather see that fixed by a "cast to string" function (probably with a try ... except for the different behaviours in *_that_* function), this will work. I'll push my aesthetic qualms to one side and merge...

PR #177, opened 2019-06-24
test_pybats.py: Fixed an issue where pytest failed, FileNotFoundError

balarsen:
Fixed an issue where pytest from the the spacepy root directory failed on pybats due to not finding the files (pytest tests/test_pybats.py). Fox was to use:
        `pth = os.path.dirname(os.path.abspath(__file__))`
and    `os.path.join(pth, ...)`

### DIFF
--- a/tests/test_pybats.py
+++ b/tests/test_pybats.py
@@ -10,6 +10,7 @@ import matplotlib
 matplotlib.use('Agg')
 
 import datetime as dt
+import os
 import unittest
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


drsteve:
Rather than defining `pth` again every time it's needed, can you define it before the tests? An alternative would be to define it once per test class, in the setup method (rather than in test functions).

balarsen:
Good idea, I will update.

balarsen:
Close this in favor of #180

PR #182, opened 2019-06-24
toolbox: update to close #181. The num parameters are explicitly cast to int

### LINKED ISSUE #181: Toolbox DeprecationWarning: object of type <class 'numpy.float64'> cannot be safely interpreted as an integer
`toolbox.linspace()` produces a DeprecationWarning

Fix is likely to make the number input an int. Likely also an issue for `logspace`.

```
In [1]: import spacepy.toolbox as tb
In [2]: tb.linspace(1, 4, 3.0)
/Users/balarsen/git/spacepy/spacepy/toolbox/__init__.py:1310: DeprecationWarning: object of type <class 'float'> cannot be safely interpreted as an integer.
  return np.linspace(min, max, num, **kwargs)
Out[2]: array([1. , 2.5, 4. ])
```

### Error message/Traceback:
```
/Users/balarsen/git/spacepy/spacepy/toolbox/__init__.py:1310: DeprecationWarning: object of type <class 'float'> cannot be safely interpreted as an integer.
  return np.linspace(min, max, num, **kwargs)
```

```
Darwin-18.6.0-x86_64-i386-64bit
sys.version_info(major=3, minor=7, micro=3, releaselevel='final', serial=0)
numpy=1.16.4
scipy=1.3.0
matplotlib=3.1.0
```

balarsen:
This will close #181. The fix is a simple cast to int for the `num` parameter in `linspace` and `logspace`.

### DIFF
--- a/spacepy/plot/spectrogram.py
+++ b/spacepy/plot/spectrogram.py
@@ -215,10 +215,10 @@ class spectrogram(dm.SpaceData):
                 # use the toolbox version of linspace so it works on dates
                 self.specSettings['bins'] = [dm.dmarray(tb.linspace(self.specSettings['xlim'][0],
                                                  self.specSettings['xlim'][1],
-                                                 np.sqrt(len(data[self.specSettings['variables'][0]])))),
+                                                 int(np.sqrt(len(data[self.specSettings['variables'][0]]))))),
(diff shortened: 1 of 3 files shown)


balarsen:
See comments in #181, reverted to not cast in linspace/logspace but in spectrogram

Issue #184, opened 2019-06-25
Toolbox.quaternionNormalize first/last error


balarsen:
toolbox.quaternionNormalize gives the same result for `scalarPos='first'` and `scalarPos='last'` this should not be the case. 

### Minimal example to reproduce issue:
```
In [1]: import spacepy.toolbox as tb
This unreleased version of SpacePy is not supported by the SpacePy team.

In [2]: tb.quaternionNormalize([0.707, 0, 0.707, 0.2])
Out[2]: array([0.69337122, 0.        , 0.69337122, 0.19614462])

In [3]: tb.quaternionNormalize([0.707, 0, 0.707, 0.2], scalarPos='first')
Out[3]: array([0.69337122, 0.        , 0.69337122, 0.19614462])

In [4]: tb.quaternionNormalize([0.707, 0, 0.707, 0.2], scalarPos='last')
Out[4]: array([0.69337122, 0.        , 0.69337122, 0.19614462])
```

drsteve:
See discussion in #268.
@balarsen - I'll close this in the near future unless I hear objections. The TL;DR version is that normalization basically divides each element by the norm and therefore you don't expect to see a difference depending on the scalar position.

balarsen:
I am a bit confused here, why do we have `scalarPos` if it doesn't do anything?

balarsen:
If this is a non-issue then close as invalid but it seemed funny to me.

jtniehof:
As described in #268, it only makes a difference when constructing a scalar quaternion to avoid degeneracy:
```
In [3]: spacepy.toolbox.quaternionNormalize([1e-20, 1e-20, 0, 0], scalarPos='first')
   ...: 
Out[3]: array([ 1.,  0.,  0.,  0.])

In [4]: spacepy.toolbox.quaternionNormalize([1e-20, 1e-20, 0, 0], scalarPos='last')
Out[4]: array([ 0.,  0.,  0.,  1.])
```

drsteve:
I agree that it's not necessarily obvious, but it is correct behaviour. As @jtniehof just said, the `scalarPos` argument is so that the identity quaternion is constructed correctly when the quaternions have a very small magnitude. Most of the time it makes no difference.

Now that we've aired this out I'll close.

PR #185, opened 2019-06-25
 test_toolbox.py: updated for more test coverage

balarsen:
Minor updates to test_toolbox.py for more test coverage. This uncovered  #184.


balarsen:
Well the fail is bad... I don't understand the error but I bet that `indsFromXrange` doesn't actually work...

balarsen:
Sorted, range vs xrange thing. \

PR #188, opened 2019-07-15
poppy.boots_ci: fix handling of multiple return values in C version

jtniehof:
poppy.boots_ci was failing (C version only) if the aggregating function returned multiple values. This PR adds a test and fixes this.

In the case of multiple returns, the result isn't sorted, since there isn't one sort order. This shouldn't affect the results; I think the sorting was largely an efficiency measure

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,7 @@
 Changes in Version 0.2.1 (2019-xx-xx)
 =====================================
+poppy
+ - Fix bootstrapping when aggregating function has multiple outputs
 
 Changes in Version 0.2.0 (2019-06-22)
 =====================================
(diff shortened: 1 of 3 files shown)


jtniehof:
I originally had them as separate tests but for some reason a little voice suggested combining them and there was some reason but of course I don't recall why now (lots of tests but one idea?) I'd rather not put in the work of breaking them at this point.

At some point we should probably have standards for test length but I've found it awfully hard to really have rules that are clear and always make sense.

drsteve:
Understood. My preference is for tests where if something fails, then I don't have to work through and understand that failure before I see what else in the test is failing (if that makes sense). How do you make that a clear rule? No idea.

Merging now.

PR #189, opened 2019-07-18
Bootstrap-based histogram with error bars

jtniehof:
This PR adds a toolbox function to calculate bootstrap-based confidence intervals on histograms (and optionally plot.) It's nothing earthshattering but I had a student who wasn't putting error bars on a histo and figured it was worth making easy.

### DIFF
--- a/Doc/source/toolbox.rst
+++ b/Doc/source/toolbox.rst
@@ -52,6 +52,7 @@ Other functions
     :toctree: autosummary
 
     assemble
+    bootHisto
     dictree
     eventTimer
     feq
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -1211,7 +1211,7 @@ def binHisto(data, verbose=False):
             print("Used F-D rule")
     return (binw, nbins)
 
-def bootHisto(data, inter=90., n=100, seed=None, plot=False, **kwargs):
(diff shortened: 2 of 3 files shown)


jtniehof:
- No problem bumping n. Will do that.
- For `set_target`, your suggestion is to axe the `plot` kwarg in favour of basically passing through the `set_target` kwargs and then drawing on the return? How should we specify not to plot? Setting `target` to `None` would always plot. I like the idea of trying to match API though.
- It's in tb by analogy with `binHisto` and then the plotting was a bit of an afterthought. I'm not sure plot is the _right_ place for it but I agree tb isn't necessarily either (and tb is a bit of a dumping ground.)

drsteve:
For `set_target` I was originally thinking _in addition_ to the `plot` kwarg. So `plot` would be Boolean and `set_target` would get called only if `plot` was `True`.

If multiple kwargs are annoying then it could use the same calling syntax as now, but pass the not-equal-to-False value handed to `plot` to `set_target` so at least it's being used internally.

The issue is that `set_target` always expects to have a target to plot to, and wasn't designed to have an option for not returning plotting objects. So other thoughts on, for example, using `False` to specify no plot and `None` to get the default figure and axes seem to be pointless.

I'm open to other suggestions...

jtniehof:
Okay, updated. Note that the docs build _terribly_ because numpydoc basically breaks cross-referencing. Open to correction on this front but it's why I want to ultimately get off numpydoc.

PR #192, opened 2019-08-06
Access "attrs" via "meta"

jtniehof:
This PR adds access to the "attrs" attributes on dmarray, SpaceData, and CDFs (plus CDF variables) via a "meta" property.

This was a recommendation of the pyhelio conference in Boulder in November 2018 (strong recommendation was to use one or the other, weak to use both just to make it easier to find.) There was also a recommendation to have "data" return the raw data without any fanciness, but since we subclass dmarray `.data` already has meaning (it's basically the memory underlying the array.)

Incorporating this PR basically helps us play nicer in the broader ecosystem, just making it easier for people to find things. But it does keep us from using "meta" for something else in the future. (E.g. we could in theory leave "attrs" as having no interpretation, like now, and use "meta" for metadata where we define the meaning, potentially parsed from "attrs".)


jtniehof:
@drsteve , I pulled this up since you had noticed the branch languishing. I'm okay ditching it if we think it's a bad idea, but if we don't have a reason not, it will help play nice.

PR #194, opened 2019-08-13
Short doc on pythonic code

jtniehof:
This has come up in the pyhelio discussions of "hey, what's Pythonic code" and I figure we can help provide an answer. There's some overlap with the tips document but this is an attempt to explain why (and then link out to the exhaustive lists of do-this-not-that.)

This can definitely be improved so I'm happy to iterate on the branch for awhile, or merge and improve later.

### DIFF
--- a/Doc/source/index.rst
+++ b/Doc/source/index.rst
@@ -63,6 +63,7 @@ SpacePy Documents
     configuration
     case_studies
     publications
+    pythonic
 
 SpacePy Code
 ============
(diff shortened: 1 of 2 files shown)


spacecataz:
If we do that (which I support), I strongly request that we decide on a
docstring format and include docstring formatting in the style guide.  Plz.

-dw

jtniehof:
I've opened #200. I think it's a large enough issue that I'd rather merge this as-is now and then work on the style guide as another issue.

PR #195, opened 2019-08-13
pycdf: convert array scalers to Python types (Closes #123)

### LINKED ISSUE #123: pycdf scalar indexing returning array scalar
This shows how little I've been doing single scalar indexing...indexing a variable with a scalar returns a numpy array scalar rather than a single value:
```
In [7]: f['MET'][0]
Out[7]: array(291859185)
```

I'm not sure if this crept in somewhere or has been the case since the conversion to numpy, or even if it's necessarily undesirable, but definitely do need to check if it's a regression (a good one for people wanting to learn git bisect!) and, if so, fix and add a test.

If it's not a regression, question is if we're okay with this.

jtniehof:
Right now reading numerical types from a CDF, e.g. `f['MET'][0]`, returns an array scalar. This isn't true for Epochs (which return a single `datetime`) and only seems to be sometimes true for strings.

For consistency I think it makes sense to always have Python objects, and that's what this PR does. Only one test fails with the conversion and it's fixed to the new behavior in this version. So I think the array scalar behavior isn't something we've considered important enough to test.

I can also see an argument to be made for always returning array scalars regardless.


jtniehof:
Poke...anything that needs to be done here?

PR #199, opened 2019-08-16
pycdf: type guessing on empty numpy object arrays (Closes #58)

### LINKED ISSUE #58: Issue with pycdf.CDF.from_data, NRV?
This is the second issue, that after speaking with Jon, appears to be separate from [the other issue I opened.](https://github.com/spacepy/spacepy/issues/57)

```python
In [5]: pycdf.lib.set_backward(False)

In [8]: pycdf.CDF.from_data('test_cdf.cdf', d)
IndexError                                Traceback (most recent call last)
<ipython-input-8-916e03bc5ee8> in <module>()
----> 1 pycdf.CDF.from_data('test_cdf.cdf', d)

/usr/local/lib/python2.7/dist-packages/spacepy/pycdf/__init__.pyc in from_data(cls, filename, sd)
   1767         with cls(filename, '') as cdffile:
   1768             for k in sd:
-> 1769                 cdffile[k] = sd[k]
   1770             cdffile.attrs.clone(sd.attrs)
   1771

/usr/local/lib/python2.7/dist-packages/spacepy/pycdf/__init__.pyc in __setitem__(self, name, data)
   1597                 self[name].attrs.clone(data.attrs)
   1598         else:
-> 1599             self.new(name, data)
   1600
   1601     def __iter__(self, current = 0):

/usr/local/lib/python2.7/dist-packages/spacepy/pycdf/__init__.pyc in new(self, name, data, type, recVary, dimVarys, dims, n_elements, compress, compress_param)
   2090                 n_elements = 1
   2091         else:
-> 2092             (guess_dims, guess_types, guess_elements) = _Hyperslice.types(data)
   2093             if dims is None:
   2094                 if recVary:

/usr/local/lib/python2.7/dist-packages/spacepy/pycdf/__init__.pyc in types(data, backward)
   3589         types = []
   3590
-> 3591         _Hyperslice.check_well_formed(d)
   3592         if d.dtype.kind in ('S', 'U'): #it's a string
   3593             types = [const.CDF_CHAR, const.CDF_UCHAR]

/usr/local/lib/python2.7/dist-packages/spacepy/pycdf/__init__.pyc in check_well_formed(data)
   3523         if d.dtype == numpy.object: #this is probably going to be bad
   3524             try:
-> 3525                 len(d.flat[0])
   3526             except TypeError: #at least it's not a list
   3527                 pass

IndexError: index 0 is out of bounds for axis 0 with size 0
```

Jon mentioned this likely has to do with confusing RV and NRV things.

jtniehof:
This PR handles creating a variable when the data passed in is a zero-length object array and no type specified. In this case there's no reasonable type to guess so it throws an error. Still handles zero-length arrays of numerical type.

This will still throw an error on the case of #58, but it will be an error that makes more sense. Root cause would be to include type information in VarCopy objects (an enhancement concept that's been floating around for awhile and I have other reasons for pursuing in the short term.)


jtniehof:
Poke.

Note when I say it will throw an error on #58 and the root cause would be addressed differently, that's only because I have knowledge of the greater code ecosystem around #58. This is the fix for the _bug_ but there's a nice _enhancement_ possibility which would be useful in that situation. It wouldn't remove the need for this PR.

PR #204, opened 2019-09-20
Actual argument contains too few elements for dummy argument 'maginpu…

xredor:
…t' fixed in LAndI2Lstar.f. Also chmod -x for fortran source files made

When compiling with gcc (Debian 8.3.0-6) 8.3.0 the error
```
LAndI2Lstar.f:20680:19:

      &             maginput_tmp,Lm_tmp,XJ_tmp,Lstar_tmp)
                   1
Error: Actual argument contains too few elements for dummy argument ‘maginput’ (25/2500000) at (1)
```
happen.

This pull request fixes it.


jtniehof:
Thanks for the input. @drsteve is the real irbem expert here and is involved in the upstream project, so hopefully he'll be able to comment. i do think this is probably best sent upstream to https://sourceforge.net/p/irbem/bugs/ and then we'll incorporate when it's fixed there.

With my very limited Fortran knowledge, I'm somewhat uneasy about redimensioning the arguments to EmpiricalLstar1 (I don't think we actually use it in spacepy, so tests wouldn't necessarily catch problems).

The permissions on the source are indeed weird but it matches upstream, so again I think that's a see if it can be fixed there.

Hopefully Steve can take a look but didn't want to just leave this hanging.

drsteve:
I'll run this by some folks who are more central to IRBEM development. While I have commit access over there, there's no continuous integration so I'd like to get this out to active IRBEM developers to test and comment on.

AntoineBrunet:
Thank you for this feedback, we have noticed this bug in IRBEM with recent compilers. Our temporary fix was to set the `-std=legacy` flag at compilation. We see this bug as a good opportunity to fix globally the array parameters practices in IRBEM subroutines. In particular, we could lift the arrays sizes constraints almost everywhere. We will get back to you when we have a working version of IRBEM so it can be updated in spacepy.

jtniehof:
Thanks Antoine. @drsteve , we've had a few std=legacy things come up. Do we want to enable it on release builds only? (So anybody working from git would help find these issues, but the release is more likely to build properly.) Has it been in gfortran since basically forever, or would we need to do a dance of figuring out which compiler version we're using and if it supports the flag?

jtniehof:
@xredor , if you don't mind compiling from a PR, can you try building from #214 and see if this fixes the problem? We weren't consistently applying `std=legacy` and that PR should improve it.

jtniehof:
0.2.1 was released with `std=legacy` as the default, so can you check if you can install straight from pip. Long-term this should still be fixed.

jtniehof:
@drsteve , did #295 include a fix from upstream?

drsteve:
To my knowledge, this hasn't been addressed upstream. I haven't seen any activity on the source directory of the IRBEM repo (except my pushes) in a long time.
I guess I was assuming that this was on @AntoineBrunet's list to fix in IRBEM, possibly as part of a larger modernization effort. And since it compiles with the legacy flags I wasn't thinking about it...

jtniehof:
I'm guessing we're probably not going to merge this in anything resembling its current form, the issue in SpacePy has been addressed, and it's on Antoine's radar. So I'm inclined to close this unmerged unless anybody feels there's a better way to capture what's going on.

jtniehof:
Since the underlying issue is upstream, and we're using appropriate flags to compile with the code as it is, I'm going to close this one unmerged. Thanks for the submission; probably wouldn't have gotten #214 without it.

PR #206, opened 2019-09-25
pycdf: put CDF-specific metadata in VarCopy

jtniehof:
This PR adds information about original CDF type, record and dimension variance, compression, etc. to VarCopy so that new CDF variables made from it will match.

I thought about exposing the information as in the Var objects but there are two problems. One is that the `compress` method already exists in ndarray/dmarray so including it sort of messes things up. I suppose this is workable by a super call if any args are called. The other problem is that things can get a little bit weird if you do really bizarre stuff and then expect it to still work...so for instance slicing out a dimension doesn't update the dimension variance. This isn't trying to be perfect, just carry some information.

I'm a little conflicted on this, as it would be nice to be able to have the same interface for Var and VarCopy for stuff like checking record variance.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,7 @@
 Changes in Version 0.2.2 (2019-xx-xx)
 =====================================
+pycdf
+ - VarCopy objects now carry information to help reproduce original
 
 Changes in Version 0.2.1 (2019-10-02)
 =====================================
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@ Changes in Version 0.2.2 (2019-xx-xx)
 =====================================
 pycdf
  - VarCopy objects now carry information to help reproduce original
+ - Newly-public method nelems to get number of elements in zVar
 
 Changes in Version 0.2.1 (2019-10-02)
 =====================================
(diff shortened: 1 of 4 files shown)


jtniehof:
Actually just realized a fairly major win that could come from sticking with the same interface, so will rework that and push again in a bit.

jtniehof:
Rebased rather than merging. Going to fandango on this a bit more.

jtniehof:
I think this is good to merge now. The VarCopy interface is very similar to the Var interface. There's probably some minor cleanup that can be done between __setitem__, clone, and new, but those have some interesting subtleties and that sort of cleanup isn't necessary to get benefit from this.

jtniehof:
This looks like the numpy website went down in the middle of trying to run the CI... @drsteve , any way to tickle it to try again?

drsteve:
Job restarted.

jtniehof:
@balarsen , @drsteve , thoughts? This is the point where I can start buliding something fairly cool on top (and am on a tight schedule to do so, insert whine here.)

jtniehof:
Rebased against current master.

jtniehof:
Rebased against new master.

jtniehof:
@drsteve , if you've got time now to evaluate more comprehensively, it'd help...looks like @balarsen isn't chiming in. I've got another big drop coming soon....

balarsen:
I'm happy.

PR #209, opened 2019-09-26
irbempy: Document negative L in Lm, Lstar (Closes #104)

### LINKED ISSUE #104: Negative return values in irbempy (get_Lstar, get_Lm) are not documented
I am trying to use get_Lstar to calculate the Lstar for a magnetometer station in GDZ coordinates, but the function will not give me an Lstar value, and the returned Lm value is unphysical. Here is my code

```
from spacepy.time import Ticktock
from spacepy.coordinates import Coords
from spacepy.irbempy import get_Lstar

t = Ticktock(success['Time'][0])
y = Coords([100,56.376,-96.3], 'GDZ', 'sph',units=['km','deg','deg'])
g=get_Lstar(t,y)
```
Here, we have `success['Time'][0] = Timestamp('2008-01-01 08:00:00')`

However, the output of the function is

```
{'Lm': array([[-6.44920639]]),
 'Lstar': array([[nan]]),
 'Bmirr': array([[56384.77608068]]),
 'Bmin': array([70.56354592]),
 'Xj': array([[15.20358754]]),
 'MLT': array([1.24258515])}
```

The function returns `nan` for whatever timestamp I put in, and the Lm value is always negative which is unphysical. I've looked at the source code for Coords and I believe the input coordinates are fine, so is there something else that's going wrong or have I made an input error?
### Error message/Traceback:

```

```

jtniehof:
Docstring update explaining negative L values.

### DIFF
--- a/spacepy/irbempy/irbempy.py
+++ b/spacepy/irbempy/irbempy.py
@@ -1495,9 +1495,12 @@ def get_Lm(ticks, loci, alpha, extMag='T01STORM', intMag='IGRF', IGRFset=0, omni
 
     Returns
     =======
+
         - results (dictionary) : containing keys: Lm, Bmin, Blocal (or Bmirr), Xj, MLT 
             if pitch angles provided in "alpha" then drift shells are calculated and "Bmirr" 
-            is returned if not provided, then "Blocal" at spacecraft is returned.
+            is returned if not provided, then "Blocal" at spacecraft is returned. A negative
+            value for Lm indicates the field line is closed but particles are lost to the
+            atmosphere; the absolute value indicates the L value.
 
     Examples
     ========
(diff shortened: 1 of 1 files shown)


drsteve:
Just thinking about the phrasing in here. First, thanks for doing this. Second, I think we can be more specific than just saying "not trapped". I'll think about it and maybe submit an edit later today or tomorrow.

jtniehof:
Sure! Field lines are closed (else there wouldn't be even a negative L....) but particles are lost to the atmosphere?

drsteve:
Yes, pretty much that. Also the reference in Lm to "drift shell" should probably be changed to just "L value" or something as Lm is a particle-specific field-line quantity. If you get to it before I do (seems likely) then I have no problem with that either!

jtniehof:
Okay, updated and rebased to current master, looks messy in the PR but clean in the repo.

Issue #190, opened 2019-08-05
Issue installing/compiling IRBEMlib


grimcreeper97:
Greetings.

When trying to install Spacepy on OS X, it fails when trying to compile IRBEMlib.
Note that all requisites were installed via MacPorts.

### Minimal example to reproduce issue:
```
python setup.py build
```

### Error message/Traceback:
```
running build
running config_fc
unifing config_fc, config, build_clib, build_ext, build commands --fcompiler options
running config_cc
unifing config_cc, config, build_clib, build_ext, build commands --compiler options
running build_py
customize Gnu95FCompiler
Found executable /opt/local/bin/gfortran
Found executable /opt/local/bin/ar
Found executable /opt/local/bin/ranlib
Reading fortran codes...
        Reading file 'source/onera_desp_lib.f' (format:fix,strict)
        Reading file 'source/wrappers.inc' (format:fix)
        Reading file 'source/fortran_version.inc' (format:fix)
        Reading file 'source/wrappers.inc' (format:fix)
        Reading file 'source/fortran_release.inc' (format:fix)
        Reading file 'source/wrappers.inc' (format:fix)
        Reading file 'source/ntime_max.inc' (format:fix)
        Reading file 'source/wrappers.inc' (format:fix)
        Reading file 'source/variables.inc' (format:fix)
        Reading file 'source/ntime_max.inc' (format:fix)
        Reading file 'source/wrappers.inc' (format:fix)
Traceback (most recent call last):
  File "/opt/local/bin/f2py", line 11, in <module>
    load_entry_point('numpy==1.16.4', 'console_scripts', 'f2py')()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/numpy/f2py/f2py2e.py", line 685, in main
    run_main(sys.argv[1:])
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/numpy/f2py/f2py2e.py", line 425, in run_main
    postlist = callcrackfortran(files, options)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/numpy/f2py/f2py2e.py", line 329, in callcrackfortran
    postlist = crackfortran.crackfortran(files)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/numpy/f2py/crackfortran.py", line 3251, in crackfortran
[...]

PR #214, opened 2019-10-02
setup: use std=legacy for f2py on gnu95

jtniehof:
We were using `std=legacy` for the actual irbem compile on gnu95 but not for the f2py calls. This caused issues e.g. #190 and #204.

This PR adds std=legacy to the f2py calls when using gnu95 in hopes of avoiding these issues in the future.

Ideally we do want to fix irbem code to work without `std=legacy` but in the meantime this PR 1) makes us consistent between library compile and f2py and 2) keeps things ticking along.


jtniehof:
I've made this a blocker for 0.2.1 since so many issues lately are related to it.

Issue #218, opened 2019-10-04
OMNI2 zipfile contents not cleaned up


asreimer:
When updating spacepy files, an omni2 zip file is downloaded. This zip file is then extracted and some things are done with the extracted contents. Finally, the zip file is cleaned up, however the extracted contents are not.

Using spacepy 0.2.1, for example, executing the following:

    python -c "import spacepy.toolbox; spacepy.toolbox.update(omni2=True)"

yields a file "OMNI_OMNI2_merged_20120213-v1.cdf" in the current working directory 

I believe the issue is that an `os.remove` is missing here: https://github.com/spacepy/spacepy/blob/master/spacepy/toolbox/__init__.py#L902-L903

I think that the fix is simple, something like this:

    file_to_convert = fh_zip.namelist()[0]
    omnicdf = fromCDF(file_to_convert)
    #add RDT
    omnicdf['RDT'] = spt.Ticktock(omnicdf['Epoch'],'UTC').RDT
    #remove keys that get in the way
    del omnicdf['Hour']
    del omnicdf['Year']
    del omnicdf['Decimal_Day']
    # save as HDF5
    toHDF5(omni2_fname_h5, omnicdf)
    # delete left-overs
    os.remove(omni2_fname_zip)
    os.remove(file_to_convert)

to replaces [these lines](https://github.com/spacepy/spacepy/blob/master/spacepy/toolbox/__init__.py#L891-L903) seems to work. I'll submit a PR shortly.

PR #219, opened 2019-10-04
[BUGFIX]: clean up extracted file

asreimer:
This fixes https://github.com/spacepy/spacepy/issues/218, where it was found that the extracted zip file contents for OMNI2 data download were not cleaned up.

Testing this PR can be done as follows:
1) using the `master` branch, run the command `python -c "import spacepy.toolbox; spacepy.toolbox.update(omni2=True)"` and verify that a file named `OMNI_OMNI2_merged_20120213-v1.cdf` appears, but isn't cleaned up
2) using this PR, run the same command as above and notice that the file is cleaned up

This works in both python 2 and python 3.

### DIFF
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -888,8 +888,8 @@ def update(all=True, QDomni=False, omni=False, omni2=False, leapsecs=False, PSDd
         fh_zip = zipfile.ZipFile(omni2_fname_zip)
         fh_zip.extractall();
         fh_zip.close()
-        file_to_convert = fh_zip.namelist()[0]
-        omnicdf = fromCDF(file_to_convert)
+        file_to_read = fh_zip.namelist()[0]
+        omnicdf = fromCDF(file_to_read)
         #add RDT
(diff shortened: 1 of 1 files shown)


jtniehof:
Thanks for finding this; I was wondering why those OMNI cdfs occasionally show up. At the risk of letting the perfect be the enemy of the good, I wonder if this shouldn't be the opportunity for a little more cleanup. The extract should probably be into a temporary directory and we should probably explicitly extract just the first file in the zip (with check for horrible stuff like having paths in the ZIP). Happy to push an update that does this to your branch.

asreimer:
Hey @jtniehof, if this PR fixes the issue adequately, but more clean up might be useful, I would merge this and then make a new PR that cleans things up. Bite-sized, quick, and simple. You might find that a seemingly simple clean up takes more effort than anticipated.

A note on extracting files from the zip, it looks to me like only the first file in the zip is extracted: `fh_zip.namelist()[0]` hence the `[0]`, but I agree it might be nice to extract to a temporary directory. The built-in [tempfile](https://docs.python.org/3/library/tempfile.html) module is nice for this.

jtniehof:
Yeah, might as well do a two-step here. Thanks.

PR #221, opened 2019-10-11
pycdf.Attr: Fix writing an EPOCH16 Entry in raw mode

jtniehof:
This PR fixes a rather esoteric bug. pycdf accepts writing of "raw" values, most commonly converted Epoch values rather than datetime. With EPOCH16, each value in the CDF requires two numbers on the input, picking up an extra axis of length 2 (in numpy parlance). Most of this is handled properly but, in writing to an attribute, the extra axis wasn't appropriately handled and twice as much data was written.

This fixes the bug. The new test case should be a pretty good illustration of the problem.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,7 @@ Changes in Version 0.2.2 (2019-xx-xx)
 pycdf
  - VarCopy objects now carry information to help reproduce original
  - Newly-public method nelems to get number of elements in zVar
+ - Fix assignment of "raw" values to EPOCH16 Attribute
 
 Changes in Version 0.2.1 (2019-10-02)
 =====================================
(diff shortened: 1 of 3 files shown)


jtniehof:
Isn't is crazy? I'll give a little time for others to chime in but the big dump I'm working on will need this.

jtniehof:
@balarsen or @drsteve , do you feel comfortable punching merge? I need to open another PR that depends on this (and so the rebase would be epic if the merge happens after.)

jtniehof:
It manifests when cloning attributes from one variable to another in raw mode, which I use pretty typically when copying between CDFs (it's marginally faster for not doing the datetime conversions.)

PR #222, opened 2019-10-14
pycdf istp support for copying/slicing variables and dependencies

jtniehof:
This PR adds support for a VarBundle class (other names for the class welcome.) This makes it easy to copy one variable and all its attributes and dependencies from one CDF into another. In addition to a straight copy, it's possible to slice and sum or average over a dimension. All the dependencies are then also updated.

There are examples scattered around the documentation, but for a flavor (assume `incdf` is a CDF open for reading, and `outcdf` for writing):

```
#Copy HOPE fluxes to new CDF, averaged over pitch angle and only keeping low energies
b = VarBundle(incdf['FPDU'])
b.mean(1).slice(2, 0, 36).output(outcdf)

#Do the same but also average out into two variables, high and low energy
b = VarBundle(incdf['FPDU'])
b.mean(1).slice(2, 0, 36).mean(2).output(outcdf, suffix=_'Lo')
#The mean is persistent
b.slice(2, 36).output(outcdf, suffix='_Hi')
```

This isn't the end state of this code--there are many enhancements I'd like to make--but this PR is big enough, and it's a useful bit of functionality in its own. And I'd welcome other ideas for enhancements in issues.


jtniehof:
Interesting. I had run tests on Python 3 but clearly missed something.

jtniehof:
Some notes that are in the documentation but may not be clear in there (feedback on clarity is welcome):

The method chaining is for convenience only. Unlike, say, sqlalchemy, there's no difference between:
`b.slice(1, 0, 10).mean(1)`
and:
```
b.slice(1, 0, 10)
b.mean(1)
```
The bundle is modified in place and returned, not a modified one returned. I'm certainly open to changing this.

New slices on the same dimension will replace the old slice entirely. `sum` and `mean` are not compatible and will error if you try both on a particular dimension. I suppose I could have them override each other (sort of like successive slicing does.) It's easy to reset the slice on an dimension: `slice(dim)`. But there's no way to "undo" a sum or mean. Maybe a `reset` method? Getting a new bundle is computationally cheap so maybe there's no reason. I can also see having `slice()` as a way to reset the slicing on all dimensions.

slices always happen before mean/sum. dimension numbers don't change with successive operations. Combined, this means the order of method calls doesn't matter (except `output` of course.)

balarsen:
Super interesting. I won't get to it today but I will break this for you and provide comments :)

jtniehof:
Yeah, I imagine it'll take awhile to evaluate. And I welcome the breakage.

jtniehof:
I've been using this pretty heavily and it's holding up okay. What's a good path to merging? I understand there's no way to really beat on it in a short time and I do expect there will be issues opened to fix.

If there's a feeling that substantial API changes are needed then that should probably be pre-merge...but would like to make those soon since I'm starting to rely on this code.

balarsen:
I'll make a little wish list here:
__repr__ for istp.VarBundle could be helped. 

for a cdf variable you get:
```
In [13]: cdf['FEDU']
Out[13]:
<Var:
CDF_REAL4 [4173, 5, 16, 72]
>
```
for a VarBudle you get:
```
In [10]: istp.VarBundle(cdf['FEDU'])
Out[10]: <spacepy.pycdf.istp.VarBundle at 0x7fce30c452d0>
```

Maybe better would be a list of the names of the variables inside.

balarsen:
Ok, this is awesome!  One more suggestion here. 

This words like a charm and seem exactly useful. 

```
vb = istp.VarBundle(cdf['FEDU'])
cdf2 = pycdf.CDF('t1.cdf', '')
vb.output(cdf2)
cdf2.close()
cdf2 = pycdf.CDF('t1.cdf')
cdf2
<CDF:
Detector_LABL: CDF_CHAR*2 [5] NRV
ENERGY_Ele_DELTA: CDF_REAL4 [4173, 72]
Energy_LABL: CDF_CHAR*3 [72] NRV
Epoch_Ele: CDF_EPOCH [4173]
Epoch_Ele_DELTA: CDF_REAL4 [4173]
FEDU: CDF_REAL4 [4173, 5, 16, 72]
HOPE_DETECTOR: CDF_INT1 [5] NRV
HOPE_ENERGY_Ele: CDF_REAL4 [4173, 72]
HOPE_SECTOR: CDF_INT1 [16] NRV
Sector_LABL: CDF_CHAR*3 [16] NRV
>
```

Suggestion:
What I cannot sort out is how in vb above do I know what variables it has in it?
The only objects are:
```
 'cdf',
 'mainvar',
 'mean',
 'output',
 'slice',
 'sum'
```
cdf points back to where it came from, mainvar is the input one (FEDU), how do I access the the rest or get a list of them?

jtniehof:
Rebased.

Agreed a nice str/repr would be good. Related to the other, actually...

The dominant "feature" of the VarBundle is the main variable. In theory you shouldn't have to worry about the other variables, but I can see in practice it's helpful to know. The one caveat is that not all of those variables necessarily come onto the output...it would be MUCH more difficult to represent "the variables that will be written if you call `output`" than "all the variables that I'm currently tracking." That can be noted on the documentation, of course. And gets to whether the repr should have just the main var or all of the associated variables. It also shows that the "VarBundle" name is a little misleading.

Both of those can be added either before or after merge; I will get to work on them in a bit.

I can fix the #224 issue that is actually in this PR before merging; the other I think is in master and will be a separate PR.

balarsen:
> Rebased.
> 
> Agreed a nice str/repr would be good. Related to the other, actually...
> 
> The dominant "feature" of the VarBundle is the main variable. In theory you shouldn't have to worry about the other variables, but I can see in practice it's helpful to know. The one caveat is that not all of those variables necessarily come onto the output...it would be MUCH more difficult to represent "the variables that will be written if you call `output`" than "all the variables that I'm currently tracking." That can be noted on the documentation, of course. And gets to whether the repr should have just the main var or all of the associated variables. It also shows that the "VarBundle" name is a little misleading.
> 
> Both of those can be added either before or after merge; I will get to work on them in a bit.
> 
> I can fix the #224 issue that is actually in this PR before merging; the other I think is in master and will be a separate PR.

OK, I feel like this is good to go for a merge, so poke me after the #224 update and I will merge.

jtniehof:
Pushed a new commit that fixes the other half of #224.

Probably makes most sense to merge #225 first; then I can rebase this again if there are any conflicts.

jtniehof:
Once merged I'll open up the repr and variable information as an issue/enhancement and work it there. (I've got a bunch more enhancements I'll open up as well.)

jtniehof:
Rebased given the merge of #225 ...probably not needed since it was a clean rebase, but forces the tests just to be sure.

jtniehof:
Poke @balarsen ...ready to merge here.

PR #226, opened 2019-10-21
pycdf.istp.VariableChecks: Check shape of DELTA variables.

jtniehof:
This PR adds to the DELTA checks by comparing the shape of the DELTA to its parent variable.


balarsen:
Pathological case for sure

jtniehof:
Pathalogical, but I wrote it because I wound up with a file that didn't have them match, and all my checks were passing but stuff still didn't work! :)

PR #236, opened 2019-10-27
pycdf.istp.VarBundle: py3k fix for complicated variable-conflict test

jtniehof:
I can reproduce this but it's hard to describe. When using a single source variable and slicing it two different ways for the output, VarBundle is supposed to detect when a variable needs to be renamed and when it conflicts. In this particular case (which might be connected to the fact that one of the dependencies depends on something else), on Python 3 it erroneously thinks that the Epochs conflict between the two output variables even when there's no slicing on the record dimension.

The path that gets there is complicated but the cause is simple: bytes and strings don't compare equal, so it has to be decoded.

EDIT: To be explicit, this PR fixes the problem. (As well as testing for it.)

### DIFF
--- a/spacepy/pycdf/istp.py
+++ b/spacepy/pycdf/istp.py
@@ -1651,10 +1651,13 @@ class VarBundle(object):
                 #Check for variable renaming from the input to output
                 newval = namemap.get(oldval, oldval)
                 if preexist:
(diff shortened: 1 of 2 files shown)


balarsen:
Nice catch with a not so horrid fix.

PR #239, opened 2019-11-05
toolbox: handle object arrays in pmm (closes #183)

### LINKED ISSUE #183: Toolbox.pmm does not work with date or object arrays
`toolbox.pmm` does not work with date or object arrays

Maybe just wrap 
   ` ind = np.isfinite(a)`
in a `try` `except`

```
import spacepy.toolbox as tb
import datetime
a = [datetime.datetime.now() + datetime.timedelta(minutes=x) for x in range(100)]
tb.pmm(a)
```

### Error message/Traceback:
```
<ipython-input-16-a01c8f1b7782> in <module>
----> 1 tb.pmm(a)

~/anaconda3/envs/python3/lib/python3.7/site-packages/spacepy/toolbox/__init__.py in pmm(a, *b)
   1504     [[0, 9], [3, 12]]
   1505     """
-> 1506     ind = np.isfinite(a)
   1507     try:
   1508         ans = [[ np.min(a[ind]), np.max(a[ind]) ]]

TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

```
Darwin-18.6.0-x86_64-i386-64bit
sys.version_info(major=3, minor=7, micro=3, releaselevel='final', serial=0)
numpy=1.16.4
scipy=1.3.0
matplotlib=3.1.0
```

drsteve:
Closes #183

This PR provides:
- a test for the failures raised in issue #183
- a trap/workaround for the failure mode (objects can't be determined as finite or not)
- a warning on hitting the trap so the user knows that non-finite values may be present

Any suggestions for wording on the warning? Or other implementation issues?

### DIFF
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -1624,7 +1624,7 @@ def pmm(a, *b):
         ans = [[ np.min(a[ind]), np.max(a[ind]) ]]
     except TypeError:
         a_tmp = np.asarray(a)
-        if np.issubdtype(a_tmp.dtype, 'S') or np.issubdtype(a_tmp.dtype, 'U'):
+        if a_tmp.dtype.type in [np.dtype('S').type, np.dtype('U').type]:
             a_tmp = np.require(a_tmp, dtype=object)
         ans = [[ np.min(a_tmp[ind]), np.max(a_tmp[ind]) ]]
     for val in b:
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -1624,6 +1624,8 @@ def pmm(a, *b):
         ans = [[ np.min(a[ind]), np.max(a[ind]) ]]
     except TypeError:
         a_tmp = np.asarray(a)
+        if np.issubdtype(a_tmp.dtype, 'S') or np.issubdtype(a_tmp.dtype, 'U'):
+            a_tmp = np.require(a_tmp, dtype=object)
         ans = [[ np.min(a_tmp[ind]), np.max(a_tmp[ind]) ]]
     for val in b:
         ind = np.isfinite(val)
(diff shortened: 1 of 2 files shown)


jtniehof:
Rather than `arange` I would probably use `ind = Ellipsis` to get the whole array. It makes it a little more obvious what you're doing and dodges any potential problem with multidimensional arrays (I didn't find any in testing, but you never know.)

For the warning, there's little idea of the implications. `Unable to exclude infinite values, which may result in wrong output`?

This only works for `a`. If the `b` args are object type, they still fail. Any reason we can't just rewrite the whole thing as `*args` and loop `for a in args` appending to `ans` each time?

Finally, this doesn't work for strings. That would be a different PR except it does work for objects. Patch attached.

[string.txt](https://github.com/spacepy/spacepy/files/3811051/string.txt)

drsteve:
Thanks, updates made.

Your string-type test gave me numpy `FutureWarning`s like:
```
 FutureWarning: Conversion of the second argument of issubdtype from `'U'` to `str` is deprecated.
In future, it will be treated as `np.str_ == np.dtype('U').type`.
  if np.issubdtype(a_tmp.dtype, 'S') or np.issubdtype(a_tmp.dtype, 'U'):
```

So I've also switched that test to
```
a_tmp.dtype.type in [np.dtype('S').type, np.dtype('U').type]
```
which should do the same thing, but avoids the warning.

jtniehof:
Good catch on the deprecation. Thoughts on the b arguments? I can probably just push to your branch....

drsteve:
Right, forgot about that...
Why not just get rid of the code for the `b`s and loop the current code for the `a` argument over all inputs, appending to the answer list?

If that sounds sensible then I can give it a crack later? If you have other ideas, or wanted to push something then I'm good with that too.

jtniehof:
That's exactly what I was thinking, just make the function signature *args, start with ans=[], and then for a in args BLAH append to ans, remove the whole "for val in b" loop.

jtniehof:
Looks good. Since @balarsen hasn't chimed in I'll assume he doesn't object and will merge once Travis is happy.

balarsen:
I’m happy.

Issue #233, opened 2019-10-22
OMNI data is only available through January 2017


ljlamarche:
Updating the OMNI database with [spacepy.toolbox.update()](https://pythonhosted.org/SpacePy/autosummary/spacepy.toolbox.update.html#spacepy.toolbox.update) only updates the database up to January 31, 2017.  This is probably because the source used for OMNI data ([ViRBO](http://virbo.org/)) is no longer maintained.  NASA's [Space Physics Data Facility](https://spdf.gsfc.nasa.gov/) looks like it is the appropriate place to get this data now.

The spacepy [omni documentation](https://pythonhosted.org/SpacePy/omni.html) states that
>"The omni module primarily manages the hourly OMNI2 and Qin-Denton data, which are sourced from the Virtual Radiation Belt Observatory (ViRBO), who maintain these data sources."

However, the [ViRBO](http://virbo.org/) page states 
>"The ViRBO project was discontinued in 2014 and these pages are no longer being updated."

The contents of `~/.spacepy/spacepy.rc` confirm that both the omni2 and the qindenton data is being sourced from the ViRBO site.
```
[spacepy]
  
#apply_plot_styles: True #default in SpacePy 0.1.6
#enable_deprecation_warning: True #default in SpacePy 0.1.6
#leapsec_url: ftp://maia.usno.navy.mil/ser7/tai-utc.dat #default in SpacePy 0.1.6
#ncpus: 8 #default in SpacePy 0.1.6
#omni2_url: http://virbo.org/ftp/OMNI/OMNI2/merged/latest/OMNI_OMNI2-latest.cdf.zip #default in SpacePy 0.1.6
#psddata_url: http://spacepy.lanl.gov/repository/psd_dat.sqlite #default in SpacePy 0.1.6
#qindenton_url: http://virbo.org/ftp/QinDenton/hour/merged/latest/WGhour-latest.d.zip #default in SpacePy 0.1.6
#support_notice: True #default in SpacePy 0.1.6
```
After running [spacepy.toolbox.update()](https://pythonhosted.org/SpacePy/autosummary/spacepy.toolbox.update.html#spacepy.toolbox.update), both `~/.spacepy/data/omni2data.h5` and `~/.spacepy/data/omnidata.h5` are modified, however, the last time stamp in `omnidata.h5` is `2017-01-31T23:00:00` and the last time stamp in `omni2data.h5` is `2012-12-31T23:30:00`. 

[...]

PR #240, opened 2019-11-14
pycdf.concatCDF: function to concatenate data from several CDFs

jtniehof:
This adds a function to pycdf to read data from several CDFs and present it as if it were from one (essentially like a VarCopy from multiple CDFs). It also rearranges the pycdf docs a little bit, since this is the first top-level function. This would be part of the #233 fix.

Definitely open to suggestions on if this should be named something else or put somewhere else.


jtniehof:
Ah, bitten by bytes/str again.

balarsen:
> Ah, bitten by bytes/str again.

Is this then ready to go or does it need another commit?

jtniehof:
Yes, it's ready to go...the force-push was fixing the issue.

PR #241, opened 2019-11-15
Get OMNI2 data from CDAWeb

jtniehof:
This PR is a step towards closing #233:

- Completely reworks our downloading in toolbox.update to avoid deprecations. Ultimately this should be replaced by a more robust general downloading infrastructure.
- Switches to the 6-month OMNI2 CDFs from CDAWeb/SPDF instead of the combined from ViRBO
- Fixes a little problem in the omni documentation in specifying OMNI2 database.

### DIFF
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -909,6 +909,14 @@ def get_url(url, outfile=None, reporthook=None):
     See Also
     ========
     progressbar
+
+    Notes
+    =====
+    This function honors proxy settings as described in
+    :func:`urllib.request.getproxies`. Cryptic error messages (such as
+    ``Network is unreachable``) may indicate that proxy settings
+    should be defined as appropriate for your environment (e.g. with
+    ``HTTP_PROXY`` or ``HTTPS_PROXY`` environment variables).
     """
     r = urllib.request.Request(url)
     if spacepy.config.get('user_agent', ''):
@@ -942,6 +950,9 @@ def update(all=True, QDomni=False, omni=False, omni2=False, leapsecs=False, PSDd
     """
     Download and update local database for omni, leapsecs etc
 
+    Web access is via :func:`get_url`; notes there may be helpful in
+    debugging errors.
+
     Parameters
     ==========
     all : boolean (optional)
@@ -959,6 +970,10 @@ def update(all=True, QDomni=False, omni=False, omni2=False, leapsecs=False, PSDd
     =======
     out : string
         data directory where things are saved
+
+    See Also
+    ========
+    get_url
 
     Examples
     ========
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -882,7 +882,7 @@ class LinkExtracter(html.parser.HTMLParser):
             self.links.append(value)
 
 
-def get_url(url, outfile=None, reporthook=None):
+def get_url(url, outfile=None, reporthook=None, cached=False):
     """Read data from a URL
 
     Open an HTTP URL, honoring the user agent as specified in the
@@ -900,6 +900,10 @@ def get_url(url, outfile=None, reporthook=None):
     reporthook : callable (optional)
         Function for reporting progress; takes arguments of block
         count, block size, and total size.
+    cached : bool (optional)
+        Compare modification time of the URL to the modification time
+        of ``outfile``; do not retrieve (and return None) unless
+        the URL is newer than the file.
 
     Returns
     =======
@@ -926,6 +930,20 @@ def get_url(url, outfile=None, reporthook=None):
         r.close()
         raise RuntimeError('HTTP status {} {}'.format(r.code, r.msg))
     headers = r.info()
+    if cached:
+        if outfile is None:
+            r.close()
+            raise RuntimeError('Must specify outfile if cached is True')
+        modified = headers.get('Last-Modified', None)
+        if os.path.exists(outfile) and modified is not None:
+            remote_mod = datetime.datetime.strptime(
+                modified, "%a, %d %b %Y %X GMT")
+            #Timestamp is truncated to second, so do same for local
+            local_mod = datetime.datetime.utcfromtimestamp(os.path.getmtime(
+                outfile)).replace(microsecond=0)
+            if remote_mod <= local_mod:
+                r.close()
+                return None
     size = int(headers.get('Content-Length', 0))
     blocksize = 1024
     count = 0
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -19,6 +19,11 @@ from __future__ import absolute_import
 
 import datetime
 import glob
+try:
+    import html.parser
+except ImportError: #python2
+    import HTMLParser as html
+    html.parser = html
 import itertools
 import os
 import os.path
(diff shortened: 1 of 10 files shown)


jtniehof:
I am a little concerned that the new database (from SPDF) is smaller, about 120MB vs. I think 180 for the version from ViRBO. This is after rendering to HDF5 and stored in the .spacepy directory.

drsteve:
Hmm. Does SPDF not store the data at the same precision? Are they missing any variables? Is the start date the same? I'll try to take a look at this PR in the next couple of days.

jtniehof:
CDAWeb has two variables that ViRBO does not. Most of the ViRBO variables are DOUBLES and CDAWeb is mostly REAL4, with a few REAL8. So it's probably mostly precision.

spacepy/toolbox/__init__.py
```diff
@@ -723,6 +733,211 @@ def dictree(in_dict, verbose=False, spaces=None, levels=True, attrs=False, **kwa
...
+            progressbar(i + 1, 1, len(filenames))
+        print("Reading OMNI2 files ...")
+        data = spacepy.pycdf.concatCDF([spacepy.pycdf.CDF(os.path.join(td, f))
+                                        for f in filenames])
+    finally:
+        shutil.rmtree(td)
+    #Map keyed by the variable names as in the OMNI2 data from CDAWeb,
+    #valued by the variable names as in the OMNI2 data from ViRBO
+    #i.e. this is from: to
+    #Non means not in VirBO, so delete.
```
drsteve: Typo in the comments on toolbox/__init__.py:791
Non -> None

drsteve:
@jtniehof Any thoughts on my previous comments? I'm happy to merge, as long as you've had a chance to look at my comments.

jtniehof:
I'll be working spacepy stuff tomorrow--I got crushed at work before the holiday and then needed the real time off. There will be changes, so no need to merge just yet. Thanks.

jtniehof:
Typo fixed.

I'm reluctant to trap an exception just to guess why it happened, particularly when it comes to an essentially misconfigured user environment. I tried to address this with some notes in the docstrings.

Original PR performed a full download and reprocess regardless of what was (or wasn't) updated, as that was the behavior of the old code. I added a check to only download updated files. This does require keeping around a copy of the original CDFs. It's also starting to get into the download manager territory; we should examine that at some point.

This pushes things into code that needs some testing. I really didn't feel comfortable starting up an http server in the unit tests, so I started a parallel integration test infrastructure...still unittest-based, but intended for heavier tests. The scripts start with "integration_" instead of "test_". Up to you if we want to run integration_all.py in the CI; I would suggest yes.

jtniehof:
@drsteve , does this address your comments and happy to merge? I know it's a pretty big delta from the original PR...

drsteve:
Looks good as a drag-and-drop replacement for the old OMNI2.

Question(s):
- Fill values? Should we be doing something with these? Or should that be handled on read by the omni module?

For example: if you do
```
import spacepy.time as spt
import spacepy.omni as om
tt = spt.tickrange('2019/11/12','2019/12/12', 3/24.)
omni = om.get_omni(tt, dbase='OMNI2hourly')
omni['Kp_index']
```
You get
```
dmarray([25. , 15. ,  7. ,  5. ,  6.5, 10. ,  6.5,  1.5,  0. ,  0. ,  0. ,
          0. ,  0. ,  0. ,  3.5, 10. , 16.5, 15. ,  6.5,  3. ,  5. ,  5. ,
          1.5,  0. ,  0. ,  0. ,  0. ,  1.5,  3. ,  6.5, 11.5, 10. , 53. ,
         99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. ,
         99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. ,
         99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. ,
         99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. ,
         99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. ,
         99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. ,
         99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. ,
         99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. ,
         99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. ,
         99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. ,
         99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. ,
         99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. , 99. ,
         99. , 99. , 99. ,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,
          nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,
          nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,
          nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,
          nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,
          nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,
          nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan])
```
The `omni` module is setting everything outside the valid range to `nan`, but the downloaded data has fill. An unsuspecting user could well automatically divide by 10 (because OMNI2 provides Kp*10) and then the fill is an invalid Kp of 9.9...

My guess is that a change to address that should be a different PR as it's changing `omni` stuff, and we probably shouldn't change values in the download/update. In which case, I think this is good to merge.

jtniehof:
Hmmm. Do you know offhand of a period in the old omni when there are fill values so we can compare? If this is an old problem I agree with opening a new issue and a new PR. But I don't want to introduce a regression if it didn't happen before.

The valid range as defined in the CDF is 0-90, with explicit FILLVAL of 99 (which isn't ISTP-compliant, but that's another issue.) I would consider it reasonable for update to replace all invalid values with nan, just before the toHDF5 call. Since `concatCDF` maintains the metadata, that's possible.

drsteve:
Not offhand, and I've rarely used the OMNI2 as opposed to the QD - but a quick search later and I think there were fill values in the proton flux.

```
tt = spt.tickrange('2004-12-02T12:00:00', '2004-12-03', 3/24)
omni = om.get_omni(tt, dbase='OMNI2hourly')
omni['Proton_flux_gt_10_MeV']
```
then
```
dmarray([0.305, 0.295,   nan,   nan, 0.295])
```

Doing the same with the SPDF-sourced data:
```
dmarray([3.05000007e-01, 2.95000002e-01, 9.99999922e+04, 9.99999922e+04,
         2.95000002e-01])
```

So this probably would be a regression?

jtniehof:
I should be able to whip up a quick update tomorrow to fix it, then. I'll
see if the ViRBO OMNI2 was doing it for us or if update is doing it in a
different code path.

jtniehof:
ViRBO CDFs had the nan in them already. And obnoxiously, even integers were cast to float, probably to allow for NaN. I'll reproduce that but at some point in the future we might want to stop focusing on reproducing the ViRBO setup.

jtniehof:
Rebased against master since it's moved a lot, and force-pushed.

Four new commits:
- Add integration test to check this specific case. The way this one works is pushing me away from the idea of including integration tests in CI, but it could potentially be updated to actually do the download before checking. I think that's another conversation.
- Add a generic pycdf function to replace fill values with NaN
- Add some more feedback on the download progress since it was annoying me. This isn't very well tested because I'm on a connection that for some reason the Python web stuff doesn't work on (even spoofing UserAgent).
- Change all OMNI2 variables to float and replace fill with NaN

jtniehof:
I tested the download feedback changes a bit more; definitely ready to merge now. @drsteve , we might want to open a new issue (or project?) with the longer-term changes we were discussing.

drsteve:
Sorry about the delay - just took a quick look and everything seems good, in terms of a replacement for the old OMNI2. Merging.

PR #245, opened 2019-11-22
Fix plot.utils.EventClicker with non- (or differing-)timezone datetimes, closes #243

### LINKED ISSUE #243: plot.utils.EventClicker with non-tz datetimes
`matplotlib.dates.num2date` returns dates with a tzinfo object (at least now it does.) This is causing failures "TypeError: can't compare offset-naive and offset-aware datetimes" in _relim in the bisects for a plot made from tz-naive datetimes.

```
ticks = spacepy.time.tickrange('2001-01-01', '2010-12-31', deltadays=datetime.timedelta(hours=1))
data = spacepy.omni.get_omni(ticks)
plt.plot(data['UTC'], data['Pdyn'])
clicker = spacepy.plot.utils.EventClicker(n_phases=1)
clicker.analyze()
```

asher-m:
`matplotlib.dates.num2date` assumes the system timezone if no timezone is provided.  This can create problems if trying to use `EventClicker` on tz-naive data or data with a timezone different than the system.

This problem can be eliminated by handing the timezone of the data to `num2date`.  If the timezone is `None`, the [incorrect] system timezone is removed by the following line.

I've updated all calls to `num2date` to include the `tzinfo` from `xdata[0]`, and then remove the `tzinfo` from the response if it was `None` to begin with.  (This convoluted order is necessary because `num2date` assumes the system timezone if the `tz` arg is `None`.)

### DIFF
--- a/spacepy/plot/utils.py
+++ b/spacepy/plot/utils.py
@@ -294,8 +294,11 @@ class EventClicker(object):
         if self.interval is None:
             (left, right) = self.ax.get_xaxis().get_view_interval()
             if self._x_is_datetime:
-                right = matplotlib.dates.num2date(right)
-                left = matplotlib.dates.num2date(left)
+                right = matplotlib.dates.num2date(right, tz=self._xdata[0].tzinfo)
+                left = matplotlib.dates.num2date(left, tz=self._xdata[0].tzinfo)
+                if not self._xdata[0].tzinfo:
+                    right = right.replace(tzinfo=None)
+                    left = left.replace(tzinfo=None)
             self.interval = right - left
 
         if not self._xdata is None:
@@ -370,7 +373,9 @@ class EventClicker(object):
             self._data_events[-1, self._curr_phase] = \
                                   [self._xdata[idx], self._ydata[idx]]
         if self._x_is_datetime:
-            xval = matplotlib.dates.num2date(xval)
+            xval = matplotlib.dates.num2date(xval, tz=self._xdata[0].tzinfo)
+            if not self._xdata[0].tzinfo:
+                xval = xval.replace(tzinfo=None)
         if self._events is None:
             self._events = numpy.array([[[xval, yval]] * self.n_phases])
         self._events[-1, self._curr_phase] = [xval, yval]
@@ -445,7 +450,9 @@ class EventClicker(object):
         if event.key == ' ':
             rightside = self.ax.xaxis.get_view_interval()[1]
             if self._x_is_datetime:
-                rightside = matplotlib.dates.num2date(rightside)
+                rightside = matplotlib.dates.num2date(rightside, tz=self._xdata[0].tzinfo)
+                if not self._xdata[0].tzinfo:
+                    rightside = rightside.replace(tzinfo=None)
             self._relim(rightside)
         if event.key == 'delete':
             self._delete_event_phase()
@@ -455,9 +462,6 @@ class EventClicker(object):
         if self._x_is_datetime:
             xmin = left_x - self.interval/10
             xmax = left_x + self.interval + self.interval/10
-            if not self._xdata[0].tzinfo:
-                xmin = xmin.replace(tzinfo=None)
-                xmax = xmax.replace(tzinfo=None)
         else:
             xmin = left_x - 0.1 * self.interval
             xmax = left_x + 1.1 * self.interval
--- a/spacepy/plot/utils.py
+++ b/spacepy/plot/utils.py
@@ -455,6 +455,9 @@ class EventClicker(object):
         if self._x_is_datetime:
             xmin = left_x - self.interval/10
             xmax = left_x + self.interval + self.interval/10
+            if not self._xdata[0].tzinfo:
+                xmin = xmin.replace(tzinfo=None)
+                xmax = xmax.replace(tzinfo=None)
         else:
             xmin = left_x - 0.1 * self.interval
             xmax = left_x + 1.1 * self.interval


spacepy/plot/utils.py
```diff
@@ -294,8 +294,11 @@ def analyze(self):
         if self.interval is None:
             (left, right) = self.ax.get_xaxis().get_view_interval()
             if self._x_is_datetime:
-                right = matplotlib.dates.num2date(right)
-                left = matplotlib.dates.num2date(left)
+                right = matplotlib.dates.num2date(right, tz=self._xdata[0].tzinfo)
+                left = matplotlib.dates.num2date(left, tz=self._xdata[0].tzinfo)
+                if not self._xdata[0].tzinfo:
```
jtniehof:
```suggestion
```

spacepy/plot/utils.py
```diff
@@ -294,8 +294,11 @@ def analyze(self):
...
+                    right = right.replace(tzinfo=None)
+                    left = left.replace(tzinfo=None)
```
jtniehof:
```suggestion
```

spacepy/plot/utils.py
```diff
@@ -294,8 +294,11 @@ def analyze(self):
         if self.interval is None:
             (left, right) = self.ax.get_xaxis().get_view_interval()
             if self._x_is_datetime:
-                right = matplotlib.dates.num2date(right)
-                left = matplotlib.dates.num2date(left)
+                right = matplotlib.dates.num2date(right, tz=self._xdata[0].tzinfo)
+                left = matplotlib.dates.num2date(left, tz=self._xdata[0].tzinfo)
+                if not self._xdata[0].tzinfo:
+                    right = right.replace(tzinfo=None)
```
jtniehof:
```suggestion
```

spacepy/plot/utils.py
```diff
@@ -370,7 +373,9 @@ def _add_event_phase(self, xval, yval):
             self._data_events[-1, self._curr_phase] = \
                                   [self._xdata[idx], self._ydata[idx]]
         if self._x_is_datetime:
-            xval = matplotlib.dates.num2date(xval)
+            xval = matplotlib.dates.num2date(xval, tz=self._xdata[0].tzinfo)
+            if not self._xdata[0].tzinfo:
```
jtniehof:
```suggestion
```

spacepy/plot/utils.py
```diff
@@ -370,7 +373,9 @@ def _add_event_phase(self, xval, yval):
...
+                xval = xval.replace(tzinfo=None)
```
jtniehof:
```suggestion
```

spacepy/plot/utils.py
```diff
@@ -445,7 +450,9 @@ def _onkeypress(self, event):
         if event.key == ' ':
             rightside = self.ax.xaxis.get_view_interval()[1]
             if self._x_is_datetime:
-                rightside = matplotlib.dates.num2date(rightside)
+                rightside = matplotlib.dates.num2date(rightside, tz=self._xdata[0].tzinfo)
+                if not self._xdata[0].tzinfo:
```
jtniehof:
```suggestion
```

spacepy/plot/utils.py
```diff
@@ -445,7 +450,9 @@ def _onkeypress(self, event):
...
+                    rightside = rightside.replace(tzinfo=None)
```
jtniehof:
```suggestion
```

spacepy/plot/utils.py
jtniehof:
```suggestion
```

jtniehof: This one got marked as a multi-line edit so can't be applied, added as another comment and resolving this.

jtniehof:
So in theory you should be able to pop over to the Files tab, "add suggestion to batch" on all, and commit the batch, assuming you agree with my rationale.

jtniehof:
Oh, nevermind, that doesn't work because "None" is interpreted as "use system" not "None." That's really the one nasty bit of something that's otherwise really elegant. Let me poke a little more.

asher-m:
> Is there any need to do the if-not replace with None? It'll mostly be replacing None with None, correct?

No, in fact it will be replacing the system timezone with None.  This is because `num2date` assumes `tz=None` means "replace with system timezone" (so it can never hand back a tz-naive datetime,) so if the xdata has `tzinfo=None` `num2date` will assume the system timezone. See the source ([here on their docs](https://matplotlib.org/3.1.1/_modules/matplotlib/dates.html#num2date) or [here on Github](https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/dates.py#L473)) for more details.  `num2date` is written:
```
def num2date(x, tz=None):
    """ [docstring] """
    if tz is None:
        tz = _get_rc_timezone()
    [more code.  Note that their master on Github is different than the docs for 3.1.1, but this functionality remains identical.]
```
This is the same in 1.5.1 as well.


`_get_rc_timezone` then gets the system timezone, also [dating back to 1.5.1](https://github.com/matplotlib/matplotlib/blob/v1.5.1/lib/matplotlib/dates.py#L171), (possibly earlier,) though this functionality remains the same through [the most recent](https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/dates.py#L173):
```
def _get_rc_timezone():
    """Retrieve the preferred timezone from the rcParams dictionary."""
    s = matplotlib.rcParams['timezone']
    if s == 'UTC':
        return UTC
    return dateutil.tz.gettz(s)
```

I think the key is that `num2date` cannot hand back a tz-naive datetime, so even if we pass `tz=None` the system timezone will be applied to what's handed back.

asher-m:
> No, in fact it will be replacing the system timezone with None.

> Oh, nevermind, that doesn't work because "None" is interpreted as "use system" not "None." That's really the one nasty bit of something that's otherwise really elegant. Let me poke a little more.

Oh wow...writing my response took a lot longer than I expected... 😅

jtniehof:
So, this is interesting. As originally written we explicitly forced tzinfo to None, but then there was a brief period where we were using a spacepy.time.num2date and it was lost in the transition back to matplotlib.dates.

It looks like we've never tried doing all the scaling in matplotlib number space and only converting to datetime on the return, or at least that's never been in a commit. I did some poking and going for set_xlim with numbers instead of datetimes doesn't change the axis away from being a datetime. It really does seem to be handling them as numbers everywhere except in the handling of interval. I wonder if self.interval shouldn't be converted to a number and then a lot of the datetime-specific stuff ripped out.

jtniehof:
Looks like staying in number space would also require updating get_events and get_events_data

jtniehof:
Having completely munged the diff from the comments, I looked back over and honestly, the "extra checks" bits are small, and going into pure-number space would probably not actually simplify things much. So let's run with this...thanks.

PR #255, opened 2019-12-09
plot.utils.EventClicker pcolormesh (and non-line object) support

asher-m:
These changes add support to EventClicker for matplotlib.pyplot.pcolormesh (QuadMesh) objects, and should be extensible enough to include any pyplot object that doesn't have line data associated with it.

I've included a new example (that may need reformatting) in the doctstring for EventClicker to show how this works.  EventClicker.get_events_data does not work for a pcolormesh/non-line object, as there isn't any "event data" to grab, but get_events will still get the x-value or datetime (whichever is used for the x-axis) of wherever was clicked.

(Incidentally Github wants to include some of my previous commits from a different PR in this.  I don't know why, nor how to exclude them, but if this needs changing let me know.)

### DIFF
--- a/spacepy/plot/utils.py
+++ b/spacepy/plot/utils.py
@@ -143,6 +143,24 @@ class EventClicker(object):
     >>> min(troughvals) <= -1.0 #should bottom-out at -1
     True
 
+    >>> import spacepy.plot
+    >>> import datetime
+    >>> import matplotlib.pyplot as plt
+    >>> import numpy
+    >>> import pytz
+    >>> xmax = 1000
+    >>> tz = pytz.timezone('America/New_York') #get a timezone
+    >>> x, y = numpy.linspace(0, xmax, xmax + 1), numpy.linspace(0, 100, 101)
+    >>> x_like_date = numpy.array([tz.localize(
+    ... datetime.datetime(2019, 12, 1, 0, 0, 0) + datetime.timedelta(hours=i))
+    ... for i in range(0, xmax + 1)])
+    >>> xx, yy = numpy.meshgrid(x, y)
+    >>> z = 1 + numpy.exp(-(yy - 20)**2 / (25)**2)
+    ... * numpy.sin(2 * numpy.pi * xx * 3 / xmax)**2 #something like a spetrogram
+    >>> plt.pcolormesh(x_like_date, y, z)
+    >>> clicker = spacepy.plot.utils.EventClicker(n_phases=1)
+    >>> clicker.analyze()
+
     .. autosummary::
          ~EventClicker.analyze
          ~EventClicker.get_events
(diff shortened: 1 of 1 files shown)


jtniehof:
Since your master still has the commits from your previous PR, and the commit IDs are not maintained in a rebase, those commits are considered to be changes against spacepy/master. spacepy/master has two commits that you don't, and you have three commits that spacepy/master doesn't. The fact that the first two have the same deltas doesn't matter.

I think this should fix it:
```
git fetch origin/master
git rebase --interactive origin/master
```
(remove the two tzinfo commits)

Then you can push --force.

This is why doing a PR against your master can be a pain, although really it comes down to not updating against upstream master.

jtniehof:
Ah, wait, you didn't do the PR against you master. *headdesk*. Do the same procedure, though, and also be sure to force-pull your master.

asher-m:
Okay, after a bit of cleverness I've managed to rebase the one important commit here ontop of Spacepy while leaving the 2 old commits from the other PR behind (now in their own branch in my fork.)  As far as rebasing goes then, this should be good.

Let me know if there should be any changes then.  All of the changes on my end are now properly in their own branch(es).

jtniehof:
There should probably also be a CHANGELOG note; it's a fairly substantial improvement.

asher-m:
I haven't updated the CHANGELOG yet, but I want to ask about the example.  As I mention in #258, changing the shape of a dmarray doesn't quite work right now.  That leaves the following for an example:

```
    >>> import spacepy.plot.utils
    >>> import spacepy.time
    >>> import datetime
    >>> import matplotlib.pyplot as plt
    >>> import numpy
    >>> t = spacepy.time.tickrange('2019-01-01', #get a range of days
    ...                            '2019-12-31',
    ...                            deltadays=datetime.timedelta(days=1))
    >>> y = numpy.linspace(0, 100, 1001)
    >>> seconds = t.TAI - t.TAI[0]
    >>> seconds = seconds.base #base of dmarray so reshape works
    >>> tt, yy = numpy.meshgrid(seconds, y) #use TAI to get seconds
    >>> z = 1 + (numpy.exp(-(yy - 20)**2 / 625) #something like a spectrogram
    ...          * numpy.sin(3e-7 * numpy.pi * tt)**2)
    >>> plt.pcolormesh(t.UTC, y, z)
    >>> clicker = spacepy.plot.utils.EventClicker(n_phases=1)
    >>> clicker.analyze()
    >>> clicker.get_events() #returns an array of the things clicked
    >>> clicker.get_events_data() is None #should be nothing
    True
```

I don't know how to allow the array to be reshaped in meshgrid apart from doing this.  Does this look okay?

jtniehof:
I suggest this:
`seconds = numpy.asarray(seconds) #normal ndarray so reshape (in meshgrid) works`
`asarray` instead of `base` because `asarray` will work (without a copy) even given an ndarray, whereas `base` for an ndarray is `None`. And then making it explicit where the reshape comes from.

asher-m:
That looks better, and it's at least explicit what's going on.  ...changes made and pushed.

jtniehof:
Looks good! Thanks!

Issue #242, opened 2019-11-18
plot utils docs not being built


jtniehof:
The `plot.utils` docs aren't being generated, probably something weird with the autosummary for submodules.

jtniehof:
I think adding a Submodules section, as at the bottom of pycdf.rst, to plot.rst, should do the trick.

asher-m:
Yep, that seems to have done the trick.  I'll make a PR in a moment.

jtniehof:
Probably better as a separate PR that just includes this (relative to master, please.)

asher-m:
Just created the pull request.  As I mention there this _does_ fix the problem that plot.utils wasn't being linked to the docs for plot, but I've realized now that spectrogram (the submodule) and spectrogram.spectrogram (the class in the spectrogram submodule) aren't being linked either.  It'd make sense to fix both of these now.

PR #259, opened 2019-12-12
plot.utils documentation: Link plot.utils to plot docs so it can be viewed (and other plot module doc fixes)

asher-m:
This applies the changes discussed in #242.  However, after taking a closer look, this doesn't actually fix all the problems with the plot module documentation.  Specifically, there seems to be a problem generating the docs for plot.spectrogram (the submodule) and plot.spectrogram.spectrogram (the class).

Additionally, there seem to be some inconsistencies within all the submodules, (some things that are functions are called classes or modules, and vice versa (in all ways)).

I tried quite a few things but couldn't quite fix it.  I don't know how easy it is for you to checkout, (or maybe just make the same changes in your local checkout) but I think it'd make sense to fix all the documentation for the plot module now.

So, any ideas on how to get the correct linking for plot.spectrogram and its members and any other changes?

### DIFF
--- a/Doc/source/plot.rst
+++ b/Doc/source/plot.rst
@@ -7,3 +7,12 @@ plot - Plot, various specialized plotting functions and associated utilities
 .. currentmodule:: spacepy.plot
 .. automodule:: spacepy.plot
 
+Submodules
+----------
+
+.. autosummary::
+    :toctree: autosummary  
+    :template: clean_module.rst
+
+    utils
+    spectrogram


jtniehof:
Yeah, there are lots of other issues in the plot docs, particularly submodules. But this at least helps make them apparent so they can be knocked down over time. I don't think it's a requirement to fix them all before merging this. If particular ones stand out, go ahead and open an issue so we can track them.

jtniehof:
Taking a look at the output, we already had a submodule autosummary list, it just wasn't put in a toctree. So I modified that instead in #260; sorry for the misdirection.

asher-m:
No worries.  I'll close this.

Issue #26, opened 2018-09-21
Remove Python 2 support


jtniehof:
(Before anybody panics: the title of this issue is to indicate we'll close it when Python 2 support is removed, not that this is an immediate goal!)

Python 2.7 will be unsupported sometime in 2020. All our dependencies have pledged to EOL Python 2 support no later than that ( https://python3statement.org/ ). I suggest a gradual phase-out:

1. No prebuilt Python 2.7 packages: We should deliver one for 0.1.7. But potentially not as soon as the next release (presumably 0.1.8). June 2019, perhaps?
2. Allow new features/functions that don't support Python 2.7 as long as they don't break existing functionality. Beginning of 2020?
3. No workarounds for versions of dependencies that don't support Python 3; I'd lump the date in with 2.
4. No fixes for bugs that cannot be reproduced on Python 3. I would take this as the latter of June 2020 and when numpy, scipy, matplotlib have all released versions that do not support Python 2, with a caveat of no later than fall AGU of 2020.
5. Actively remove existing Python 2 code from the codebase: January 1, 2021.

This would mean a committment to full support of Python 2.7 through June 2019 (or whatever date is chosen for 1), followed by a phaseout of 12-18 months, while still falling within the 2020 pledge.

jtniehof:
Actually I see CPython 2.7 support officially dies 1 Jan 2020, so I think this should happen sooner.

balarsen:
Here is numpy's plan:
https://github.com/numpy/numpy/blob/master/doc/neps/nep-0014-dropping-python2.7-proposal.rst#plan-for-dropping-python-27-support


Our current plan is as follows.

Until December 31, 2018, all NumPy releases will fully support both Python2 and Python3.

Starting on January 1, 2019, any new feature releases will support only Python3.

jtniehof:
Thanks Brian. I think the idea of trying to make sure "pip install spacepy" on Python 2 will always get the latest python 2 supporting one makes sense.

Reasonably compatible with that, I think my #1, #2, #3 could be lumped together as "after the first release after June 2019" (note #3 should be "no workarounds for dependencies that don't support Python 2"); leave #4 as June 2020 and #5 as January 2021. That still puts us as "no" to Python 2 in 2020, while being about 6mo behind numpy.

I do want to draw a distinction between "no further development on Python 2" and "ripping out Python 2" because I think that's a reasonable transition, as much as I'm itching to take a chainsaw to Python 2 code.

[...]

PR #265, opened 2020-01-03
Document py2k EOL plan (Closes #211)

### LINKED ISSUE #211: Put python 3 transition roadmap in docs
The Python 3 roadmap per #26 should be written into the documentation.

jtniehof:
This PR documents the plan for ending Python 2 support in SpacePy as in #26. The document is linked from the dependencies page (which now recommends Python 3 instead of 2). The install page is updated to more strongly recommend Python 3.

I think we all need to be on board with this so I'm explicitly requesting review from everybody...that can be a review of the plan and/or its presentation in this PR.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
-Changes in Version 0.2.2 (2019-xx-xx)
+Changes in Version 0.2.2 (2020-xx-xx)
 =====================================
+- Document end of support for Python 2
 plot
  - utils.EventClicker support for non-line plots (e.g. pcolor)
 pycdf
--- a/Doc/source/dependencies.rst
+++ b/Doc/source/dependencies.rst
@@ -14,15 +14,13 @@ Hard Dependencies
 =================
 Without these packages installed, SpacePy will not function.
 
-Python 2.6+
+Python 2.7+
 -----------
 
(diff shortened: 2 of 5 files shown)


drsteve:
You've staged this out more generously than I probably would've, but I'm happy to sign off on this. The only real downside is that it's committing us to several releases, and I'm not sure how much of a downside that really is...

jtniehof:
When I wrote it all out it did look like "wow, that's a lot of releases" but, as you say, may also be a benefit. And it gets us into a little more of a roadmap approach instead of release-what's-there, which I think can be a benefit.

drsteve:
So we just need a review/approval from @balarsen now?

jtniehof:
Yep!

balarsen:
Thumbs up. Super interesting how aggressive python is with moving beyond 2. The only thing I still have in python2 is ECT processing. That will end October so not a bad change over.

jtniehof:
I'm going to take that as an approved review :) I request somebody else do the merge just so there's more than my fingerprints on the commit.

This seems aggressive but the goal was for 2 to go away years ago; I think they just finally decided to stop asking and start telling. There's definitely still going to be some scrambling on my side.

PR #267, opened 2020-01-07
pybats/pwom.py: Added option to read pwom binary files for line class. 

aglocer:
This change modifies the _read part of the Line class. The read will now automatically detect 
if the file is ascii or binary and reads appropriately.  This feature is particularly useful for large runs with lots of lines as binary line output is much smaller. The change was tested with ascii and binary pwom lines generated on both linux and mac and with different compilers. The speed of the binary read appears to be substantially faster than the ascii read which may help when analyzing large numbers of lines. 

## PR Checklist
Please try to follow as much of this checkl;ist as you can for

### DIFF
--- a/spacepy/pybats/pwom.py
+++ b/spacepy/pybats/pwom.py
@@ -94,6 +94,213 @@ class Line(PbData):
 
     def _read(self, starttime):
         '''
+        Read line file; should only be called upon instantiation.
+        '''
+
+        #detect if file is binary or ascii
+        textchars = bytearray({7,8,9,10,12,13,27} | set(range(0x20, 0x100)) - {0x7f})
+        is_binary_string = lambda bytes: bool(bytes.translate(None, textchars))
+        IsBinary = is_binary_string(open(self.attrs['file'], 'rb').read(1024))
+                
+        if not IsBinary:
+            # Reads file assuming it is ascii
+            # Slurp whole file.
+            f=open(self.attrs['file'], 'r')
+            lines=f.readlines()
+            f.close()
+            
+            # Determine size of file.
+            nTimes=lines.count(lines[0])
+            nAlts =int(lines[2].strip())
+            self.attrs['nAlt']=nAlts; self.attrs['nTime']=nTimes
+
+            # Start building time array.
+            self['time']=np.zeros(nTimes, dtype=object)
+            
+            # Get variable names; pop radius (altitude).
+            var=(lines[4].split())[1:-1]
+            self._rawvar=var
+            
+            # Get altitude, which is constant at all times.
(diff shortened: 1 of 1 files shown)


jtniehof:
Looks like a big chunk of additional functionality; thanks. @spacecataz is best-placed to comment on the actual functionality (requesting review from him), but some implementation comments:

Since it looks like there's no commonality between the binary and text readers, does it make sense to rename the old function to `_read_ascii`, put most of the new stuff in `_read_binary`, and then have `_read` just do the binary check and then call either `_read_ascii` or `_read_binary` as appropriate?

Can the binary check be a little more described? It looks like it just examines if there are non-ASCII characters in the first 1k; is that an arbitrary number? Is there a more robust way to determine if it meets an expected format? There should probably be a unit test for this, as well (it's okay to check a small test file into `tests/data`).

Per [PEP-8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations), `is_binary_string` should be an internal function definition rather than an assigned lambda.

aglocer:
Thanks for the comments. I did also find one minor issue of an attribute not being set that needs to be set. I will wait for the additional review and make all changes at once.

spacecataz:
Looking at this, there may be a more elegant way to do this but I could use some example data to play with.  Pinging @aglocer offline for this...

jtniehof:
@spacecataz, any progress/status on this?

spacecataz:
> @spacecataz, any progress/status on this?

Let me take a look at this again.  I was looking at some long term solutions for these types of files, but I think we'll proceed with what's here and and adapt in the future.

jtniehof:
@spacecataz , anything further, or should we do some light cleanup and merge this?

jtniehof:
I am going to close and reopen this to retrigger the CI.

jtniehof:
I think the documentation fail is actually present in main now and fixed in #790.

spacepy/pybats/pwom.py
```diff
@@ -142,6 +149,146 @@ def _read(self, starttime):
...
+            i += 1
+        NextDataStart = i + 8
+
+        #get the total length of a snapshot
+        #note the 164 comes from the byte pads between variables
+        SnapshotLength = NextDataStart + nAlts*dbl_size*(nVars+1) + 164
+        #get number of times in the file
+        nTimes = int(len(fileContent) / SnapshotLength)
+        #correct length of snapshot due to padding issues
+        #SnapshotLength=int(len(fileContent)/nTimes)
```
jtniehof, line 209: I removed a few other commented-out debugging print statements, but wasn't sure about this. If it's necessary we should uncomment it; otherwise probably should go. @aglocer ?

jtniehof:
Rebased and updated since this will be merged in 0.8.0 not 0.7.0

PR #268, opened 2020-01-07
test_toolbox: additional tests for quaternion math

drsteve:
Expands test coverage for quaternion math.

This was motivated by issue #184. That issue is invalid, and should be closed, but to ensure/demonstrate correctness I've added more unit tests.

These tests are:
- quaternion conjugation (for both scalar positions)
- quaternion multiplication (for a simple case)
- quaternion normalization, testing for
  - known case, both scalar positions
  - near-zero case, both scalar positions

The reason #184 is invalid is that, in essence, the norm of the quaternion (w, i, j, k) is simply (w/n, i/n, j/n, k/n) where n is sqrt(w^2 + i^2 + j^2 + k^2). So the position of the scalar is generally irrelevant in the normalization and the reported behaviour is expected.

The reason that the `scalarPos` kwarg is still present is for handling the case where the quaternion magnitude is very small (which we take as 1e-12). In this case we define the quaternion as the identity quaternion (scalar: 1, vector (0,0,0)) such that no rotation is performed. To ensure a valid quaternion we need to know which position to write the scalar to.

### DIFF
--- a/Doc/source/toolbox.rst
+++ b/Doc/source/toolbox.rst
@@ -11,6 +11,7 @@ toolbox - Toolbox of various functions and generic utilities
 - `Array binning`_
 - `Array creation`_
 - `Array searching and masking`_
+- `Quaternion math`_
 - `Other functions`_
 - `Multithreading and multiprocessing`_
 - `System tools`_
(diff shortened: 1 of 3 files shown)


drsteve:
Realized that the quaternion math never got added to the `spacepy.toolbox.__all__` or to the docs. So I just pushed another commit to do that.

jtniehof:
Yikes, good catch on not being exposed in the docs.

PR #269, opened 2020-01-09
Document commitments for dependency versions

jtniehof:
This PR proposes a policy for removing support for old versions of dependencies. It tries to capture the discussion in #8 and also be reasonable compliant with [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html).

The idea is that if we have to bump a dependency version, rather than having a lot of back-and-forth we have a pretty clear line of if it's compatible with this policy and can then move on.

In addition it also makes some minor updates to `dependencies.rst` where it's clearly pathalogically wrong. I probably will open a different PR to finish up #8 but this is also a step towards that.

Until #211 is merged there's a cross-link that will fail, but otherwise this all builds fine. It'll probably also conflict with #211 but I'll fix this up when #211 is merged.

As it's a policy proposal, again I'm requesting review from everybody. Comments on making it shorter are particularly welcome.

### DIFF
--- a/Doc/source/dep_versions.rst
+++ b/Doc/source/dep_versions.rst
@@ -3,10 +3,13 @@ Dependency version support
 **************************
 
 SpacePy will occasionally drop support for old versions of
-:doc:`dependencies <dependencies>`. Dependency support is based on
-these principles:
+:doc:`dependencies <dependencies>`. Failures with older versions will
+not be treated as SpacePy bugs. Dependency support is based on these
+principles:
 
- #. The latest released version of a dependency is always supported.
+ #. SpacePy supports released versions of a dependency that meet a
+    minimum version requirement; there is no maximum supported
+    version.
  #. Support for old versions of dependencies will be dropped only for
     reason, e.g. if a new version is required to support a new feature
     or fix a bug. Maintenance of convoluted workarounds is included in
@@ -31,7 +34,7 @@ these principles:
        subminor releases (x.y.z for the same x.y). Where x.y.0 is supported,
        so is x.y.z for all z.
     #. All versions of all dependencies and all combinations thereof will
-       *not* necessarily be supported in continuous integration.
+       *not* necessarily be tested in continuous integration.
 
  #. No support will be provided for conflicting versions of
     dependencies. E.g. SciPy 1.4 requires NumPy 1.13. Although SpacePy
--- a/Doc/source/dep_versions.rst
+++ b/Doc/source/dep_versions.rst
@@ -1,0 +1,133 @@
+**************************
+Dependency version support
+**************************
+
+SpacePy will occasionally drop support for old versions of
+:doc:`dependencies <dependencies>`. Dependency support is based on
+these principles:
+
+ #. The latest released version of a dependency is always supported.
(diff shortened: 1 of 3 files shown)


jtniehof:
It's not clear, but the intention is we would maybe update the table every new Ubuntu LTS release, or if we really are bumping up against the limits and need to drop support for something. I don't think it needs to be updated every new month and every new release of every dependency.

balarsen:
This is a really good idea (comment in the code review). This was on my plate at one time and proved hard to do completely. Nice job and the policy save my comment I think is good.

jtniehof:
> The latest released version of a dependency is always supported

Yes, it's probably too strong. I considered including a statement that says "Supported doesn't always mean it works, just that if it doesn't work it's considered a bug." That really applies to the entire thing. Do we want wiggle language like that? It could also be rephrased as "The latest version at time of SpacePy release is always supported," i.e. we don't necessarily support future versions. (Then the question arises if we will delay a release solely for that.)

What I want to promise is that we won't say "We only support up to numpy 1.16 and that's not going to change."

jtniehof:
Rebased against current master just to make sure no conflicts, and force-pushed.

(I am envisioning Palpatine force-pushing somebody across the room...)

balarsen:
I fully agree with your last statement. Only supporting to some version is a huge annoyance as others move past eventually leading to a conflict. 

I think we would not want to hold releases up for a dependency (depending of course). 

I don’t know that we have to have weasel words but if one has a policy it should be stuck to as best as possible. Leading to wanting some words.

balarsen:
Looking at the deps work the Numpy 1.15.0 thing is my point. That came out and was the latest and we could not support.

jtniehof:
I was trying to cover that case with "Support for a particular version of a dependency does not imply a commitment to work around bugs in that version." 1.15.0 was definitely broken.

At very least I should drop the "always". How about replacing the first point with:

> SpacePy supports released versions of a dependency that meet a minimum version requirement; there is no maximum supported version.

and the preamble:

> SpacePy will occasionally drop support for old versions of :doc:`dependencies <dependencies>`. Failures with older versions will not be treated as SpacePy bugs. Dependency support is based on these principles:

balarsen:
I am happy with that.

jtniehof:
I've incorporated the changes from my comment plus a wording change on continuous integration: from all possible combinations not *supported* in CI to not *tested* in CI. The point is that we don't have a CI setup for itertools.product([numpy-1.4, numpy-1.5, numpy-1.6...numpy-1.14], [scipy-0.10, scipy-0.11, scipy-0.12...], [matplotlib-1.5, matplotlib-2.0, matplotlib-2.1...]). I only mention it because NEP-29 suggests testing all supported minor versions of Python and the combinations can quickly get out of hand. Exactly how we want to build the CI matrix is another discussion and I think ad-hoc is working okay for now.

jtniehof:
Then I think @spacecataz can merge if/when he approves.

jtniehof:
Pinging @spacecataz ...

jtniehof:
Thanks @spacecataz . Anybody want to merge so it's not just my fingerprints? :)

PR #271, opened 2020-01-20
Fixes/modernization of pybats.ram: plotting, python 3, etc.

drsteve:
Updates to the ram module in pybats. Addressing various related issues that arose when designing quicklook plotting for RAM-SCB for the CCMC.

Main changes:
- RamSat now inherits from SpaceData to get its dict-like behaviour (rather than from object and adding dict-like methods)
- Minor updates to docstrings to describe new kwargs
- Overhaul of omnidirectional flux calculation to be more flexible with species names as runs may not have particular species and new species may be added (Also, RAM-SCB has changed how species are named - this change allows for new and old files, plus future additions)
- Masking of bad values in omnidirectional flux calculation updated for robustness.
- Integer divide for finding indices for python 3 compatibility in PressureFile
- RamSat test updated to test for datetimes, not float seconds since run start -- It appears that the code was always intended to have datetimes here but this wasn't working correctly and so the test for seconds was passing. Updating the RamSat code fixed the behaviour to expected, so the test needed to be corrected.
- Broadened test in set_target to allow subclasses
- CHANGELOG updated to reflect changes in present branch

Tests pass, but I'd appreciate a closer look at:
1. Whether my update to set_target has side effects (@jtniehof)
2. Whether I've changed intended design in parts of RamSat through the changed inheritance (@spacecataz)

Any other comments, of course, welcome.

There's plenty more in ram.py that needs attention, but for that I should probably file an issue tracker so we can discuss what needs to happen.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -23,6 +23,8 @@ pybats
  - Fixed open/closed tolerance kwarg not working in add_b_magsphere.
  - Added new stream trace visualization method for Bats2d objects: add_stream_scatter
  - Bats2d.add_b_magsphere and add_stream_scatter now have kwargs for adding arrows on ray traces
+ - Updates to ram.RamSat including omnidirectional flux calculation
+ - ram.PressureFile fixed for Python 3.x failures
 toolbox
  - New function get_url to retrieve data from a URL
  - Change update to retrieve OMNI2 data from SPDF not ViRBO
--- a/spacepy/plot/utils.py
+++ b/spacepy/plot/utils.py
@@ -762,7 +762,7 @@ def set_target(target, figsize=None, loc=None, polar=False):
 
     '''
     # Is target an axes?  Make no new items.
-    if issubclass(type(target), plt.Axes):
+    if isinstance(target, plt.Axes):
         ax  = target
         fig = ax.figure
     else: # Make a new axis
--- a/spacepy/pybats/ram.py
+++ b/spacepy/pybats/ram.py
@@ -584,8 +584,8 @@ class RamSat(SpaceData):
 
         Usage: just call this method to integrate all species.
 
-        Arguments
-        ---------
+        Other Parameters
+        ================
         check : Boolean
                 If check is True (default) then the presence of existing
                 omni variables will cause this calculation to abort cleanly.
--- a/spacepy/pybats/ram.py
+++ b/spacepy/pybats/ram.py
@@ -1233,7 +1233,7 @@ class PressureFile(PbData):
         # Grid spacing/size:
         self.attrs['dTheta'] = self['theta'][1]-self['theta'][0]
         self.attrs['nTheta'] = int(np.round(2.*np.pi/self.attrs['dTheta'])+1)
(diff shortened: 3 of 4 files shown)


jtniehof:
Also, with using both SpaceData and dmarray, any reason to not just import spacepy.datamodel instead of multiple froms? I know it's got a lot of those and we don't have a standard right now, but PEP20 "Namespaces are one honking great idea".

drsteve:
In order, mostly:
- Good catch on the change from read-on-demand. Yes, that change was a side-effect of moving to the SpaceData subclass from the hand-rolled.
- Large files are unlikely to be an issue, but input from @spacecataz would be helpful here. For example, a 10-day simulation (which is long) at 1-minute output resolution (which is higher than typical) only comes in at around 0.8GB. If the files get to be big enough to worry (for me that'd be several times larger than this) then NetCDF has good tools for splitting (and combining), and filtering NetCDF. If needed, smaller files can be made easily. But I should be able to add back the read-on-demand by overloading `__getitem__` and `keys` again.
- Unless we change back to read-on-demand, I think it is a good idea to close the NetCDF - I can't think of any reason to keep it open. While it was exposed as a public method previous, it certainly wasn't advertised and I doubt the original intent was to have folks use it. I certainly wouldn't have a problem closing the file and dropping access through the `f` attribute.
- In `get_species_label` I used `in` because I forgot that `startswith` was a thing. I'm happy to change that as well as switching from a fixed offset to the length of `uvar` (plus the offset if an underscore is present).
- Let me think about the omni calculation. I'll write a test and then take a look. And FWIW I agree whole-heartedly with explicitly specifying the axis kwarg, I really dislike passing kwargs as positionals.
- The only reason I didn't change to importing `spacepy.datamodel` and using the namespace is that it hit a lot of different spots in the file. So basically laziness... I totally dig namespaces. I can update that. There are also a lot of repeated imports within functions, and they could probably largely be pulled out. The question is whether that belongs in this PR.

jtniehof:
Unless @spacecataz thinks it's important, I think read-on-init, remove the f attribute, and close the file is fine. Probably include a note in the docstring describing the change.

Given the style of the rest of the module, probably stick with the import from, no sense reworking all of that in this PR. At some point we should set explicit style and rip through the entire codebase.

spacecataz:
Hi, all.  Catching up on this-

My experience from a long time ago is that sat files CAN get large, but @drsteve's estimates make me feel better about that.  I think that keeping the read-on-init method is fine and if it becomes and issue, we can create a new ticket/PR.  

I support the other suggestions (namespacing, startswith) and am happy with everything else.

drsteve:
It's been a while, but I think I hit everything in the review with the latest updates.

drsteve:
Okay, I rewrote the branch history to remove my CHANGELOG update, rebased the branch against an update master, and then re-added my CHANGELOG updates (actually a cherry-pick with a conflict fix). Then I force-pushed to update this PR branch.

It looks like everything is lined up properly to me.

PR #273, opened 2020-01-30
pybats.bats: Fix tolerance passing in add_b_magsphere

### LINKED ISSUE #272: Open/closed boundary tolerance not passed to add_b_magsphere
My issue is that tolerance for finding the last closed field line can be set in spacepy.pybats.bats.Bats2d.find_earth_lastclosed, but **not** when doing the
same action via spacepy.pybats.bats.Bats2d.add_b_magsphere().  A kwarg exists for
setting the tolerance, but it is not passed through to find_earth_lastclosed.

**A patch is being pushed through for this.**

```
import matplotlib.pyplot as plt; plt.ion()
from spacepy.pybats import bats
mhd = bats.Bats2d('python/spacepy/tests/data/pybats_test/y0_binary.out')
mhd.add_b_magsphere(tol=0.5)
mhd.add_b_magsphere(tol=0.0001)
```

### Error message/Traceback:

Plots should be distinctly different, but they're not.  

Darwin-18.7.0-x86_64-i386-64bit
sys.version_info(major=3, minor=7, micro=5, releaselevel='final', serial=0)
numpy=1.17.4
scipy=1.3.3
matplotlib=3.1.2

-->

### Version of SpacePy
dev version

spacecataz:
Closes #272 

Small fix to properly pass tolerance kwarg within tracing routines.

## PR Checklist
- [ x] New code has inline comments where necessary
- [ x] Any new modules, functions or classes have docstrings consistent with SpacePy style
- [x ] Added an entry to CHANGELOG if fixing a major bug or providing a major new feature

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -17,6 +17,7 @@ pycdf
 pybats
  - Updated gitm.GitmBin reader to fix failures in Python 3.x.
  - Fix for incorrect Kyoto Kp plots.
+ - Fixed open/closed tolerance kwarg not working in add_b_magsphere.
 toolbox
  - New function get_url to retrieve data from a URL
  - Change update to retrieve OMNI2 data from SPDF not ViRBO
--- a/spacepy/pybats/bats.py
+++ b/spacepy/pybats/bats.py
@@ -1632,7 +1632,7 @@ class Bats2d(IdlFile):
 
         # Start by finding open/closed boundary.
         tilt, thetaD, thetaN, last1, last2 = self.find_earth_lastclosed(
-            method=method)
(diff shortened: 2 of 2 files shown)


jtniehof:
At some point we need to figure out how to unit test our plotting routines.

spacecataz:
Well, we could test against attributes of figure, axes, and artist
objects.   That would at least guarantee uniformity across platforms.
Some things are subjective, so we'll never be able to test them- this is
something that would have passed everywhere and only came up because I
found a case where closed field lines were appearing outside of the
last-closed boundary.  It was a pretty unique edge case...

Issue #52, opened 2018-11-13
Convert to astropy.time


jtniehof:
This is a big "maybe". It may be worth replacing ticktock or rebasing it on astropy.time (http://docs.astropy.org/en/stable/time/). Open for discussion here....

drsteve:
It looks like the underlying library behind that has very similar
functionality to LANLGeoMag's time stuff. I do note that leap seconds are
hardcoded in their implementation, so can't be trivially updated by the
user.

I'm happy to compare/contrast, but if we're going to wrap C I think we
should take a look at LGM too.

jtniehof:
I absolutely would love to switch over to LGM for coordinates, and if we get time for it reasonably, so much the better. AstroPy time obviously has more people using it but hardcoded leapseconds are a huge bummer.

drsteve:
Remember that an overhaul of coordinates is on my list, and I'm looking at
LGM there. It might be easiest to pull out the time and coordinates
together from LGM and wrap that. We should talk when I get back.

jtniehof:
Yes, we should totally do that...I'm talking to a plasmapy guy here about coordinates, too.

jtniehof:
Hmmm, I wonder if it makes sense to steal the AstroPy time API so people can use either one with similar code.

jtniehof:
I've been told astropy has a coordinate transformation framework into which you can register other conversion routines. Might be worth supporting.

ayshih:
SunPy makes use of Astropy's coordinates framework, which we extend with coordinate frames relevant to solar physics.  Each coordinate frame can be transformed to any other coordinate frame in either SunPy or Astropy.  Here's an illustrative transformation graph:
![transform_whatsnew](https://user-images.githubusercontent.com/991759/76759908-02864a80-6763-11ea-88d6-f758aeceac64.png)

PR #286, opened 2020-03-10
Added Coords methods to convert to/from Astropy SkyCoord instances

ayshih:
For your consideration, this PR enables interacting back and forth with the Astropy coordinates framework (which is also used by SunPy).  It adds the following methods to `Coords`:

- `to_skycoord()` is a regular method which creates an Astropy `SkyCoord` instance from a `Coords` instance
- `from_skycoord()` is a class method which creates a `Coords` instance from an Astropy `SkyCoord` instance

The conversion utilizes the GEO coordinate frame, which is common to both libraries.

### DIFF
--- a/spacepy/coordinates.py
+++ b/spacepy/coordinates.py
@@ -13,6 +13,7 @@ Copyright 2010 Los Alamos National Security, LLC.
 import numpy as np
 from spacepy import help
 import spacepy
+import spacepy.time
 
 __contact__ = 'Steven Morley, smorley@lanl.gov'
 
@@ -81,9 +82,16 @@ class Coords(object):
     .. autosummary::
         ~Coords.append
         ~Coords.convert
+        ~Coords.from_skycoord
+        ~Coord.to_skycoord
     .. automethod:: append
     .. automethod:: convert
+    .. automethod:: from_skycoord
+    .. automethod:: to_skycoord
     '''
+
+    Re = 6371200.0  # meters
+
     def __init__(self, data, dtype, carsph, units=None, ticks=None):
 
         from . import irbempy as op
@@ -129,7 +137,6 @@ class Coords(object):
         self.dtype = dtype
         self.carsph = carsph
         # setup units
-        self.Re = 6371000.0 #metres
         if units is None and carsph == 'car':
             # use standard units
             self.units = ['Re', 'Re', 'Re']
@@ -391,6 +398,71 @@ class Coords(object):
         data.extend(list(otherdata.data))
         newobj = Coords(data, dtype=self.dtype, carsph=self.carsph)
         return newobj
+
+    # -----------------------------------------------
+    def to_skycoord(self):
+        '''
+        Create an Astropy SkyCoord instance based on this instance
+
+        Returns
+        =======
+        out : `astropy.coordinates.SkyCoord`
+            This coordinate as an Astropy SkyCoord
+
+        Notes
+        =====
+        This method requires Astropy to be installed.
+
+        This method uses the GEO coordinate frame as the common frame between the two libraries.
+        '''
+        if self.ticks is None:
+            raise ValueError("This method requires the attribute `ticks` to be specified.")
+
+        try:
+            import astropy.coordinates
+            import astropy.time
+        except ImportError as e:
+            raise e.__class__("This method requires Astropy to be installed.")
+
+        coord = self.convert('GEO', 'car')  # GEO is Astropy ITRS
+        data = coord.data * self.Re  # convert to meters
+        obstime = coord.ticks.APT
+
+        return astropy.coordinates.SkyCoord(x=data[:, 0], y=data[:, 1], z=data[:, 2],
+                                            unit='m', frame='itrs', obstime=obstime)
+
+    # -----------------------------------------------
+    @classmethod
+    def from_skycoord(cls, skycoord):
+        '''
+        Create a Coords instance from an Astropy SkyCoord instance
+
+        Parameters
+        ==========
+        skycoord : `astropy.coordinates.SkyCoord`
+            The coordinate to be converted
+
+        Returns
+        =======
+        out : Coords instance
+            The converted coordinate
+
+        Notes
+        =====
+        This method requires Astropy to be installed.
+
+        This method uses the GEO coordinate frame as the common frame between the two libraries.
+        '''
+        try:
+            import astropy.coordinates
+        except ImportError as e:
+            raise e.__class__("This method requires Astropy to be installed.")
+
+        skycoord = astropy.coordinates.SkyCoord(skycoord).itrs  # Astropy ITRS is GEO
+        data = (skycoord.cartesian.xyz.to_value('m') / cls.Re).T  # convert Cartesian to Re units
+        ticks = spacepy.time.Ticktock(skycoord.obstime)
+
+        return cls(data, 'GEO', 'car', ticks=ticks)
 
 
 def quaternionNormalize(Qin, scalarPos='last'):
--- a/spacepy/coordinates.py
+++ b/spacepy/coordinates.py
@@ -459,7 +459,7 @@ class Coords(object):
             raise e.__class__("This method requires Astropy to be installed.")
 
         skycoord = astropy.coordinates.SkyCoord(skycoord).itrs  # Astropy ITRS is GEO
-        data = (skycoord.cartesian.xyz.to_value('m') / cls.Re).T  # convert Cartesian to Re units
+        data = (skycoord.cartesian.xyz.to('m').value / cls.Re).T  # convert Cartesian to Re units
         ticks = spacepy.time.Ticktock(skycoord.obstime)
 
         return cls(data, 'GEO', 'car', ticks=ticks)
--- a/tests/test_coordinates.py
+++ b/tests/test_coordinates.py
@@ -76,7 +76,7 @@ class coordsTest(unittest.TestCase):
         np.testing.assert_allclose(sc_data, self.cvals.data * self.cvals.Re, rtol=1e-10)
 
         # Check that the time was loaded correctly
-        assert np.all(sc.obstime.isclose(self.cvals.ticks.APT))
+        np.testing.assert_allclose((sc.obstime - self.cvals.ticks.APT).to('s').value, 0)
 
     @unittest.skipUnless(HAVE_ASTROPY, 'requires Astropy')
     def test_to_skycoord_with_ticks_and_conversion(self):
@@ -95,7 +95,7 @@ class coordsTest(unittest.TestCase):
         np.testing.assert_allclose(sc_data, self.cvals.data * self.cvals.Re, rtol=1e-10)
 
         # Check that the time was loaded correctly
-        assert np.all(sc.obstime.isclose(self.cvals.ticks.APT))
+        np.testing.assert_allclose((sc.obstime - self.cvals.ticks.APT).to('s').value, 0)
 
     @unittest.skipUnless(HAVE_ASTROPY, 'requires Astropy')
     def test_from_skycoord(self):
@@ -110,7 +110,7 @@ class coordsTest(unittest.TestCase):
         np.testing.assert_allclose(coords.data * coords.Re, sc_data, rtol=1e-10)
 
         # Check that the time was loaded correctly
-        assert np.all(sc.obstime.isclose(coords.ticks.APT))
+        np.testing.assert_allclose((sc.obstime - coords.ticks.APT).to('s').value, 0)
 
     @unittest.skipUnless(HAVE_ASTROPY, 'requires Astropy')
     def test_from_skycoord_with_conversion(self):
@@ -126,7 +126,7 @@ class coordsTest(unittest.TestCase):
         np.testing.assert_allclose(coords.data * coords.Re, sc_data, rtol=1e-10)
 
         # Check that the time was loaded correctly
-        assert np.all(sc.obstime.isclose(coords.ticks.APT))
+        np.testing.assert_allclose((sc.obstime - coords.ticks.APT).to('s').value, 0)
 
 
 class QuaternionFunctionTests(unittest.TestCase):
--- a/tests/test_coordinates.py
+++ b/tests/test_coordinates.py
@@ -72,7 +72,7 @@ class coordsTest(unittest.TestCase):
         assert sc.frame.name == 'itrs'
 
         # Check that the data was loaded correctly
-        sc_data = sc.cartesian.xyz.to_value('m').T
+        sc_data = sc.cartesian.xyz.to('m').value.T
         np.testing.assert_allclose(sc_data, self.cvals.data * self.cvals.Re, rtol=1e-10)
 
         # Check that the time was loaded correctly
@@ -91,7 +91,7 @@ class coordsTest(unittest.TestCase):
         assert sc.frame.name == 'itrs'
 
         # Check that the data converts back correctly
-        sc_data = sc.cartesian.xyz.to_value('m').T
+        sc_data = sc.cartesian.xyz.to('m').value.T
         np.testing.assert_allclose(sc_data, self.cvals.data * self.cvals.Re, rtol=1e-10)
 
         # Check that the time was loaded correctly
@@ -106,7 +106,7 @@ class coordsTest(unittest.TestCase):
         coords = spc.Coords.from_skycoord(sc)
 
         # Check that the data was loaded correctly
-        sc_data = sc.cartesian.xyz.to_value('m').T
+        sc_data = sc.cartesian.xyz.to('m').value.T
         np.testing.assert_allclose(coords.data * coords.Re, sc_data, rtol=1e-10)
 
         # Check that the time was loaded correctly
@@ -122,7 +122,7 @@ class coordsTest(unittest.TestCase):
         coords = spc.Coords.from_skycoord(sc.gcrs)
 
         # Check that the data was loaded correctly
-        sc_data = sc.cartesian.xyz.to_value('m').T
+        sc_data = sc.cartesian.xyz.to('m').value.T
         np.testing.assert_allclose(coords.data * coords.Re, sc_data, rtol=1e-10)
 
         # Check that the time was loaded correctly
--- a/tests/test_coordinates.py
+++ b/tests/test_coordinates.py
@@ -12,6 +12,11 @@ try:
     import spacepy.irbempy as ib
 except ImportError:
     pass #tests will fail, but won't bring down the entire suite
+try:
+    import astropy.coordinates
+    HAVE_ASTROPY = True
+except ImportError:
+    HAVE_ASTROPY = False
 import spacepy.toolbox as tb
 
 __all__ = ['coordsTest']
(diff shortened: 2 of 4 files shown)


drsteve:
Thanks for the PR @ayshih 

Tests are currently failing on this PR because of a broken URL for the NASA CDF library in our `.travis.yml` file (fixed in #288). If you can rebase against spacepy/master and push the missing commit then the Travis CI will be able to run.

While I'm supportive of interoperability between packages in the wider helio community, I'm also hesitant to approve this without a strategy for testing. Any thoughts? (Also paging @jtniehof for thoughts)

On the actual implementation:
- I'd probably not go via ISO time strings, for performance reasons, and would prefer to pass times using a numerical system that's shared between astropy and spacepy.
- There should also be appropriate checking for the presence of astropy so that spacepy returns a tailored error message with the `ImportError` if the dependency isn't present.
- The Earth radius used by IRBEM is present as an attribute (`Re`) on the `spacepy.coordinates.Coords` class, so there's no need to hard-code it here (although I note that it's not used anywhere, and incorrectly gives 6371000 m rather than 6371200 m). It's an instance attribute, not a class attribute, so won't be available until instantiation. That would either need a reorder of how the `Coords` instance is made/populated, or the `Re` attribute could be moved so it's a class attribute. I'd suggest having `Re` as a class method with the number corrected to 6371200.
- Single letter names in imports are strongly discouraged.

Note that it's on our plan to overhaul the coordinates module in SpacePy. Whenever that happens it should be transparent to users, but will provide higher accuracy conversions for the geophysical systems (see #52 ; also https://github.com/drsteve/LANLGeoMag/blob/master/libLanlGeoMag/Lgm_CTrans.c). Ideally I think we'd have an internal, well-tested conversion code (probably based on LANLGeoMag) and interoperability with other packages. That's still up for discussion though.

ayshih:
> Tests are currently failing on this PR because of a broken URL for the NASA CDF library in our `.travis.yml` file (fixed in #288). If you can rebase against spacepy/master and push the missing commit then the Travis CI will be able to run.

Rebased (but I haven't yet added tests for these enhancements).

> While I'm supportive of interoperability between packages in the wider helio community, I'm also hesitant to approve this without a strategy for testing. Any thoughts? (Also paging @jtniehof for thoughts)

My suggestion is that the associated unit tests would be purely to check whether the Astropy SkyCoords are created properly and read in properly.  It seems overly complicated to do more than that (e.g., testing Astropy's transformations within your tests).

> * I'd probably not go via ISO time strings, for performance reasons, and would prefer to pass times using a numerical system that's shared between astropy and spacepy.

This was a conscious choice on my part because the ISO time string functionally retains a lot of precision.  Astropy Time supports many other scales (JD, MJD, TAI, etc.), but there may be precision loss when you go to extremes.

> * There should also be appropriate checking for the presence of astropy so that spacepy returns a tailored error message with the `ImportError` if the dependency isn't present.

Done (with a little bit of extra contortion now that `ModuleNotFoundError` exists in Python 3.6+)

> * The Earth radius used by IRBEM is present as an attribute (`Re`) on the `spacepy.coordinates.Coords` class, so there's no need to hard-code it here (although I note that it's not used anywhere, and incorrectly gives 6371000 m rather than 6371200 m). It's an instance attribute, not a class attribute, so won't be available until instantiation. That would either need a reorder of how the `Coords` instance is made/populated, or the `Re` attribute could be moved so it's a class attribute. I'd suggest having `Re` as a class attribute with the number corrected to 6371200.

Done (I wasn't sure whether the wrong value for `Re` was being used anywhere)

> * Single letter names in imports are strongly discouraged.

It's actually "standard" practice to use `u` for `astropy.units`, but I understand the resistance.  I've revised the code so I don't even need to explicitly import it.

> Note that it's on our plan to overhaul the coordinates module in SpacePy. Whenever that happens it should be transparent to users, but will provide higher accuracy conversions for the geophysical systems (see #52 ; also https://github.com/drsteve/LANLGeoMag/blob/master/libLanlGeoMag/Lgm_CTrans.c). Ideally I think we'd have an internal, well-tested conversion code (probably based on LANLGeoMag) and interoperability with other packages. That's still up for discussion though.

Great!  I've put a huge amount of effort into SunPy's coordinate framework, and I look forward to pairing it with a high-accuracy geophysical framework.

spacepy/coordinates.py
```diff
@@ -391,3 +397,81 @@ def append(self, other):
...
+        =====
+        This method requires Astropy to be installed.
+
+        This method uses the GEO coordinate frame as the common frame between the two libraries.
+        '''
+        try:
+            from astropy.coordinates import SkyCoord
+            from astropy.time import Time
+        except ImportError as e:
+            raise e.__class__("This method requires Astropy to be installed.")
```
Cadair: It's good to propagate the error forward:
```suggestion
            raise e.__class__("This method requires Astropy to be installed.") from e
```

spacepy/coordinates.py
```diff
@@ -391,3 +397,81 @@ def append(self, other):
...
+        Notes
+        =====
+        This method requires Astropy to be installed.
+
+        This method uses the GEO coordinate frame as the common frame between the two libraries.
+        '''
+        try:
+            from astropy.coordinates import SkyCoord
+        except ImportError as e:
+            raise e.__class__("This method requires Astropy to be installed.")
```
Cadair:
```suggestion
            raise e.__class__("This method requires Astropy to be installed.") from e
```

spacepy/coordinates.py
ayshih: Done

spacepy/coordinates.py
ayshih: Done

spacepy/coordinates.py
ayshih: Reverted.  SpacePy still tests against Python 2.7, where this is not valid syntax.

spacepy/coordinates.py
ayshih: Reverted.  SpacePy still tests against Python 2.7, where this is not valid syntax.

ayshih:
@drsteve I've now added tests.  Any further comments?

In the future, when you overhaul the transformation code (using LGM or otherwise), I would strongly advocate that you take that opportunity to incorporate native support for `astropy.time` and `astropy.coordinates`, because that would make transformations frictionless across libraries.

Cadair:
I would second that, it would be great if we had more common object classes across all of heliophysics.

spacepy/coordinates.py
```diff
@@ -391,3 +397,78 @@ def append(self, other):
...
+        if self.ticks is None:
+            raise ValueError("This method requires the attribute `ticks` to be specified.")
+
+        try:
+            from astropy.coordinates import SkyCoord
+            from astropy.time import Time
+        except ImportError as e:
+            raise e.__class__("This method requires Astropy to be installed.")
+
+        # Convert to GEO (i.e., Astropy's ITRS)
```
jtniehof, line 426: This style of comment, statement, blank line soaks up vertical space and isn't adding much...it's obvoius it's converting to GEO, just need the ITRS info. Since the github multi-line suggestion thing doesn't work well, I'm envisioning something like:

```
coord = self.convert('GEO', 'car') # Astropy ITRS is GEO
data = coord.data * self.Re # metres
obstime = astropy.time.Time(coord.ticks.ISO, format='isot', scale='utc')
```

As @drsteve suggested, maybe use, say, TAI instead of ISO ticks?

spacepy/coordinates.py
```diff
@@ -391,3 +397,78 @@ def append(self, other):
...
+        This method requires Astropy to be installed.
+
+        This method uses the GEO coordinate frame as the common frame between the two libraries.
+        '''
+        try:
+            from astropy.coordinates import SkyCoord
+        except ImportError as e:
+            raise e.__class__("This method requires Astropy to be installed.")
+
+        # Cast the input as a SkyCoord and then transform to Astropy's ITRS (i.e., GEO)
```
jtniehof, line 465: Similar to above

spacepy/coordinates.py
```diff
@@ -391,3 +397,78 @@ def append(self, other):
...
+        skycoord = SkyCoord(skycoord).itrs
+
+        # Convert the SkyCoord's data to a Cartesian representation with units of R_earth
+        data = (skycoord.cartesian.xyz.to_value('m') / cls.Re).T
+
+        # Convert the SkyCoord's obstime to a Ticktock instance
+        ticks = spacepy.time.Ticktock(skycoord.obstime.utc.isot, 'ISO')
+
+        return Coords(data, 'GEO', 'car', ticks=ticks)
```
jtniehof, line 474:
```suggestion
        return cls(data, 'GEO', 'car', ticks=ticks)
```

tests/test_coordinates.py
```diff
@@ -11,6 +11,10 @@
     import spacepy.irbempy as ib
 except ImportError:
     pass #tests will fail, but won't bring down the entire suite
+try:
```
jtniehof, line 14: I wonder about using a skip decorator on these, since astropy is optional. The reason for just forging ahead if irbem didn't import is because we still want to let other tests go ahead (to get that feedback in case there were other things wrong) but that's still a really pathalogical case, didn't seem worth skip-decorating the whole file. But for astropy coords, somebody might benefit from having a clean test run even without astropy as an optional dependency.

spacepy/coordinates.py
drsteve: There's a _big_ nomenclature difference between TAI as used in key sections of the space physics community, and TAI as used by the broader astronomical and astrodynamics community (which is the correct one, of course). We use the "incorrect" version because it's fairly well entrenched in a number of key areas of magnetospheric physics. Our "TAI" is integer seconds since 1958/1/1, where the _proper_ TAI is offset from UTC by the number of leapseconds.
For a numeric time system, I'd probably use MJD. Although I think UTC datetimes get generated for pretty much any input time system, so it's probably least overhead to use those rather than forcing a conversion. If a conversion is desired, then MJD is my preference.
ISO strings are fine by me as long as we don't think the performance hit will be an issue.

TL;DR - MJD or UTC datetimes are my preference. ISO is fine if the performance hit is negligible.

spacepy/coordinates.py
ayshih: Changed

tests/test_coordinates.py
ayshih: Tests now decorated with `skipUnless`

spacepy/coordinates.py
```diff
@@ -391,3 +397,78 @@ def append(self, other):
...
         return newobj
+
+    # -----------------------------------------------
+    def to_skycoord(self):
+        '''
+        Create an Astropy SkyCoord instance based on this instance
+
+        Returns
+        =======
+        out : Astropy SkyCoord instance
```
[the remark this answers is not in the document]
ayshih: Changed to intersphinx-style notation since you'll probably end up there eventually.

spacepy/coordinates.py
```diff
@@ -391,3 +397,78 @@ def append(self, other):
...
+        =====
+        This method requires Astropy to be installed.
+
+        This method uses the GEO coordinate frame as the common frame between the two libraries.
+        '''
+        if self.ticks is None:
+            raise ValueError("This method requires the attribute `ticks` to be specified.")
+
+        try:
+            from astropy.coordinates import SkyCoord
```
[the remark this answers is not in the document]
ayshih: Changed here and elsewhere

spacepy/coordinates.py
ayshih: Condensed

spacepy/coordinates.py
ayshih: So...  It turns out that how you handle ISO, MJD, and UTC are all problematic because I believe they all repeat at the leap second.  After some experimentation, I believe the most sensible approach is to define a custom format for Astropy Time that matches your definition for "TAI".  That way, there is no degeneracy at the leap second and conversion overhead is kept to a minimum.

This custom time format is defined at the bottom of `spacepy.coordinates` to keep it conceptually grouped with the SkyCoord conversion functions, but of course it may make sense to move it to `spacepy.time`.

Example:
```python
>>> from spacepy.time import Ticktock
>>> Ticktock(1861920037.0, 'TAI').ISO
dmarray(['2017-01-01T00:00:00'], dtype='<U19')

>>> import spacepy.coordinates
>>> import astropy.time
>>> Time('2017-01-01 00:00:00', scale='utc').spacepy_tai
1861920037.0
>>> Time(1861920037.0, format='spacepy_tai').utc.iso
'2017-01-01 00:00:00.000'
```

ayshih:
> Can you add Astropy, including the minimum version, to the soft dependencies in dependencies.rst?

Done

jtniehof:
@drsteve and I looked over the time situation. Generally, as long as you're using a time system that can count leap seconds appropriately, Ticktock should be handling them properly. It's possible that some transforms might pass through intermediate representations that are losing leap seconds, in which case we'd like to identify and fix those.

Taking the leap second at the end of 2016, if I take an ISO timestamp for
the last non-leap second, it's two seconds to the beginning of 2017:
```
>>> tt = spacepy.time.Ticktock(['2016-12-31T23:59:59',
'2017-01-01T00:00:00'])
>>> tt[1] - tt[0]
[datetime.timedelta(0, 2)]
```

Contrasted with a year without a leap second:
```
>>> tt = spacepy.time.Ticktock(['2017-12-31T23:59:59',
'2018-01-01T00:00:00'])
>>> tt[1] - tt[0]
[datetime.timedelta(0, 1)]
```

It's true that this is a bit weird:
```
>>> spacepy.time.Ticktock('2016-12-31T23:59:59') +
datetime.timedelta(seconds=1)
Ticktock( [datetime.datetime(2017, 1, 1, 0, 0)], dtype=UTC)
```

But note the dtype is "UTC" and the SpacePy Ticktock handling of UTC is with datetimes, not really a true UTC. So we're looking into addition using objects other than timedeltas (#317).

If we hand in a TAI it does get handled properly:
```
>>> spacepy.time.Ticktock('2016-12-31T23:59:59').TAI[0]
1861920035.0
>>> tt = spacepy.time.Ticktock([1861920035, 1861920036, 1861920037],
dtype='TAI')
>>> tt.UTC
dmarray([datetime.datetime(2016, 12, 31, 23, 59, 59),
         datetime.datetime(2017, 1, 1, 0, 0),
         datetime.datetime(2017, 1, 1, 0, 0)], dtype=object)
```
UTC looks bad, but that's because datetime doesn't handle it. We do still
have the correct TAI:
```
>>> tt.TAI
dmarray([1861920035, 1861920036, 1861920037])
```

Now there does seem to be a problem with MJD:
```
>>> tt.MJD
dmarray([ 57753.99998843,  57754.        ,  57754.        ])
```

This is because our JD and MJD handling is touching Python datetime. So MJD is probably the right way to go but we're going to fix it first (#319)

The difference in "space physics" TAI is that it's just represented as a number of seconds rather than a timestamp, and as @drsteve said it's relative to the 1958-01-01 epoch.

I haven't dug deeply into AstroPy time, but working from the docs, looks like the easiest way to get a SpacePy TickTock into AstroPy is via GPS time, since AstroPy uses gps as a count of seconds in TAI time scale, just with a different reference epoch. So if, using our example above, you do:
```
>>> astropy_tt = astropy.time.Time(tt.TAI - 694656019.0, format='gps')
```

Then you can see it's converted properly:
```
>>> astropy_tt.format='isot'
>>> astropy_tt.utc
<Time object: scale='utc' format='isot' value=['2016-12-31T23:59:59.000'
'2016-12-31T23:59:60.000'
 '2017-01-01T00:00:00.000']>
```

Ticktock does do GPS time, but it does enough work through datetime that the conversion doesn't work properly here, on the list to fix (#320).

So medium-term: SpacePy will fix MJD and GPS time to eschew datetime and then one of those can be used. In the short-term, TAI plus an offset works. This should be a lot more straightforward than introducing a new subclass.

But we think it's better to separate out the time conversion into Ticktock, and should be able to get to that fairly directly (#321). Once that's done we'll loop back to this PR.

Thanks for bringing the time issues up. This was definitely a place that could use some polishing.

jtniehof:
Since #417 is now merged, SpacePy's Ticktock supports representation as AstroPy time. So at a high level, this PR needs to be rebased to current master (conflict-resolved) and switched to using the Ticktock support. There are likely other small things to check but those are the biggies.

@ayshih , what's your preference for how to proceed? I can leave this all in your court, or I could do the rebase just to make sure the untangling goes well, or I could do the rebase and also the switchover of the time.

ayshih:
Sure, I'll update this PR (although it might take me a little while to get to it).

A point of contention may be the minimum version of Astropy that will be supported, which you set in #417 as Astropy 1.0.  While your support plan is laudable, I think supporting 1.0, 1.1.1 (for Ubuntu 18.04 LTS), or even 2.0.4 (for Ubuntu 20.04 LTS) is risky for `astropy.coordinates`, and may be risky even for `astropy.time`.  Such ancient versions might work from the API perspective (i.e., errors are not raised), but there may be issues with the accuracy of calculations using those versions because of bugs that have been fixed in subsequent point releases (e.g., 2.0.16).

In this PR, I have currently listed Astropy 3.2 as the minimum version.  I know that Astropy 3.0+ may be problematic for SpacePy until you drop Python 2 support, but I'd recommend at least 2.0.x (preferably a point release more recent than 2.0.4).

jtniehof:
As long as the code builds a reasonable time/coord object with AstroPy 1.0, I don't want to take the responsibility for whether the user might later take it into places that aren't suitable. Someone may genuinely get along fine with the earlier versions and no sense in blocking that.

ehsteve:
@ayshih what are the plans on getting this finished up? Inquiring minds want to know ;)

ayshih:
@ehsteve Soon™?

I'll see if I can give myself a solid kick on this next week.

ayshih:
Okay, I *finally* got around to updating this PR.  Please review!

spacepy/coordinates.py
Cadair: Does it still test against 2.7?

ayshih: As far as I can tell, SpacePy still tries to support 2.7

drsteve:
Thanks for the update!

It looks like all of the test errors here are the same cause:
So that folks don't have to dig too deep in the GitHub Actions logs, here's one of the tracebacks
```
======================================================================
ERROR: test_to_skycoord_with_ticks (test_coordinates.coordsTest)
----------------------------------------------------------------------
  File "/home/runner/work/spacepy/spacepy/tests/test_coordinates.py", line 79, in test_to_skycoord_with_ticks
    assert np.all(sc.obstime.isclose(self.cvals.ticks.APT))
  File "/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/site-packages/astropy/time/core.py", line 1152, in __getattr__
    return self.__getattribute__(attr)
AttributeError: 'Time' object has no attribute 'isclose
```

The line `assert np.all(sc.obstime.isclose(self.cvals.ticks.APT))` in most of the new tests will fail on any astropy version prior to 4.2 (released on 2020-11-24), according to the astropy `CHANGELOG`.

For testing purposes, it'd probably make sense to do an explicit test of the times rather than using a new `astropy.time.Time` method. After all, the absence of the new `isclose` method won't affect whether the conversion worked or not. My suggestion would be to check that the _scale_, _format_, and _values_ are as expected.

ayshih:
Ah, yes, I shouldn't have used `astropy.time.Time.isclose()`.  Rather than checking `scale`/`format`/`values` explicitly, I now check that the time difference is zero.

ayshih:
> Based on the test failures and the astropy changelog, I think we're looking at requiring a bump in minimum supported version of astropy to `>=2.0` (or working around the `to_value` call in `from_skycoord`).

I think it'd be prudent to bump your minimum version of Astropy regardless, but I've replaced the `to_value()` call with something that should work on all versions of Astropy.

spacepy/coordinates.py
```diff
@@ -459,7 +459,7 @@ def from_skycoord(cls, skycoord):
             raise e.__class__("This method requires Astropy to be installed.")
 
         skycoord = astropy.coordinates.SkyCoord(skycoord).itrs  # Astropy ITRS is GEO
-        data = (skycoord.cartesian.xyz.to_value('m') / cls.Re).T  # convert Cartesian to Re units
+        data = (skycoord.cartesian.xyz.to('m').value / cls.Re).T  # convert Cartesian to Re units
```
drsteve, line 462: Thanks for that change.

The CI cancelled the tests on 3.5 because they're running on an Ubuntu 16.04 image - don't worry about that. The failures look like they need this same change made on lines 75, 94, 109 & 125 of `test_coordinates.py`, but I'll let you double-check that...

ayshih: Oops!  Changes included now

drsteve (approved):

spacepy/coordinates.py
```diff
@@ -392,6 +399,71 @@ def append(self, other):
...
+            raise ValueError("This method requires the attribute `ticks` to be specified.")
+
+        try:
+            import astropy.coordinates
+            import astropy.time
+        except ImportError as e:
+            raise e.__class__("This method requires Astropy to be installed.")
+
+        coord = self.convert('GEO', 'car')  # GEO is Astropy ITRS
+        data = coord.data * self.Re  # convert to meters
```
jtniehof, line 428: This assumes the input coordinates are in Re, but they can be (e.g.) km. We're not very good about handling this elsewhere--for instance, irbempy calls assume they're Re regardless of the units attribute--but it's probably worth doing something.

The fairly fancy approach would be to support Re, km, m and throw an error if they 1) aren't the same unit in all dimensions 2) not one of the above. I don't want to overburden this PR with fanciness. @drsteve , thoughts?

drsteve: My thought is that that's out of scope here. The documentation isn't as clear as it could be, but all Cartesian coordinate systems are assumed to be in Re, with spherical systems in degrees and Re. The only one that's different is geodetic, which is assumed/required to be in degrees and km. There's checking for that in #296, when we finally get to working through that (or breaking it for parts). The units are currently there to tell the user, not to let the user switch...

So this PR does the right thing for a correctly specified `Coords` per documentation. I'm happy as is.

jtniehof (approved):
Looks great, thanks for pushing through with this!

ayshih:
Wonderful, thanks for the merge!

Issue #1, opened 2015-03-04
When using spacepy.coordinates.Coords, getting 'ImportError: cannot import name irbempylib'


johncoxon:
I have installed spacepy 0.1.5 using `pip install spacepy` and then done `from spacepy import time as spt, coordinates as spc`. I attempt to create a `Coords` object by writing `my_gse_coords = spc.Coords(my_coords, 'GSE', 'car')`, where `my_coords` is a numpy array of shape (n, 3). Python fails on this line and I get an error: `ImportError: cannot import name irbempylib`. The full traceback is below.

	~/anaconda/lib/python2.7/site-packages/spacepy/coordinates.pyc in __init__(self, data, dtype, carsph, units, ticks)
		 77     def __init__(self, data, dtype, carsph, units=None, ticks=None):
		 78 
	---> 79         from . import irbempy as op
		 80         from spacepy.irbempy import SYSAXES_TYPES as typedict
		 81 

	~/anaconda/lib/python2.7/site-packages/spacepy/irbempy/__init__.py in <module>()
		 68 
		 69 
	---> 70 from .irbempy import *
		 71 

	~/anaconda/lib/python2.7/site-packages/spacepy/irbempy/irbempy.py in <module>()
		 21 import spacepy.coordinates as spc
		 22 import spacepy.datamodel as dm
	---> 23 from . import irbempylib as oplib
		 24 import spacepy.toolbox as tb
		 25 

	ImportError: cannot import name irbempylib

Am I doing something wrong, or is this just a bug in spacepy?

graingert:
I'm going to have to point you to https://sourceforge.net/p/spacepy/bugs/?source=navbar

My suspicion is that irbem failed to compile

https://github.com/spacepy/spacepy/blob/master/setup.py#L285

Not sure why there's a mix of relative and absolute imports there though...

johncoxon:
Ah, I installed spacepy before I'd done anything with Fortran on my machine, so uninstalling and reinstalling did the trick. Thanks for the assist!

PR #294, opened 2020-04-17
Updated PR template to be clearer and fix issue references

drsteve:
I didn't make an issue for this, so what does this do?
- In the old PR template the text gave an example `This PR resolves issue #1`, which doesn't link the issue because `issue` is between the magic word and the actual issue reference. This has been updated to show correct syntax.
- Experience suggests that the PR template doesn't really give clear instructions on what to write. As far as I'm concerned deviation from the template is fine as long as the information is present and the boxes have been checked (figuratively speaking), but that assumes the pull-requestor has a good idea of what we need to see in the PR description. To that end I added some of our suggestions as checklist items.
- I've given explicit direction in the updated PR template to replace the preamble text from the template with a PR description. Can't hurt to be more direct about that, right?
- Typo fixed in the old template.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,17 +1,24 @@
 ## PR Checklist
 New code is easier to review, integrate and maintain if it's
 consistent with the style of the rest of the SpacePy code.
-Please try to follow as much of this checkl;ist as you can for
+Please try to follow as much of this checklist as you can for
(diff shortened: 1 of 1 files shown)


spacecataz:
This update looks good; I recommend merging if there are no more comments.

drsteve:
Pinging @balarsen or @jtniehof 
This should be an easy one... just a quick peek and merge would be appreciated (unless you have concerns, obvs)

PR #296, opened 2020-04-23
Implement alternate backend for coordinate transforms

### LINKED ISSUE #302: test coverage: coordinates
At 73%. Kicking this up looks like it shouldn't be too rough.

### LINKED ISSUE #477: Coordinates: Transform to GDZ does not check for/enforce units
Issue raised after user report from Hugh Evans at ESA.

### Issue
Transformation to geodetic coordinates (GDZ) assumes input in Earth radii and returns value in km.
While the other coordinate system transformations are unit-agnostic, to/from GDZ is not.

The documentation does say (coordinates.py, lines 26-27 : docstring for `spacepy.coordinates.Coords`)
```
    A class holding spatial coordinates in Cartesian/spherical
    in units of Re and degrees
```
but there's no enforcement or checking of units, so the user can definitely get bad return values without any signs.

Because they're just matrix-based transformation, all conversions will work with input in km except geographic to geodetic conversions. The GDZ conversions, if given input in the wrong coordinate system, will give bogus results without raising an error. And since units can be taken as a keyword argument, there's a tacit implication that the user can choose which units to use:
```
>>> import spacepy.coordinates as spc
>>> import spacepy.time as spt
>>> tt = spt.Ticktock(2459218.5, 'JD')
>>> pos = [2367.83158, -5981.75882, -4263.24591]

>>> cc = spc.Coords(pos, 'GEI', 'car', ticks=tt, units=['km', 'km', 'km'])
>>> print(cc.units)
['km', 'km', 'km']
>>> cc = spc.Coords(pos, 'GEI', 'car', ticks=tt)  # Remake cc with default units
>>> print(cc.convert('RLL', 'sph'))
>>> print(cc.convert('GDZ', 'sph'))

Coords( [[7717.728335528844, -33.53159979001943, -172.22899523107847]] , 'RLL', 'sph')
Coords( [[49164819.1520793, -33.53159979001943, -172.22899523107847]] , 'GDZ', 'sph')

>>> cc.data /= 6371.2  # Divide to get Re using IRBEM definition of Earth radius
>>> print(cc.convert('GDZ', 'sph'))
Coords( [[1346.134155662018, -33.67800148310953, -172.22899523107847]] , 'GDZ', 'sph')
```

Along with this, even though the default units are different when instantiating `Coords` in GDZ, the units are not set correctly on conversion to GDZ.

```
>>> print(cc.units)
['Re', 'Re', 'Re']

>>> gdz = spc.Coords([1346.134508626895, -33.6790290525164, -172.2309539745096] , 'GDZ', 'sph')
>>> print(gdz.units)

['km', 'deg', 'deg']

>>> print(cc.convert('GDZ', 'sph').units)
['Re', 'deg', 'deg']
```
That is, `convert` is just propagating the distance unit without any checking.

### Versions
Linux-4.19.128-microsoft-standard-x86_64-with-glibc2.10
sys.version_info(major=3, minor=8, micro=5, releaselevel='final', serial=0)
numpy=1.19.2
scipy=1.5.2
matplotlib=3.3.2
spacepy is from HEAD of repo

I don't know what system Hugh was testing on, but this should be reproducible across all systems/versions.

drsteve:
This PR provides an alternate backend for the coordinates module that is python-based. Code here is based on the LANLGeoMag implementations (see #52 ), and is intended to provide a basic set of relevant transforms.

This supports future SpacePy organization into core and sub-packages, where the core package doesn't need to have IRBEMlib with it. That is, coordinate transforms will not have to depend on IRBEM. ~This PR doesn't cover astropy interoperability (e.g., #286 )), but it should be easy enough to add.~ (Edit: that was merged in, so *should* be present now...) ~I'm just looking to keep the scope narrower for now. This is already going to be a pain to review!~

### What this PR adds
- New backend for coordinate transforms added in the new `ctrans` module
- `ctrans` backend added as a non-default option in `coordinates.Coords`
- Using default "IRBEM" backend gives a `DeprecationWarning`
- Enforces unit-checking on conversions, and clarifies documentation of unit handling.
- Comprehensive testing: `coordinates` is up to 91%; `ctrans` is at 97%; `igrf` is at 95%.

### Coordinate systems supported
- "Core" (do not require magnetic field)
  - Earth Centered Inertial (J2000)  [`irbempy`'s _J2000_]
  - Earth Centered Inertial (Mean of date)
  - Earth Centered Inertial (True of date)  [`irbempy`'s _GEI_/_TOD_]
  - Geocentric Solar Ecliptic  [`irbempy`'s _GSE_]
  - Geocentric Geographic  [`irbempy`'s _GEO_/_SPH_]
  - Geodetic [`irbempy`'s _GDZ_]
  - (Geocentric radius, geodetic lat/long) [`irbempy`'s _RLL_]
- "Magnetic" (use IGRF dipole axis)
  - Geomagnetic  [`irbempy`'s _MAG_]
  - Geocentric Solar Magnetospheric  [`irbempy`'s _GSM_]
  - Solar Magnetic  [`irbempy`'s _SM_]

### Notes
- The overhaul also now enforces units on conversion to/from GDZ, and clarifies the documentation, so this PR closes #477
- Expanded testing bring coordinates test coverage up to 90%, so this PR closes #302 
- `ctrans` is not in the documentation yet, although the module is internally documented
- It's possibly worth expanding the documentation (in `coordinates`, I think) on the Earth radius. IRBEM's value (6371.2km) is non-standard, but close to the volumetric mean radius of 6371km. SpacePy uses WGS84's semi-major axis (6378.137), which is fairly standard and in line with the IAU recommendation of using the equatorial radius as a nominal Earth radius. Astropy uses 6378.1 the IAU's "nominal zero-tide equatorial radius".
- ~`CHANGELOG`~ Release notes are not updated yet. Happy to take suggestions on how best to include the impacts of this PR in there...
- The speed of the new conversions are ~untested~ not adequately benchmarked. For small arrays the time taken is negligible, scaling has been improved with a few structure changes including a tolerance so that times a small delta apart can reuse the transforms (delta is configurable).

PR Checklist:
- [x] Added an entry to ~CHANGELOG~ release notes if fixing a major bug or providing a major new feature


balarsen:
It is worth noting that astropy time is perfectly reasonable to
support/use. Pandas has a lot of the time manipulation as well especially
for string to date time and epoch based systems.

I have trained myself to think like and use pandas and it has drastically
improved my workflow. Worth contemplating it as a dependency.  I have
become a big fan of utilization of the big core packages where it makes
sense to help us not reinvent the wheel. Pandas has very much become of of
those big core packages.

I did a benchmark for the hell of it and pycdf v_epoch_to_datetime beats
pandas which is also interesting.

Brian

drsteve:
At this point, code coverage of `ctrans` is at 92%.
To do:
- Add geodetic coordinates
- Hook into `spacepy.coordinates.Coords` as a backend
   - Make it optional
   - Add deprecation warning for IRBEM backend as default
   - Earth radius definition as option??
- Documentation! (If anyone wants to chip in, feel free)
   - Docstrings in CTrans
   - Update docs in Coordinates (when hooked in)
- CHANGELOG

Currently thinking this will get merged in after 0.2.2. Happy to take thoughts on that too.

jtniehof:
Definitely a post 0.2.2 thing. I'm also just now realizing that you're doing a Python reimplementation of the lanlgeomag code. Is that our best way forward, or should we accelerate switching to Lgm as a backend for everything (coordinates and otherwise)? That would involve some lgmpy work of course and a good solution for the installer, but it would open up a lot.

drsteve:
I thought pretty hard about how to bring in the coordinate transforms and this seemed like least work. LGM is _huge_ and between that, the installer, and unresolved wrapping issues, I think it'll be a giant stack of headache. My current leaning for bringing in LGM is to do that after we've broken out "core" spacepy and then have it as a "space-kit" or whatever we're calling it. I'd love to have some lgmpy work go on, but I think the horizon is too long for usefully bringing it into spacepy.

With the Python reimplementation I can make it consistent with LGM (it's not completely identical, and I'm leaving out a bunch of stuff), and more flexible. And getting it done faster means we can drop IRBEM as a core requirement earlier (and they still have issues of reduced precision and inconsistency between a couple of their coordinate systems). And it's nearly done now anyway!

jtniehof:
Sounds good, thanks.

drsteve:
Okay, this is in functional form now so I've opened it up for review.
Note that I *am* expecting to make changes.
I have also left Astropy support to one side for now to limit the scope of the PR a little. It's big enough already.

As noted in the updated PR description, `CHANGELOG` hasn't been updated and the `ctrans` module is documented but not *in* the docs. But I would appreciate folks looking at it and trying to break it. Where the tests are inadequate, please suggest/add new tests so that this can be rationalized in the future without breaking anything or regressing.

drsteve:
Just pushed a few more tests, most of which did not pass initially. Followed by a few fixes. Some of the issues:
- I had some in-place multiplication and division to handle conversion Re<->km where necessary, and these could modify the values in the parent `Coords` object on calling `convert`. This is now resolved.
- Converting single points to/from geodetic coordinates worked fine, even handling singularities correctly. Multiple points for a single time were not handled correctly (Nx3 vs 3xN problem which didn't arise for 1D arrays). This is now resolved.

There's still an outstanding issue where a conversion from Re to km is not being done when it should be. I'm tracking it down, but that's why the tests are failing.

drsteve:
Note that, as of the latest commit, the test case for #577 will function correctly using the new backend:
```
import spacepy.time as spt
import spacepy.coordinates as spc

tt = spt.Ticktock(['2012-06-13T15:00:00', '2012-06-13T15:00:00', '2012-06-13T15:00:00'])
spc.Coords([[1.063871355909377, -90.0, 360.0], [1.063871355909377, -90.0, 0.0], [1.063871355909377, 90.0, 180.0]], 'GEO', 'sph', ticks=tt, use_irbem=False).convert('GDZ', 'sph')
```
and will return
```
Coords( [[428.7649441205885, -90.0, -1.4033418597069752e-14], [428.7649441205885, -90.0, 0.0], [428.7649441205885, 90.0, 180.0]] , 'GDZ', 'sph')
```

drsteve:
Got around to some benchmarking I can actually share, so that's added in the last commit (`developer/benchmarks/coordinates/convert.py`).

Outputs on local WSL2 are given below. The test sets up 20k positions and three time arrays to go with them.
Time array 1 is where all times are the same (so should only calculate transform matrices once).
Array 2 has times separated by 1 second (so should calculate a transform for every 30 seconds and use the nearest to transform each position).
Array 3 has times separated by 4 minutes (each position/time goes through the full setup and transform).

Then tests are done that set up a `Coords` object and do a coordinate conversion. The tests are split out into non-magnetic, magnetic and then geodetic. Non-magnetic doesn't need to evaluate IGRF to get the dipole axis (and then set up GSM, MAG, etc.). Geodetic doesn't actually need to calculate any transforms here because GEO to geodetic is time-independent, but I currently don't have a switch to check whether I need to initialize the transformations if it's _just_ between geographic and geodetic.

Results are marked as `IRB` if they use the IRBEM backend, `SPA` if they use the new SpacePy backend.

```
Each test uses 20000 positions to transform
SAME-TIME GEO->GSE (IRB): 0:00:00.243141
SAME-TIME GEO->GSE (SPA): 0:00:00.006097
SAME-TIME GEO->MAG (IRB): 0:00:00.153613
SAME-TIME GEO->MAG (SPA): 0:00:00.119078
CLOSE GEO->GSE (IRB): 0:00:00.183319
CLOSE GEO->GSE (SPA): 0:00:01.546313
CLOSE GEO->GSM (IRB): 0:00:00.046622
CLOSE GEO->GSM (SPA): 0:00:02.114280
SPACED GEO->GSE (IRB): 0:00:00.045815
SPACED GEO->GSE (SPA): 0:00:01.549922
SPACED GEO->GSM (IRB): 0:00:00.046296
SPACED GEO->GSM (SPA): 0:00:02.102209
SPACED GEO->GDZ (IRB): 0:00:00.034138
SPACED GEO->GDZ (SPA): 0:00:03.151237
```

- For large position arrays all at the same time the new backend is super-efficient and shakes out slightly faster than dropping to the Fortran.
- It doesn't look like the close-spacing vs wide-spacing is different here, but I was seeing a very clear difference in some other testing.
- Doing the extra calculations to get magnetic coordinates is slower, but not outrageous. I did try initializing the IGRF at noon every day and caching the result, but my attempt at that gave no noticeable speed-up.
- The geographic to geodetic calculation is slow. There are probably some good gains to be had by taking a closer look at this (and if necessary moving some to `libspacepy`).
- It wouldn't be too bad to do the same chunking and multiprocessing that we use for Lstar via irbempy, and that'd be a "pure python" way of speeding up conversions on large data sets.
- On the whole, I think that 20k positions is a decent "everyday usage" test though, and I don't think any of these speeds are a death knell. Especially as IRBEM is still intended to be the default backend in the next release. The question is whether we want to start the deprecation cycle (for IRBEM as default) on the next release, or just drop this in now and add a deprecation warning sometime after the next release when we hit whatever criteria we want to make that happen).

jtniehof:
I think if we're planning to do a transition, we should start the deprecation cycle immediately, and not wait until we know that the new backend is better for all situations.

Does it make more sense for `use_irbem` to be something specified on every initialization, or would it make more sense as a config file option?

If every call: I would think the default `use_irbem` kwarg value should be `None`, and the warning about using a default is only issued if it is `None` (and then it is set to `True` for the rest of the call). Then someone who really wants to stick with irbem can use `True` and not have a warning, and someone wanting to be explicit can use `False` and not have a warning. Then when we want to change the default *behaviour*, the default *value* can remain `None` and a similar warning is raised, but then it's set to `False` for the rest of the call. Implicit: default behaviour + warning; explicit: specified behaviour with no warning.

If in config file: I could see use the config value if default not specified (in which case, again `None` as the default kwarg, meaning "use config".) We could probably get fancy and have the config default be `None` as well (warn), so if there's a config specified it's used without warning, but otherwise there's a fall-through...that's getting so fancy it would need a truth table though, so not going to do that unless we really do think a config option is a good idea.

drsteve:
> I think if we're planning to do a transition, we should start the deprecation cycle immediately, and not wait until we know that the new backend is better for all situations.

Sure. I think I was getting too far ahead of myself and wondering when and how we'd pull the IRBEM-dependent code out to be an optional install.

> Does it make more sense for `use_irbem` to be something specified on every initialization, or would it make more sense as a config file option?

Setting the default backend in config seems like a good idea, without any further thought towards _how_ that gets handled. But should things like preference of Earth radius definition also be in config, if we're opening that can of worms?

> If every call: I would think the default `use_irbem` kwarg value should be `None`, and the warning about using a default is only issued if it is `None` (and then it is set to `True` for the rest of the call). Then someone who really wants to stick with irbem can use `True` and not have a warning, and someone wanting to be explicit can use `False` and not have a warning. Then when we want to change the default _behaviour_, the default _value_ can remain `None` and a similar warning is raised, but then it's set to `False` for the rest of the call. Implicit: default behaviour + warning; explicit: specified behaviour with no warning.

That also seems reasonable to me. 
 
> If in config file: I could see use the config value if default not specified (in which case, again `None` as the default kwarg, meaning "use config".)

This tracks with the _per call_, and seems reasonable.

> We could probably get fancy and have the config default be `None` as well (warn), so if there's a config specified it's used without warning, but otherwise there's a fall-through...that's getting so fancy it would need a truth table though, so not going to do that unless we really do think a config option is a good idea.

I can't decide whether this is easy (for both implementing and for users to understand) or whether it would be too much of a pain for users to figure out what they'd be getting.

jtniehof:
Ohboy, I didn't even think about the earth radius thing. I can see a couple of ways:

- Go ahead and make several config file options. I don't think this is the end of the world. (Option: add another section for coordinates specifically).
- Always use kwargs to get non-default behaviour
- Make subclasses for different behaviour and/or have class members that can be overriden (so for instance `Re` can be a class member and you can subclass it with a different value).
- Have some sort of call to set defaults so that Coords objects made after that call take on those defaults.

drsteve:
Just edited numbers into your last comment so it's easier to be clear when replying...
1. Config is a blessing and a curse for Earth radius, because it gets used in a bunch of places and for the new backend I explicitly define an ellipsoid class that has the relevant parameters. For geodetic everything gets converted to km first, so I _probably_ wouldn't get in trouble by using a configurable Earth radius in `Coords`.
2. Sure. Some of this is honestly pretty hard to propagate through though. Earth radius is perhaps relatively straightforward. If someone wants to use a different ellipsoid for geodetic? I don't know. At that point I think perhaps they should use the underlying conversions directly and not bother with `Coords`.
3. Thanks, I hate it. I mean, yes, it would work, but I just don't want to.
4. This is my current compromise preference. It can be used globally by updating module-level settings (as long as I don't end up in circular dependency hell), and can be trivially used to set different ellipsoids, etc.
5. I think I shot myself in the foot by even presenting options here. I probably should've just left everything as it is and kept my mouth shut.

drsteve:
Okay, I've added module-level defaults. That let me clean up some warnings-related code, and I also added some docstrings that'd previously been missing. The defaults are now propagated through to the conversion calls in `ctrans` so that the temporal grouping for efficiency is exposed through the defaults object, as well as the default ellipsoid (Which, right now, is the only defined ellipsoid, although a user could define one themselves.)

drsteve:
Current test coverage of relevant modules:
```
Name                                                                  Stmts   Miss  Cover
-----------------------------------------------------------------------------------------
/home/smorley/github/spacepy/build/lib/spacepy/coordinates.py           376     37    90%
/home/smorley/github/spacepy/build/lib/spacepy/ctrans/__init__.py       489     13    97%
/home/smorley/github/spacepy/build/lib/spacepy/ctrans/iau80n.py          33      0   100%
/home/smorley/github/spacepy/build/lib/spacepy/igrf.py                   89      4    96%
/home/smorley/github/spacepy/build/lib/spacepy/irbempy/__init__.py        2      0   100%
/home/smorley/github/spacepy/build/lib/spacepy/irbempy/irbempy.py       848    419    51%
```

drsteve:
Quick note for speed testing (just in case): to change the "close enough" time interval to re-use the calculated transforms in the benchmarking script (`developer/benchmarks/coordinates/convert.py`), add
```
spc.DEFAULTS.set_values(itol=10)
```
at the top. Default is 30s, line above sets 10s. Anything above 1s should buy a speed-up for the benchmark script.

jtniehof (dismissed):
I think we probably will want to do some scrubbing for style and the like, but it's long past time to merge this and treat that as a separate task. The testing is phenomenal on here, which is what matters the most.

We should open up an issue to implement updates for the IGRF coordinates in toolbox. Anything else that might need similar update?

jtniehof (approved):
It will come as no surprise that I still approve.

PR #298, opened 2020-05-04
ImfInput now rounds sub-millisecond times on write.

### LINKED ISSUE #291: spacepy.pybats.ImfInput.write can produce invalid time output
The write method of spacepy.pybats.ImfInput rounds the microseconds when writing the time, as it writes milliseconds to the IMF file. If the microseconds are 999500 or greater, this rounds to 1000, which is treated (rightly) as invalid on read.

Suggestions:
- Rather than rounding, we could just truncate to milliseconds.
- It might be more accurate to convert to a numeric time, round, then back-convert and write out to the text file.

```
from spacepy import pybats
imf = pybats.ImfInput()
for key in imf: imf[key]=[0]
imf['time'] = [dt.datetime(2017,9,6,16,42,36,999500)]
imf.write('testme.dat')
imf2 = pybats.ImfInput('testme.dat')
```

### Error message/Traceback:
```
ValueError                                Traceback (most recent call last)
<ipython-input-23-59a42331f5c9> in <module>
----> 1 imf2 = pybats.ImfInput('testme.dat')

~/projects/env/lib/python3.6/site-packages/spacepy/pybats/__init__.py in __init__(self, filename, load, npoin
ts, *args, **kwargs)
   1323         # Load/create data vectors.
   1324         if filename and load:  # Load contents from existing file.
-> 1325             self.read(filename)
   1326         else:
   1327             units   = ['nT', 'nT', 'nT', 'km/s', 'km/s', 'km/s', 'cm^-3', 'K']

~/projects/env/lib/python3.6/site-packages/spacepy/pybats/__init__.py in read(self, infile)
   1508                     int(parts[4]), #min
   1509                     int(parts[5]), #sec
-> 1510                     int(parts[6]) * 1000 #micro seconds
   1511                     ))
   1512             for j, name in enumerate(self.attrs['var']):

ValueError: microsecond must be in 0..999999

```
The issue is that the data line has 1000 for the milliseconds, as can also be seen through pdb:
```
ipdb> parts
['2017', '09', '06', '16', '42', '36', '1000', '0.00', '0.00', '0.00', '0.00', '0.00', '0.00', '0.00', '0.00'
]
```

```
Linux-4.4.0-18362-Microsoft-x86_64-with-Ubuntu-18.04-bionic
sys.version_info(major=3, minor=6, micro=7, releaselevel='final', serial=0)
numpy=1.16.2
scipy=1.2.1
matplotlib=3.0.3
```

### Version of SpacePy
From git repo, last commit was March 19th.

spacecataz:
Closes #291 

Repeat of previous PR that had conflicting merges.

Changes writing ImfInput objects to file: sub-millisecond values now use floor when converting micro- to millisecond values instead of round.
A new test has been added to make sure that this works using the example given in the issue.

I want to note @jtniehof's suggestion:

```
t = self.time[i]
ms = int(round(t.microsecond / 1000.))
if ms = 10000:
    ms = 0
    t += dt.timedelta(microseconds=1000)
out.write('{:%Y %m %d %H %M %S} {:03d} '.format(t, ms))
```

This is an alternative worth discussing, but I am hesitant to nest an `if` statement in a sometimes very long `for` loop.  IMF files can contain upwards of month(s) worth of data at sub-minute resolution, so writing can already be slow.  Worth discussing.

PR Checklist
[ n/a] New code has inline comments where necessary
[ n/a] Any new modules, functions or classes have docstrings consistent with SpacePy style
[ n/a] Added an entry to CHANGELOG if fixing a major bug or providing a major new feature
[ x] New features and bug fixes should have unit tests

### DIFF
--- a/tests/test_pybats.py
+++ b/tests/test_pybats.py
@@ -329,12 +329,32 @@ class TestImfInput(unittest.TestCase):
     knownImfRho  = [5., 15.]
     knownImfTemp = [.80E+05, 1.20E+05]
     knownImfIono = [4.99, 0.01]
+    knownSubMilli= dt.datetime(2017, 9, 6, 16, 42, 36, 999000)
 
     def tearDown(self):
         # Remove temporary files.
         for f in glob.glob('*.tmp'):
             os.remove(f)
 
+    def testSubMillisec(self):
+        '''
+        Test case where sub-millisecond time values can create problems on
+        read/write.
+        '''
+        # Create an IMF object from scratch, fill with zeros.
+        imf = pb.ImfInput()
+        for key in imf: imf[key]=[0]
+        
+        # Add a sub-millisecond time:
+        imf['time'] = [dt.datetime(2017,9,6,16,42,36,999500)]
+
+        # Write and test for non-failure on re-read:
+        imf.write('testme.tmp')
+        imf2 = pb.ImfInput('testme.tmp')
+
+        # Test for floor of sub-millisecond times:
+        self.assertEqual(self.knownSubMilli, imf2['time'][0])
+            
     def testWrite(self):
         # Test that files are correctly written to file.
         
--- a/tests/test_pybats.py
+++ b/tests/test_pybats.py
@@ -329,7 +329,7 @@ class TestImfInput(unittest.TestCase):
     knownImfRho  = [5., 15.]
     knownImfTemp = [.80E+05, 1.20E+05]
     knownImfIono = [4.99, 0.01]
-    knownSubMilli= dt.datetime(2017, 9, 6, 16, 42, 36, 999000)
+    knownSubMilli= dt.datetime(2017, 9, 6, 16, 42, 37, 0)
 
     def tearDown(self):
         # Remove temporary files.
@@ -346,7 +346,7 @@ class TestImfInput(unittest.TestCase):
         for key in imf: imf[key]=[0]
         
         # Add a sub-millisecond time:
-        imf['time'] = [dt.datetime(2017,9,6,16,42,36,999500)]
+        imf['time'] = [dt.datetime(2017,9,6,16,42,36,999600)]
 
         # Write and test for non-failure on re-read:
         imf.write('testme.tmp')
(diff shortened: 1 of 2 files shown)


tests/test_pybats.py
```diff
@@ -329,12 +329,32 @@ def setUp(self):
...
+        # Create an IMF object from scratch, fill with zeros.
+        imf = pb.ImfInput()
+        for key in imf: imf[key]=[0]
+        
+        # Add a sub-millisecond time:
+        imf['time'] = [dt.datetime(2017,9,6,16,42,36,999500)]
+
+        # Write and test for non-failure on re-read:
+        imf.write('testme.tmp')
+        imf2 = pb.ImfInput('testme.tmp')
```
drsteve, line 353: This doesn't get cleaned up... can we use a temporary file or have some code to remove the file after the test? (I assume that if, for some reason, the write fails then there's nothing to clean up)

tests/test_pybats.py
```diff
@@ -329,12 +329,32 @@ def setUp(self):
     knownImfRho  = [5., 15.]
     knownImfTemp = [.80E+05, 1.20E+05]
     knownImfIono = [4.99, 0.01]
+    knownSubMilli= dt.datetime(2017, 9, 6, 16, 42, 36, 999000)
 
     def tearDown(self):
```
spacecataz, line 334: Can confirm that tearDown exists and works properly on my machine.  @drsteve, can you confirm that tearDown is working for you?

jtniehof:
You're doing a lot more expensive stuff in that loop than an if. Unless
there's some evidence that's a substantial contributor, I think it makes
more sense to round. I can do a dumb test but if you've got a sample big
file that would be better.

drsteve:
I can provide a chunky input file... Let me just find it and I can send a link.

spacecataz:
I used the file provided by @drsteve, got average write times over 50 writes, and compared with the extra rounding work and without.  Minor impact, so I've added that change.  Have also updated the test.  I think we're good to go here.

PR #318, opened 2020-05-15
time.Ticktock: Clarify UTC, Epoch, UNIX time systems (Closes #316)

### LINKED ISSUE #316: Ticktock time system documentation
The ticktock docs need a few clarifications

- The "CDF" time is "old CDF Epoch", which is integral milliseconds (as a float....) since an epoch, but like UNIX time it cannot represent leap seconds (i.e. they are skipped).
- The "UTC" is handled as Python datetimes and so can't really represent leap seconds

jtniehof:
Just changes to the Ticktock docstring
 * Note that UTC is datetime and thus can't do leap seconds
 * Note that Epoch is CDF_EPOCH not CDF_TIME_TT2000
 * Remove note on UNIX time changing length of second; this is a Google
   interpretation not definition of UNIX time
 * Reformat list of time systems as definition list

- [ ] (N/A) New code has inline comments where necessary
- [ ] (N/A) Any new modules, functions or classes have docstrings consistent with SpacePy style
- [ ] (N/A) Added an entry to CHANGELOG if fixing a major bug or providing a major new feature
- [ ] (N/A) New features and bug fixes should have unit tests

### DIFF
--- a/spacepy/time.py
+++ b/spacepy/time.py
@@ -131,19 +131,31 @@ class Ticktock(MutableSequence):
     (TAI, UTC, ISO, JD, MJD, UNX, RDT, CDF, DOY, eDOY)
 
     Possible input data types:
-    ISO: ISO standard format like '2002-02-25T12:20:30'
-    UTC: datetime object with UTC time
-    TAI: elapsed seconds since 1958/1/1 (includes leap seconds)
-    UNX: elapsed seconds since 1970/1/1 (all days have 86400 secs sometimes unequal lenghts)
-    JD: Julian days elapsed
-    MJD: Modified Julian days
-    RDT: Rata Die days elapsed since 1/1/1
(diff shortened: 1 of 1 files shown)


spacepy/time.py
```diff
@@ -131,19 +131,31 @@ class Ticktock(MutableSequence):
...
-    All those listed above, plus
-    DOY:  Integer day of year, starts with day 1
-    eDOY: Fractional day of year, starts at day 0
+
+    ISO
+        ISO standard format like '2002-02-25T12:20:30'
+    UTC
+        datetime object with UTC time
+    TAI
+        elapsed seconds since 1958/1/1 (includes leap seconds)
```
drsteve, line 140: Elapsed should be capitalized for consistency with the other entries

jtniehof: Good point. Force-pushed updated commit (seemed not worth having another commit in history just for this.)

PR #324, opened 2020-05-15
datamodel.py: h5py is required already, treat it that way. Closes #323

### LINKED ISSUE #323: h5py is a requirement or optional?
h5py is called out in requirements.txt and setup.py as required. 

Yet in datamodel.py there is 
```
    try:
        import h5py as h5py
    except ImportError:
        raise ImportError('h5py is required to use HDF5 files')
```
in a few places. 

Thoughts on ripping out the above from datamodel.py?

Looking at spacepy 0fbb879

balarsen:
- [X ] Pull request has descriptive title
- [ X] Pull request gives overview of changes
- [ N/A] New code has inline comments where necessary
- [ N/A] Any new modules, functions or classes have docstrings consistent with SpacePy style
- [ X] Added an entry to CHANGELOG if fixing a major bug or providing a major new feature
- [ N/A] New features and bug fixes should have unit tests
- [ X] Relevant issues are linked to (e.g. `See issue #` or `Closes #`)

Minor cutting code out of datamodel.py where h5py was treated as optional. It is required already in setup.py and requirements.txt. 

Updated changelog. 

Closes #323

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,8 @@ Changes in Version 0.2.2 (2020-xx-xx)
 - Document end of support for Python 2
 coordinates
  - Include upstream fix for SPH->RLL coordinate transform
+datamodel
+ - The h5py requirement is utilized rather than a check
 irbempy
  - Update IRBEMlib to rev620
  - Rev620 adds IGRF13 coefficients
--- a/spacepy/datamodel.py
+++ b/spacepy/datamodel.py
@@ -174,6 +174,7 @@ try:
 except ImportError:
     import io as StringIO
 
+import h5py
 import numpy
 from . import toolbox
 from . import time as spt
@@ -995,13 +996,8 @@ def fromHDF5(fname, **kwargs):
                                     'key = {0} ({1})\n'.format(key, type(key)) +
                                     'value = {0} ({1})'.format(value, type(value)), DMWarning)
 
-    try:
-        import h5py as hdf
-    except ImportError:
-        raise ImportError('HDF5 converter requires h5py')
-
     if type(fname) in str_classes:
-        hfile = hdf.File(fname, mode='r')
+        hfile = h5py.File(fname, mode='r')
     else:
         hfile = fname
         #should test here for HDF file object
@@ -1012,7 +1008,7 @@ def fromHDF5(fname, **kwargs):
         path = kwargs['path']
 
     SDobject = SpaceData()
-    allowed_elems = [hdf.Group, hdf.Dataset]
+    allowed_elems = [h5py.Group, h5py.Dataset]
     ##carry over the attributes
     hdfcarryattrs(SDobject, hfile, path)
     ##carry over the groups and datasets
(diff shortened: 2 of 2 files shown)


jtniehof:
h5py is a [soft dependency](https://spacepy.github.io/dependencies.html). SpacePy will (and should) work without it. The listing in setup.py and in requirements.txt is so that people installing via e.g. pip will get h5py and not have to install it by hand.

This is true of just about everything--the installer will work just fine without them (except Python and numpy), but we declare dependencies so people get full functionality.

balarsen:
> This is true of just about everything--the installer will work just fine without them (except Python and numpy), but we declare dependencies so people get full functionality.

But we don't... missing sphinx and numpydoc from the list for full dependency.

balarsen:
>  SpacePy will (and should) work without it

And it does not...
FAILED (errors=37, expected failures=3)

test_empiricals.py, test_datamodel.py, datamodel.py all fail.

jtniehof:
Sphinx and numpydoc are not required to run any porttion of SpacePy.

Yes, tests will fail if you don't have the dependency installed that supports the tested functionality.

balarsen:
the failing tests mean we have no way to verify that soft dependencies are
actually soft.

I think we should reevaluate how we do this. Either add tests and CI in to
make sure or move to the spacepy modules we have tossed around.

Feels really awkward.

jtniehof:
I can see the argument that we want to make sure the soft dependencies really are only affecting what we want. That would probably involve skip-decorating a lot of tests, and then the CI would have to have a LOT of possible combinations of dependencies.

What you're also finding is that for some cases we might need something to *test* a module but not to *run* it, e.g. omni. There's also a documentation issue where empiricals requires OMNI (thus hdf5) and that isn't documented. Actually since empiricals can often get the inputs elsewhere, maybe it shouldn't hard-require OMNI.

The other thing we can do, instead of skip-decorating the crap out of the tests, is to just define a few sub-test scripts and a few CI configurations that skip the "big ones." So maybe something without HDF5, something without Fortran, something without CDF. Because if we just skip-decorate everything I'd be afraid of missing, say, CI failing to install a dependency or something in the situation where we are trying to test it all. If we want to test "reduced dependency environments," sounds like a candidate for a separate issue. I don't think blowing up if people don't have h5py is necessarily getting us to a better place.

Yes, there's the modules stuff and all, but it's a lot of work and honestly not my highest priority. It wouldn't dodge the issue of testing.

jtniehof:
Objections to closing this and #323 assuming we proceed with reasonable soft dependency testing as in #333? @drsteve ?

balarsen:
Fine by me.

drsteve:
Works for me.
To handle running different commands in different jobs under Travis, I think we'll need to move the `scripts` section up under each job... or set the test script name as a variable in each job definition and then call via the variable in the `script` section.

I'm closing, but feel free to reopen if we need to. I imagine that further discussion can happen elsewhere.

Issue #198, opened 2019-08-16
pycdf: make v3 CDFs and TT2000 the default


jtniehof:
Right now pycdf sets backward compatible mode by default (the CDF library does _not_). At this point there should be nobody using version 2 of the library so the default should be not backward compatible.

I think this is going to require three releases: one to issue a warning that the default will change in the future (if set_backward hasn't been called yet one way or the other), one to similarly warn that the default has changed, and finally one to remove that warning.

jtniehof:
"about to change" warning should be on next release.

jtniehof:
I'm also thinking in terms of "hiding" the backward-compatible setting and just adding a kwarg to the CDF constructor if they want to make a backward-compatible one. There's not really much reason to go toggling it in the library. That's probably a secondary consideration though.

jtniehof:
Also, thoughts on making this a DeprecationWarning vs. some other warning type? FutureWarning used to be the warning for "hey this is going to change" vs. being deprecated; now it's meant for end user vs. developer.

balarsen:
I have no particular preference here, if I read the [CDF changelog](https://cdf.gsfc.nasa.gov/html/cdf_changesnote.html) correction v3 came out in 2005. Based on that I vote DeprecationWarning

jtniehof:
Warning of upcoming change: 0.2.2
Change behavior and warn that it's changed: 0.3.0
Remove warning: 0.4.0

jtniehof:
It's probably also worth changing the default time type to TT2000 as part of this transition....might as well get it all done at once.

PR #335, opened 2020-05-27
Warn if create CDF without making explicit backward-compat or not.

jtniehof:
This is the first step towards #198, warning the user that the default for backward-compatible CDFs is about to change. Basically if the user hasn't explicitly set backward-compatible one way or the other, this will warn when making a new CDF (if made from a master, the master's backward-compatible carries over.)

The real nastiness is in the test suite, because there's a bug in Python 2.7 that requires some work-arounds to make sure the deprecation warning is really raised. And then the test suite needs to quiet all the places where the tests are assuming the default--thus a second commit to separate that out.

- [ ] (N/A) Any new modules, functions or classes have docstrings consistent with SpacePy style

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -24,6 +24,8 @@ pycdf
  - istp.VariableChecks and FileChecks add check for empty attribute Entries
  - istp.VarBundle class to copy variables, slices, and depends between CDFs
  - istp.nanfill function to replace fill values with NaN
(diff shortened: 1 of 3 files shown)


balarsen:
@jtniehof , I;m good with merging this if you are.

jtniehof:
I'm good to merge. Thanks.

PR #352, opened 2020-06-09
irbempy: add support to the landi2lstar IRBEM routine to compute Lstar.

AntoineBrunet:
## PR Checklist

This PR adds support for the landi2lstar IRBEM routine to compute Lstar values. It allows for fast Lstar computation when the magnetic fields are compatible. It is added as a boolean flag to the `get_Lstar` function, which defaults as False for compatibility reasons.

Also, the subroutine call code in the _get_Lstar function has been corrected to call appropriate routines (make_Lstar, landi2Lstar, and their shell_splitting counterparts). The shell_splitting versions are called if more than one alpha angle is provided, or the provided angle is not 90 degrees.

This PR requires to rebuild the IRBEM library, as it adds wrappers to new subroutines.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -512,15 +512,17 @@ class build(_build):
         olddir = os.getcwd()
         os.chdir(builddir)
         F90files = ['source/onera_desp_lib.f', 'source/CoordTrans.f', 'source/AE8_AP8.f', 'source/find_foot.f',\
-                    'source/drift_bounce_orbit.f']
+                    'source/LAndI2Lstar.f', 'source/drift_bounce_orbit.f']
         functions = ['make_lstar1', 'make_lstar_shell_splitting1', 'find_foot_point1',\
                      'coord_trans1','find_magequator1', 'find_mirror_point1',
                      'get_field1', 'get_ae8_ap8_flux', 'fly_in_nasa_aeap1',
-                     'trace_field_line2_1', 'trace_field_line_towards_earth1', 'trace_drift_bounce_orbit']
+                     'trace_field_line2_1', 'trace_field_line_towards_earth1', 'trace_drift_bounce_orbit',
+                     'landi2lstar1', 'landi2lstar_shell_splitting1']
 
         # call f2py
         cmd = self.f2py + ['--overwrite-signature', '-m', 'irbempylib', '-h',
                'irbempylib.pyf'] + F90files + ['only:'] + functions + [':']
+        print(cmd)
         subprocess.check_call(cmd)
         # intent(out) substitute list
         outlist = ['lm', 'lstar', 'blocal', 'bmin', 'xj', 'mlt', 'xout', 'bmin', 'posit', \
--- a/setup.py
+++ b/setup.py
@@ -522,7 +522,6 @@ class build(_build):
         # call f2py
         cmd = self.f2py + ['--overwrite-signature', '-m', 'irbempylib', '-h',
                'irbempylib.pyf'] + F90files + ['only:'] + functions + [':']
-        print(cmd)
         subprocess.check_call(cmd)
         # intent(out) substitute list
         outlist = ['lm', 'lstar', 'blocal', 'bmin', 'xj', 'mlt', 'xout', 'bmin', 'posit', \
--- a/spacepy/irbempy/irbempy.py
+++ b/spacepy/irbempy/irbempy.py
@@ -1322,7 +1322,7 @@ def get_AEP8(energy, loci, model='min', fluxtype='diff', particles='e'):
     
 
 # -----------------------------------------------
-def _get_Lstar(ticks, loci, alpha, extMag='T01STORM', options=[1,0,0,0,0], omnivals=None): 
+def _get_Lstar(ticks, loci, alpha, extMag='T01STORM', options=[1,0,0,0,0], omnivals=None, landi2lstar=False): 
     """
     This will call make_lstar1 or make_lstar_shell_splitting_1 from the irbem library
     and will lookup omni values for given time if not provided (optional). If pitch angles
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -512,15 +512,17 @@ def compile_irbempy(self):
...
                      'coord_trans1','find_magequator1', 'find_mirror_point1',
                      'get_field1', 'get_ae8_ap8_flux', 'fly_in_nasa_aeap1',
-                     'trace_field_line2_1', 'trace_field_line_towards_earth1', 'trace_drift_bounce_orbit']
+                     'trace_field_line2_1', 'trace_field_line_towards_earth1', 'trace_drift_bounce_orbit',
+                     'landi2lstar1', 'landi2lstar_shell_splitting1']
 
         # call f2py
         cmd = self.f2py + ['--overwrite-signature', '-m', 'irbempylib', '-h',
                'irbempylib.pyf'] + F90files + ['only:'] + functions + [':']
+        print(cmd)
```
jtniehof, line 525: Was this for debugging and left in accidentally?

spacepy/irbempy/irbempy.py
```diff
@@ -1433,16 +1435,29 @@ def _get_Lstar(ticks, loci, alpha, extMag='T01STORM', options=[1,0,0,0,0], omniv
     nTAI = len(ticks)
     d = prep_irbem(ticks, loci, alpha, extMag, options, omnivals)
     nalpha = d['nalpha']
+    if d['kext'] != 5 or d['options'][4] != 0:
+        landi2lstar = False
+    no_shell_splitting = (nalpha == 0) or (nalpha == 1 and alpha[0] == 90)
         
-    if isinstance(alpha, numbers.Number): # no drift shell splitting
-        raise Exception
```
jtniehof, line 1438: This looks like this was a case SpacePy never actually handled before, is that accurate?

spacepy/irbempy/irbempy.py
```diff
@@ -1433,16 +1435,29 @@ def _get_Lstar(ticks, loci, alpha, extMag='T01STORM', options=[1,0,0,0,0], omniv
...
+        else:
+            lm, lstar, blocal, bmin, xj, mlt = oplib.make_lstar1(nTAI, d['kext'], d['options'], \
+                d['sysaxes'],d['iyearsat'], d['idoysat'], d['utsat'], d['xin1'], d['xin2'], \
+                d['xin3'], d['magin'])
 
     elif nalpha > 0 and nalpha <= d['nalp_max']: # with drift shell splitting
-        lm, lstar, bmirr, bmin, xj, mlt = oplib.make_lstar_shell_splitting1(nTAI, nalpha, \
-            d['kext'], d['options'], d['sysaxes'], d['iyearsat'], d['idoysat'], d['utsat'], \
-            d['xin1'], d['xin2'], d['xin3'], d['degalpha'], d['magin'])
+        if landi2lstar:
```
jtniehof, line 1453: I wonder if it might be clearer that the only difference between these is the function called by doing something like:
```
func = oplib.landi2lstar_shell_splitting1 if landi2lstar else oplib.make_lstar_shell_splitting
lm, lstar, bmirr, bmin, xj, mlt = func(all the args)
```
Something similar might collapse this whole if tree down...I can whip up something if that sounds reasonable. It's just easy to miss the differences when there's so much similar.

spacepy/irbempy/irbempy.py
```diff
@@ -1455,13 +1470,13 @@ def _get_Lstar(ticks, loci, alpha, extMag='T01STORM', options=[1,0,0,0,0], omniv
     mlt[np.where( tb.feq(mlt,d['badval'])) ] = np.NaN
 
     results = {}
-    if nalpha == 0:
-        results['Lm'] = lm[0:nTAI]
-        results['Lstar'] = lstar[0:nTAI]
+    if no_shell_splitting:
+        results['Lm'] = lm[0:nTAI][:,None]
```
jtniehof, line 1474: Why is the extra axis required on here now when it wasn't before? Is it because this code was never reachable before (with the `raise Exception`)?

setup.py
AntoineBrunet: Yes, sorry for this.

spacepy/irbempy/irbempy.py
AntoineBrunet: I think it was some debug code to ensure the _get_Lstar function will not be called with only one value.

spacepy/irbempy/irbempy.py
AntoineBrunet: Yes, I believe it would be clearer.

spacepy/irbempy/irbempy.py
AntoineBrunet: Yes, previously the code only called the *_shell_splitting version, which I believe returns arrays with different number of dimensions. The returned arrays are post-processed in the `reassemble` function, but we need to ensure that the routines have the same number of dimensions for both versions.

AntoineBrunet:
Thanks for considering this PR.

> * You mention "This PR requires to rebuild the IRBEM library, as it adds wrappers to new subroutines." We usually build the library on install anyhow, is there anything else that needs to be done? It looks like LAndI2Lstar is in rev620 that we have incorporated into SpacePy, so do we need to pull anything further from the irbem tree?

No, this routine has been included in IRBEM for a long time, so there is no need to update IRBEM.

> * There should be unit tests of the new functionality/code paths. I'm happy to help contribute these.
I can have a look at how you test the Lstar computation and propose some tests for this version. It should be pretty similar.

> * The unit tests we do have are now failing. Bmirr is no longer populated in the output of AlphaOfK and the L* value for T01 has changed very slightly--it's probably not wrong but it's clear we're going down a different code path. I can try to help puzzle that out but @drsteve is probably the best-equipped to comment on why certain functions were being called in certain circumstances and what the implications are of changing that.

I think the issue comes from the fact that previously Spacepy only called the _shell_splitting version of make_Lstar, which returns `Bmirr`, while the normal version returns `Blocal`. When alpha is 90, I believe they should have the same values. Since irbempy already post-process the results, it could make sense to rename one to the other.

AntoineBrunet:
I have commited a test suit for the new functions. Also I have corrected the output of the function to return `Bmirr` even if the regular (non shell_splitting) subroutines are called (by copying `Blocal`). I believe all tests pass (I had to change the expected results for T01 tests, because the routine call changed, but the difference is very small, as expected).

jtniehof:
Looks good. I'll submit a separate PR to tidy up the differ-only-by-function-call bits, so it can be evaluated separately from this (and not hold this up.) Since @drsteve hasn't commented I'm going to presume he's okay.

Thanks for this, and thanks for your patience with how long it took me to get to it.

PR #355, opened 2020-06-11
New function to rebin array by another array

jtniehof:
This PR adds a function that allows the rebinning of one dimension of an array by data in another array. The canonical example of this is going from an array organized by sensor number and tagged by pitch angle, to one binned by pitch angle.

This is the type of thing I've found myself doing a _lot_ and nailing it down in general is actually nontrivial.

I think the one thing lacking in the documentation is a full end-to-end example, but since this usually involves big arrays it's hard to come up with a concise example.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,8 @@ Changes in Version 0.2.2 (2020-xx-xx)
 - Document end of support for Python 2
 coordinates
  - Include upstream fix for SPH->RLL coordinate transform
+datamanager
+ - New function "rebin", rebin an axis of an array by contents of another
 irbempy
  - Update IRBEMlib to rev620
  - Rev620 adds IGRF13 coefficients
--- a/Doc/source/pyplots/datamanager_rebin_1.py
+++ b/Doc/source/pyplots/datamanager_rebin_1.py
@@ -1,0 +1,18 @@
+#!/usr/bin/env python
+
+from datamanager_rebin_common import *
+
+times = numpy.arange(0., 7200, 5)
+alpha = numpy.arange(0, 181., 2)
+# Add a dimension so the flux is a 2D array
+flux = j(1., numpy.expand_dims(times, 1),
+             numpy.expand_dims(alpha, 0))
+# The ordering of pcolormesh inputs is (y, x), so transpose
+matplotlib.pyplot.pcolormesh(
+    spacepy.toolbox.bin_center_to_edges(times),
+    spacepy.toolbox.bin_center_to_edges(alpha),
+    flux.transpose(), norm=matplotlib.colors.LogNorm())
+matplotlib.pyplot.ylabel('Pitch angle (deg)')
+matplotlib.pyplot.xlabel('Time (sec)')
+matplotlib.pyplot.title('Flux at 1 MeV')
+matplotlib.pyplot.show()
--- a/Doc/source/pyplots/datamanager_rebin_1.py
+++ b/Doc/source/pyplots/datamanager_rebin_1.py
@@ -7,11 +7,8 @@ alpha = numpy.arange(0, 181., 2)
 # Add a dimension so the flux is a 2D array
 flux = j(1., numpy.expand_dims(times, 1),
              numpy.expand_dims(alpha, 0))
-# The ordering of pcolormesh inputs is (y, x), so transpose
-matplotlib.pyplot.pcolormesh(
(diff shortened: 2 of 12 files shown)


jtniehof:
If you look at the code, you'll note I use `numpy.digitize`. It's only the first step in a process.

balarsen:
Ok, fair enough. 

I do this all the time but using pandas. 
```

cuts = pd.cut(data, [1,4,7,10])
df.groupby(df['data'].loc[cuts]).mean(). # or .std()
```

jtniehof:
Dataframes aren't N-D, last I checked. It says "Two-dimensional" at the top of the documentation.

If you can tell me how in Pandas I can take an array of flux that's dimensioned Nt times by Nl look directions by Ne energies and convert it to Nt times by Np pitch angles by Ne energies via an array of pitch angles dimensioned Nt by Np by Ne, I'm all ears.

This is not rebinning in the sense of scaling up or down.

balarsen:
that package is called `xarray` does all that. I find it a PITA but it does do all that. At one point I made HOPE spectrograms from it using only xarray built ins.

jtniehof:
I'll see what I can do. As I said, it's hard to come up with an example that's self-contained and works but it's clear from this discussion the purpose isn't apparent. And then you can either tell me how to do it in xarray or note that it can't be done.

balarsen:
> I'll see what I can do. As I said, it's hard to come up with an example that's self-contained and works but it's clear from this discussion the purpose isn't apparent. And then you can either tell me how to do it in xarray or note that it can't be done.

Sounds good. I am not really saying that xarray should be a dependency (maybe a bit too obscure unless we need to do a lot of this stuff) but overall the conversation is good to have.

balarsen:
>     This is clearest with an example. Consider a flux as a function of time,
>     energy, and the look direction of a detector (could be multiple detectors,
>     or spin sectors.) The flux is then 3-D, dimensioned Nt x Ne x Nl. Now
>     consider that each look direction has an associated pitch angle that is
>     a function of time and thus stored in an array Nt x Nl. Then this function
>     will redimension the flux into pitch angle bins (rather than tags.)

One place would be what are the output expect dimensions of the above? Maybe adding in concrete numbers like assume there are 3400 times, 16, sectors, and 3 detectors.

jtniehof:
Read the paragraph after your quote and it'll give you the output dimensions.

The short is: this takes one dimension of an array and replaces it with a different dimension that can be determined from the first.

jtniehof:
I've pushed a pretty extensive example in the docstring. You're definitely going to need to pull, install, and build the docs to follow it.

jtniehof:
The build passed; there's something screwed up with Travis where it's not reporting out as successful. Since they're actually the same build I can't restart it. Travis also seems to keep logging me out every time I blink, so I think it's messed up.

balarsen:
That is an awesome example! Let me study it a bit more.

balarsen:
@jtniehof , thanks for this great example. I am happy with this. Let me know what you are thinking with the travis #357 , can merge this when you are happy.

jtniehof:
This should be independent of #357 ...I got to playing with that because of the double-check thing here, but they're actually unrelated.

jtniehof:
So good to merge when you are.

jtniehof:
Incidentally, I did find what happens with the double-checks...apparently there's one for the PR itself and one for the branch underlying it, they do go to a single build but sometimes it isn't properly reported out on both.

jtniehof:
Actually, maybe hold off a bit...I want to look if I can use plot.spectrogram.simpleSpectrogram to make the example a little shorter (and more SpacePy-y).

jtniehof:
This is _much_ nicer with `simpleSpectrogram`. The latest push, in addition to switching the example, makes two simpleSpectrogram changes:

1. It fixes the case where Y is already bin edges instead of bin centers. For X this was checked, but Y was assumed to be bin centers.
2. It fixes the case where no colorbar label is specified. This was being passed None which got cast to a string, 'None', labeling the colorbar with "None" which looks weird.

Sorry for mixing the stuff together but it didn't seem worth opening a PR just for those, particularly since this one depends on the first fix.

jtniehof:
Ready to merge once the checks pass (but do take a look at those fixes.)

jtniehof:
Poke for the merge, unless @drsteve wants to take a look at the simpleSpectrogram changes.

drsteve:
I can take a look tomorrow...

PR #356, opened 2020-06-11
time.Ticktock: allow non-str string-like input (Closes #354)

### LINKED ISSUE #354: Ticktock with isotime input fails on non-str input
My issue is that the result of HDF5 import into datamodel data type changed between 0.1.6 (which I use under Python2.7) and 0.2.1 (which I am trying to use under Python3).

```
from __future__ import print_function

import os
os.putenv("CDF_LIB", "/usr/local/cdf/lib")
import spacepy.datamodel as dm
import spacepy.time as spt

hd = os.path.expanduser('~')

fn = '{}/data/RBSP/A/MagEphem/2015/rbspa_def_MagEphem_T89Q_20151225_v1.0.0.h5'.format(hd)
sd = dm.fromHDF5(fn)
print ('type(sd[\'IsoTime\'] is {}'.format(type(sd['IsoTime'])))
print ('type(sd[\'IsoTime\'][0] is {}'.format(type(sd['IsoTime'][0])))
ep = spt.Ticktock(sd['IsoTime'])
```

### Error message/Traceback:

```
>>> type(sd['IsoTime'] is <class 'spacepy.datamodel.dmarray'>
type(sd['IsoTime'][0] is <class 'numpy.bytes_'>
  File "<stdin>", line 1, in <module>
  File "/home/andrei/lib/python/scripts/tt_test.py", line 17, in <module>
  File "/usr/local/lib/python3.8/dist-packages/spacepy/time.py", line 236, in __init__
    elif self.data[0] > 1e13:
TypeError: '>' not supported between instances of 'numpy.bytes_' and 'float'
 ...
```

### Version of SpacePy

The error is produced when running 0.2.1. When I am running 0.1.6 under Python 2.7, the result for type(sd['IsoTime'][0]) is <class numpy.string_>, and then it is easily converted to TickTock.

jtniehof:
Ticktock currently fails when the constructor is called with ISO timestrings in a string-like class that isn't actually str. One manifestation of this is in reading times from HDF5 files that were saved in Python 2 in Python 3. This PR fixes that. The Python 2 code is pretty separable from the Python 3 so hopefully it'll be easy to rip out when the time comes.

Note this still leaves the input data alone, not converting it to str. That's consistent with our usual approach in Ticktock but I can see the argument for converting it at construction time.

This seemed minor enough that I left it out of the CHANGELOG, but happy to hear other opinions.

- [X] (N/A?) Added an entry to CHANGELOG if fixing a major bug or providing a major new feature

### DIFF
--- a/spacepy/time.py
+++ b/spacepy/time.py
@@ -1144,23 +1144,19 @@ class Ticktock(MutableSequence):
                    else (d.encode('ascii') for d in self.data) if str is bytes \
                         else (d.decode('ascii') for d in self.data)
             # try a few special cases that are faster than dateutil.parser
-            try:
-                UTC = [datetime.datetime.strptime(isot, '%Y-%m-%dT%H:%M:%S') for isot in data]
-            except ValueError:
+            for strfmt in ('%Y-%m-%dT%H:%M:%S',
+                           '%Y-%m-%dT%H:%M:%SZ',
+                           '%Y-%m-%d',
+                           '%Y%m%d',
+                           '%Y%m%d %H:%M:%S'):
                 try:
-                    UTC = [datetime.datetime.strptime(isot, '%Y-%m-%dT%H:%M:%SZ') for isot in data]
+                    UTC = [datetime.datetime.strptime(isot, strfmt)
+                           for isot in data]
+                    break
                 except ValueError:
-                    try:
-                        UTC = [datetime.datetime.strptime(isot, '%Y-%m-%d') for isot in data]
-                    except ValueError:
-                        try:
-                            UTC = [datetime.datetime.strptime(isot, '%Y%m%d') for isot in data]
-                        except ValueError:
-                            try:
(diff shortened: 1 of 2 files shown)


jtniehof:
I also got cranky about the deeply-nested try-except in converting ISO to UTC and changed that in a commit I just pushed. It's not directly on topic to the bug but cleans things up a bit.

jtniehof:
Fails with the magephem because somehow winds up with self.data being length 1441 and self.TAI being length 1440. So something isn't adequately tested; might be this fix, might be something else. I will check and test + fix will go to this PR.

jtniehof:
Fixed with the magephem. It was a result of trying to be too clever and not make a new copy of the data, but then I just tried to iterate over it twice.

Just pushed another commit that does it with numpy, even better.

balarsen:
Good, stuff. I'm happy with this. Will be nice to drop all the py2 py3 stuff eventually. #26

PR #357, opened 2020-06-12
Update Travis CI

jtniehof:
Our builds were all running on xenial; now that focal is out it seemed time to move to bionic. We also weren't testing Python 3.8. My thought is we probably want to test "latest" numpy on most minor versions of Python, with tests for the special cases that caused trouble on the latest python. I couldn't find back the issues that caused us to test particular versions of numpy; it's possible we can drop some in the future.

Here's the Python/numpy version matrix. "Latest" numpy is anything >=1.6 excluding 1.15.0, since 1.15.0 had a bug we couldn't work around (in I think f2py).

- Python 2.7 (moved to bionic, no other changes)
  - numpy 1.15.4 (probably the latest Python 2 numpy)
  - numpy 1.13.3 (a problem child?)
  - numpy 1.11.3 (a problem child?) 
- Python 3.5 (stays on xenial, no bionic support in travis)
  - numpy latest
  - numpy 1.13.3 (problem child?)
- Python 3.6 (moved to bionic)
  - numpy latest
  - DELETED numpy 1.13.3
- Python 3.7 (moved to bionic)
  - numpy latest
  - DELETED numpy 1.14.5
 - Python 3.8 (all new, bionic)
   - numpy latest
   - numpy 1.14.5 (problem child?)

If I can do math right, this means we stay at 9 builds, add Python 3.8, and maintain our numpy version coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,12 +25,12 @@ matrix:
   - python: 3.7
     dist: bionic
     env: NUMPY_VERSION=">=1.6,!=1.15.0"
+  - python: 3.7
+    dist: bionic
+    env: NUMPY_VERSION="==1.14.5"
   - python: 3.8
     dist: bionic
     env: NUMPY_VERSION=">=1.6,!=1.15.0"
-  - python: 3.8
-    dist: bionic
-    env: NUMPY_VERSION="==1.14.5"
 
 before_install:
  - sudo apt-get update -qq
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,35 @@
 language: python
 
+sudo: yes
+
 matrix:
  include:
   - python: 2.7
+    dist: bionic
     env: NUMPY_VERSION="==1.15.4"
   - python: 2.7
+    dist: bionic
     env: NUMPY_VERSION="==1.13.3"
   - python: 2.7
+    dist: bionic
     env: NUMPY_VERSION="==1.11.3"
   - python: 3.5
+    dist: xenial
     env: NUMPY_VERSION=">=1.6,!=1.15.0"
   - python: 3.5
+    dist: xenial
     env: NUMPY_VERSION="==1.13.3"
   - python: 3.6
-    env: NUMPY_VERSION=">=1.6,!=1.15.0"
-  - python: 3.6
-    env: NUMPY_VERSION="==1.13.3"
-  - python: 3.7
-    dist: xenial
-    sudo: yes
+    dist: bionic
     env: NUMPY_VERSION=">=1.6,!=1.15.0"
   - python: 3.7
-    dist: xenial
+    dist: bionic
+    env: NUMPY_VERSION=">=1.6,!=1.15.0"
+  - python: 3.8
+    dist: bionic
+    env: NUMPY_VERSION=">=1.6,!=1.15.0"
+  - python: 3.8
+    dist: bionic
     env: NUMPY_VERSION="==1.14.5"
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,31 +6,34 @@ matrix:
  include:
   - python: 2.7
     dist: bionic
-    env: NUMPY_VERSION="==1.15.4"
+    env: NUMPY_VERSION=">=1.6.0,<1.7.0"
   - python: 2.7
     dist: bionic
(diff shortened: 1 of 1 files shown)


jtniehof:
numpy 1.16.6 supports Python 2, so maybe we should switch the Python 2 build to that. We were probably specifying 1.15 because we had Fortran build problems on 1.16 for awhile that we fixed on our end.

jtniehof:
It looks like the specific versions tested were not problem children as such, but the latest subminor version available of a few minor versions intended to get a range. I suggest we only do those tests on 2.7 and 3.8, and switch them from e.g. `==1.13.3` to `>=1.13.0,<1.14.0`

jtniehof:
@drsteve , am I evaluating the versions appropriately? Also, any idea how I can trigger a build on this PR? Even when it wasn't draft, it wasn't triggering.

drsteve:
Not sure about the lack of a build. Travis should be triggering on PR commits (just not on drafts).
I'll kick this back out of draft status to see whether it triggers.

Your thoughts on why the different subminor versions were being explicitly tested are about right. Some of them were, as I recall, subminor versions that worked after the initial releases of those minor versions didn't. E.g., changing `==1.13.3` to `>=1.13.0,<1.14.0` should work because it'll grab 1.13.3 anyway (unless, I suppose, there was another subminor release).

drsteve:
Still no build, I have two thoughts.
- Perhaps if you can rewrite your branch and force-push? That may provide a trigger?
- If that doesn't work, then I'd suggest finding another file you can touch (delete whitespace, something inconsequential) and adding that to your commit (then force pushing again)

If all else fails, maybe kill the PR and redo? If it still doesn't work then the fallback is probably merge (w/o testing) to a feature branch on the upstream and hopefully the branch will test. Then we can do a PR from the branch into master.

drsteve:
Scratch that ... I was digging further into Travis logs and it says it couldn't parse the run. So there's an error in the YAML.

Might be where you've moved the sudo setting outside the matrix.

jtniehof:
Thanks. I'll take a look at the format. The Travis documentation is one of those annoying things that seems to assume you mostly know what you're doing.

.travis.yml
```diff
@@ -1,27 +1,35 @@
...
-    sudo: yes
+    dist: bionic
     env: NUMPY_VERSION=">=1.6,!=1.15.0"
-  - python: 3.7
-    dist: xenial
+  - python: 3.8
+    dist: bionic
+    env: NUMPY_VERSION=">=1.6,!=1.15.0"
+  - python: 3.8
+    dist:bionic
```
drsteve, line 32: This could be the culprit... you're missing a space after `dist`

jtniehof:
This time I made sure the yaml could at least be read before pushing, and the build seems to be going, so that's good.

Can you point me to where in the logs you looked to find this? I need to continue to up my Travis game.

"subminor versions that worked after the initial releases of those minor versions didn't" applies, I think, only to 1.15.0. We thought we were facing something similar in 1.16 but it was a real issue that we fixed. The only one we're excluding in requirements.txt is 1.15.0, so I think for all the others we can go to just testing the subminor. I'll push another commit to shuffle that a bit once I know the current build works.

jtniehof:
Looks like will probably have to force the SciPy version on the NumPy 1.14.5 check.

drsteve:
travis_requests_menu
travis_requests_list

jtniehof:
Turns out there were two problems. SciPy 1.4.1 is installed in the 3.8 environment but it requires numpy 1.17.3 to run in 3.8, so 1.14 failed. Also numpy 1.14 only supports Python 3.6 anyhow (but "should work for the upcoming Python 3.7").

So I rolled the 1.14 tests back to Python 3.7.

jtniehof:
Latest commit tests the earliest supported minor version of numpy and the latest supported minor version of numpy, on every minor version of Python. The latest subminor is always tested. This means 1.18 gets tested five times...I don't know if we want to maybe spread stuff out a bit (e.g. test 1.17 on 3.7, 1.16 on 3.6) but at least this is fairly consistent and easy to remember.

drsteve:
This works for me. Covering both the oldest-supported and newest versions of numpy on each Python version is good. I'm not sure there's a convincing reason to test intermediate versions of numpy...

I'll go ahead and merge since the test suite is working.

Issue #311, opened 2020-05-11
test coverage: pybats


jtniehof:
There's a lot to try and nail down, so I'm not going to go looking at everything, and I know a lot of this is hit by the plotting issue. But coverage by module, excluding those over 80%:
init: 70%
bats: 39%
batsmath: 64%
gitm: 18%
qotree: 62%
ram: 9%
rim: 32%
trace2d: 17%

jtniehof:
I am self-assigning for trace2d (and I might be able to do a little something with ram.) Otherwise it's up to @spacecataz .

spacecataz:
I'll work on this progressively as I go; lots of tests to add, of course.
A lot of ram stuff is contingent on @drsteve's push updating that module as that PR makes quite a bit of changes.

drsteve:
That was merged already (#271)... but yes. Although I believe that PR also added some testing (not much, but some).

jtniehof:
Looks like pwom is completely untested

PR #359, opened 2020-06-23
Automatically read poorly formatted RIM files

spacecataz:
Small push to add flexibility to reading of RIM files.  On some systems, RIM produces output with line breaks in the middle of records.  While this could previously be addressed using the spacepy.pybats.rim.fix_format() function, this PR improves the Iono object so that it will now read such files without requiring a call to fix_format.

fix_format() now comes with a deprecation warning.

New tests for RIM have been added to exercise this new functionality as well as other previously untested features, partially addressing #311.

Small indentation change in pybats.__init__.LogFile to push reassignment of array outside loop.

- [x ] Pull request has descriptive title
- [ x] Pull request gives overview of changes
- [x ] New code has inline comments where necessary
- [x ] New features and bug fixes should have unit tests


jtniehof:
I don't have the mental bandwidth to really dig in today, but a couple of things that I noticed off the top:

Usually the import should be at the top of the file rather than in a function; we don't do that in a few places where it's importing a soft dependency that's only needed for that function. `os` and `warnings` generally should be present...we haven't reached the point of adopting a style and scrubbing everything for it, but should probably try to follow [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports) for new code in the meantime.

Also take a look at the [whitespace](https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements) stuff while you're at PEP 8; some of the new stuff is oddly spaced.

The docstring should mark the deprecation; since we're using numpydoc it really doesn't work well unless we follow [their standard](https://numpydoc.readthedocs.io/en/latest/format.html#sections). But wait! You can get the warning and the docstring fixup for free with `spacepy.deprecated`, and if you merge #376 it gets even cooler.

spacecataz:
Made changes per @jtniehof's comments. 
I don't think this one requires any update to the changelog.

jtniehof:
I think deprecating a function probably does merit a CHANGELOG entry so people are aware before calling it (or if it's called in a script where they don't check the output.)

jtniehof:
One other thing. The PyHC [standards](https://github.com/heliophysicsPy/standards/blob/master/standards.md), which we're trying to follow at least in spirit, discourage binaries in the repo if avoidable. They can disappear in diffs and if they change, they tend to actually take up more space than text files.

It's not completely avoidable--I can't think of a reasonable way to test pycdf without some CDFs in there, for instance--but we might want to consider ungzipping the test files (and then gzip them to a tempdir if necessary to test the gzip reading). It won't take up any more space, as the repository gets compressed and the release tarball is also compressed. I don't think it needs to be a blocker on this PR, either; it might be better handled by doing a scrub for binaries in the future.

PR #361, opened 2020-06-24
Various enhancements to pycdf.istp compliance checks

jtniehof:
This...grew.

Initially the concern was that the check for data in the file being within VALIDMIN/VALIDMAX threw an exception when the FILLVAL was of the wrong type. (I was testing a file that had a string for FILLVAL in a numeric variable....don't ask.) So as a result:

- VariableChecks.validrange now will complete with a bad FILLVAL rather than throwing exception.
- New VariableChecks.fillval to see if there _is_ a valid FILLVAL. This sort of duplicates funtionality in the existing NASA ISTP checks, but I think it's important because otherwise the validrange checks (which it can't do) won't necessarily return reasonable values.
- Suppressed some CDF warnings in the tests while I was here.
- Noticed FileChecks.depends didn't end a sentence with a period, so took care of that
- Got sick of manually adding every check to the `all` method, so automated that.
- That made it possible to subclass and add new methods, which made it possible to test the kwarg of `all` to catch exceptions in subtests. The test, of course, failed, so now it's fixed. The idea here is that if you're running _all_ tests on a file you probaby want a list of failures of things that could be checked and a statement that one of the tests completely blew up, rather than just a bare traceback of the one that blew up--often some of the tests that completed but failed are the reason for the test blowing up, and fixing those would fix the problem.

One simple fix that turned into a bunch of quality-of-life improvements in this corner of code.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -24,6 +24,7 @@ pycdf
    another variable
  - istp.VariableChecks adds check for VALID/SCALE MIN/MAX Entry types
  - istp.VariableChecks adds check for DELTA attributes
+ - istp.VariableChecks adds check for FILLVAL
  - istp.VariableChecks and FileChecks add check for empty attribute Entries
  - istp.VarBundle class to copy variables, slices, and depends between CDFs
  - istp.nanfill function to replace fill values with NaN
--- a/spacepy/pycdf/istp.py
+++ b/spacepy/pycdf/istp.py
@@ -130,7 +130,7 @@ class VariableChecks(object):
             except:
                 if catch:
                     errors.append('Test {} did not complete.'.format(
-                        f.__name__))
+                        func.__name__))
                 else:
                     raise
         return errors
@@ -687,7 +687,7 @@ class FileChecks(object):
             except:
                 if catch:
                     errors.append('Test {} did not complete.'.format(
-                        f.__name__))
+                        func.__name__))
                 else:
                     raise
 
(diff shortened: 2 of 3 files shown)


jtniehof:
The file I was working with still has more issues that cause less-than-helpful reports out, so I'll probably be doing some more work in here. Making draft for now.

jtniehof:
All done..basically added more cases of handling when SCALEMIN/SCALEMAX/VALIDMIN/VALIDMAX weren't values that were comparable with the data, similar to the handling of FILLVAL in 8317741.

jtniehof:
Well, I _think_ I'm done now, but I thought that before.....

jtniehof:
One weirdness I should note: isclose requires numpy 1.7 (!) I sort of backed us into a corner by saying both we'll bump the minor version of SpacePy if we bump dependencies, and that 0.3.0 is where we drop the Python 2 binary support. So I need an ugly workaround to keep 1.6 compatibility but will drop that in 0.3.0 and we can bump to 1.7.

balarsen:
There are conflicts in here. I am afraid of screwing something up here...

balarsen:
@jtniehof I am happy with this if you can sort the conflicts.

jtniehof:
Hmmm. Weird, because it wasn't showing conflicts before. That'll mean it has to wait until the build works again, so I'll fix up the rebase then.

balarsen:
See #381

jtniehof:
This should pass once #389 is merged.

balarsen:
I don't see a way to make Travis go again, @drsteve did this before as I recall.

jtniehof:
Yeah, looks like despite what it says, it doesn't merge master before trying the build. I did another rebase which should hopefully do it. I'll see what I can do about #385.

jtniehof:
Clearly I screwed up the rebase on this and I'm not going to try to push it through tonight....tomorrow....

PR #369, opened 2020-07-02
Convert between quaternions and rotation matrices

jtniehof:
This PR adds two functions to `toolbox`: `quaternionFromMatrix` and `quaternionToMatrix`. This permits conversion between the two representations of rotations. My use case is that I'm winding up in situations where I'm getting information in rotation matrices and would really have quaternions.

`quaternionFromMatrix` is a bit of a funny name; I'd ordinarily name it `matrixToQuaternion` but all the other quaternion functions start with `quaternion` so this seemed to make sense.


jtniehof:
Will have more thoughts tomorrow, but one thing affecting the round-trip is the sign ambiguity on the quaternion. If you've got a good idea for how to handle that, it would make things a little more deterministic and ease the round-trip testing. I've noticed a tendency when convering a sequence of rotation matrices to flip-flop (wild guess: probably eigenvectors with eigenvalues of 0.998 and 0.999 and the sense of the eigenvector with the maximum eigenvalue flops back and forth.) What I'm doing right now in my application is post-processing the output and if the real component of the nth quaternion plus the real component of the (n-1)th is near zero (i.e. simliar magnitude, opposite direction) and the dot product of the imaginary components of n and n-1 < -0.9 (i.e. axis of rotation reversed), I flip the sign of the nth. I could apply similar postprocessing. I could also always prefer the positive real component, but that would force an axis flip at some point in a sequence which would (presumably) screw up interpolation. So if you've got thoughts on that, all ears!

drsteve:
I think we can assume that it doesn't matter whether we return (i, j, k, w) or (-i, -j, -k, -w). For round-trip testing purposes we should probably treat them as identical and I can't see why particular, well-behaved cases shouldn't give repeatable round-trips. I think the method implemented does lend itself more to different signs in answers because of the eigenvalue approach, but that's not a problem.

Tracking sign and handling is probably useful if we also implement a SLERP (and I'd do that inside the SLERP), but I think we don't need to worry here.

Issue #364, opened 2020-06-25
test_AlphaOfK hangs with Python 3.8 and numpy 1.19.0 in CI


jtniehof:
We've had occasional Travis failures where the tests hang and eventually Travis kills for lack of output. It seems to always be the last build, which is Python 3.8 and numpy 1.19.0. With #363 I can see that, at least once, it's happened in test_irbempy.IRBEMBigTests.test_AlphaOfK. Restarting seems to work, so it might be an issue with running too many builds at once, but for now it's worth hammering on Py3.8/numpy 1.19 to see if it can be reproduced outside of CI.

jtniehof:
1.19 was just released a few days ago (it's not even on the front page of numpy.org yet) and this did start showing up a few days ago, so might be a 1.19 issue. We are building a test with 3.7 and 1.19 as well, so it would be a 3.8 + numpy 1.19 issue.

jtniehof:
3.8 and numpy 1.17.5 just hung on the daily build, can't tell where since #363 isn't merged yet.

jtniehof:
There's definitely something not properly cleaned up (3.8, numpy 1.18.5), might or might not be related...

```
python test_irbempy.py -v IRBEMBigTests.test_AlphaOfK
test_AlphaOfK (__main__.IRBEMBigTests)
test calculation of eq. pitch angle from K (regression) ... /home/jtniehof/miniconda/envs/py38/lib/python3.8/multiprocessing/pool.py:265: ResourceWarning: unclosed running multiprocessing pool <multiprocessing.pool.Pool state=RUN pool_size=8>
  _warn(f"unclosed running multiprocessing pool {self!r}",
ResourceWarning: Enable tracemalloc to get the object allocation traceback
ok
```

jtniehof:
3.8 and 1.19 just failed again (job 851.10, will kick to restart), in AlphaOfK.

jtniehof:
853.9 and 853.10, both Python 3.8 with numpy 1.17.5 and 1.19.0 respectively, both failed in AlphaOfK, which does seem to indicate it's not just "one out of ten" and probably an issue with Python version not numpy.

drsteve:
Looks like it's Python 3.8, see https://bugs.python.org/issue38744. There are other reports that seem similar, including issues with dask. It looks like we might be able to work around it though.

jtniehof:
Nice find. Thanks; that should be fairly easy to work around.

jtniehof:
[...]

Issue #351, opened 2020-06-08
get_Lstar() hangs randomly


wlwetzel:
I'm trying to call get_Lstar() a bunch of times (specifically to get the L and MLT at various times and locations). I haven't had a problem doing this in the past, but I'm having issues with this set of data. 

Calling get_Lstar() causes my program to hang, whether I loop through all the loci or try to pass them as an array. What's really driving me crazy is that get_Lstar() isn't hanging on any one input, it hangs randomly, and sometimes even runs without issue. 

### Minimal example to reproduce issue:

```
import numpy as np
from spacepy.time import Ticktock
import spacepy.coordinates as spc
import spacepy.irbempy as irb
import pandas as pd

test_data = '/path/to/data.txt'
data = pd.read_csv(test_data,names=['alt','lat','lon'],index_col=0,parse_dates=True,sep='\t')
times = data.index.tolist()
ticks = Ticktock(times)
loci = np.stack((data['alt'],data['lat'],data['lon'] ),axis=1)
for i in range(len(times)):
    tick = Ticktock(times[i])
    coord = spc.Coords(loci[i] , 'RLL','sph')
    print(irb.get_Lstar(tick,coord))
```
Here is the data I'm using:
2013-01-12 08:11:04	1.1330560370753462	-80.577	266.768
2013-01-12 08:11:08	1.1330560370753462	-80.483	265.443
2013-01-12 11:31:30	1.1330403352041967	-80.303	266.584
2013-01-17 14:47:24	1.1327892010673364	-71.558	253.659
2013-01-17 14:47:28	1.1328048933583328	-71.764	253.284
2013-01-15 13:48:24	1.1328362875204785	-72.932	265.828
2013-01-15 13:48:28	1.1328362875204785	-73.133	265.396
2013-01-15 13:48:32	1.132851989391628	-73.333	264.953
2013-01-21 16:44:40	1.1325537592214723	-66.347	231.194
2013-01-21 16:44:44	1.1325694515124687	-66.565	230.955
2013-01-11 11:52:20	1.1330246429132005	-80.105	263.596
2013-01-10 07:12:54	1.133008950622204	-78.151	261.141
2013-01-10 07:12:58	1.1329932487510546	-77.991	260.296
2013-01-10 07:13:02	1.1329932487510546	-77.828	259.471
2013-01-14 12:29:44	1.1329775564600584	-77.785	270.471
2013-01-14 12:29:48	1.1329775564600584	-77.949	269.652
[...]

PR #372, opened 2020-07-08
Fixes for python 3.8 and other warnings

jtniehof:
This PR started to address some warnings in Python 3.8 and apparently random bugs that might be related (#351, #364). I did clean up the handling of the Pool in get_Lstar, which addresses the ResourceWarning in #364 and hopefully the actual hang, and chances are it addresses the hang in #351, since other people are reporting issues with hangs in Python 3.8 when not properly closing multiprocessing Pools (https://bugs.python.org/issue38744) This is 58f2d1d4.

While I was at it, I cleaned some warnings from 3.8 regarding using `is` for comparison to literal (32efecd7), see the [3.8 release notes](https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-python-behavior), and a whole mess of numpy warnings. I then suppressed some CDF warnings in the CDF tests. The upshot is that the tests should be warning-free on latest numpy and Python.

The only numpy warning that wasn't trivial was d0c5fdb7. This actually didn't raise a warning on the newest numpy, but was relative to a change in numpy 1.11 where masked arrays no longer do a copy-on-write if they share a mask and there's an assignment. I've found masked arrays almost _always_ seem to share their masks (with what, I don't know), so I added an explicit unshare to quiet the warning. This is probably not a problem--there's probably not another reference to that array running around which might accidentally get written to--but it seems better to be safe.

### DIFF
--- a/spacepy/data_assimilation.py
+++ b/spacepy/data_assimilation.py
@@ -610,7 +610,7 @@ def getobs4window(dd, Tnow):
         L_lower = np.mean([ Lgrid[i], Lgrid[i-1] ])
         L_upper = np.mean([ Lgrid[i+1], Lgrid[i] ])
         indices = np.where( (L_lower < L) & (L <= L_upper) )
-        if np.size(indices) is 0: continue
+        if np.size(indices) == 0: continue
 
         Lmean = np.append(Lmean, iL)
         ymean = np.append(ymean, np.mean(y[indices]))
--- a/spacepy/datamodel.py
+++ b/spacepy/datamodel.py
@@ -1080,7 +1080,7 @@ def toHDF5(fname, SDobject, **kwargs):
                     try:
                         if value.nbytes: truth = True #empty arrays of any dimension are nbytes=0
                     except AttributeError: #not an array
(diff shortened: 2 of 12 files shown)


balarsen:
These are good fixes. Do you know of good guidance on `is` vs `==`? I believe that `is` is a memory location check for sameness and `==` calls `__eq__` on the object. So:

My understading

- Bad: `a is 'a string'`
- Bad: `a is 1`
- Good: `a == 'a string'`
- Good: `a is None` as None is an immutable special thing

jtniehof:
`is` checks same-ness, which on CPython is often memory location (tuples and strings are de-duplicated since they are immutable, so something constructed with the same results in referring to the same) but that's really not guaranteed. My rule of thumb is that equality comparisons are usually the way to go except as recommended in [PEP 8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) ... `None` is the clear case but True/False are often appropriate. (Basically do either `if foo` or `if foo is True` depending on situation, usually not if `foo == True`).

PR #376, opened 2020-07-09
Improvements to deprecator decorator

jtniehof:
There's a function decorator to easily deprecate a function in the top-level module. This PR makes several improvements:

- Gets the formatting of the wrapped function's docstring correct; indentation used to be based on the first line and that wasn't always right. It was also adding an extra linebreak at the end, which wasn't needed.
- Allows separation between the message in the DeprecationWarning and what gets added to the docstring
- Some work on getting the function signature to carry through. The only real functional change here is a note in the docstring that it doesn't carry through properly in Python 2, but it's a big diff because of an unwrapping of a function-in-a-function that made the operation a lot less clear.
- Fix up the deprecated quaternion functions in toolbox using these new features
- Mostly unrelated but I was looking at it, fix the docstring for quaternionConjugate where the parameter was "Qin" in the signature but "vec" in the docstring.

### DIFF
--- a/spacepy/__init__.py
+++ b/spacepy/__init__.py
@@ -95,6 +95,46 @@ if sys.platform == 'win32':
     else:
         os.environ['PATH'] = minglibs
 
+#actual deprecation decorator
+def _deprecator(version, message, docstring, func):
+    #this is the actual, deprecated function
+    @functools.wraps(func)
+    def _deprecated(*args, **kwargs):
+        warnings.warn(message, DeprecationWarning)
+        return func(*args, **kwargs)
+    if func.__doc__ is None:
+        doclines = []
+    else:
+        doclines = func.__doc__.split('\n')
+    # Docstring SHOULD be a single non-blank line with summary, blank line,
+    # and then the rest of the content. Want to put the deprecation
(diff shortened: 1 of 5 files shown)


spacepy/__init__.py
```diff
@@ -95,49 +95,105 @@ def help():
...
+        Message to include in the deprecation warning and in the
+        docstring.
+
+    Other Parameters
+    ================
+    docstring : str
+        If specified, this will be used in the docstring instead of
+        ``message``. It can be a multi-line string (separated with
+        ``\\n``). It will be indented to match the existing docstring.
+
```
spacecataz, line 161: It took a moment to realize the utility of `docstring` vs. `message`.  I recommend adding, "``message`` is still used in the warning message" or something similar in the docstring here.

jtniehof: Tried to clarify this in the update; mark "resolve" if you're good with it.

jtniehof:
The only change to the quaternion stuff was to copy the actual function signature and brief description from the new location of the functions. It's so if you're looking at the docs (or help) in the old location you get something useful besides just a referral to the new place.

PR #378, opened 2020-07-13
Doc Makefile: remove autosummary source files with "make clean"

jtniehof:
Autosummary sources are autogenerated, and the check for updated docstring inputs isn't always perfect. "make clean" should remove all intermediate/automatic steps to ensure a clean build from scratch, so this PR adds removing the autosummary sources.

Yes, there are two nested directories of "autosummary"; it's because a toctree is being defined twice, but fixing it breaks the doc build, so that's going to have to wait for a more involved docs overhaul.

### DIFF
--- a/Doc/Makefile
+++ b/Doc/Makefile
@@ -34,7 +34,7 @@ help:
 	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
 
 clean:
-	-rm -rf $(BUILDDIR)/*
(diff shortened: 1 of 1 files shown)


jtniehof:
"the check for updated docstring inputs isn't always perfect".  If you update a docstring, install, and rebuild the docs, the autosummary extension doesn't always pick up the change. Clearing out the autosummary-generated docs fixes that. Also weird cases like if a function moves or is renamed, the old one is left around. I have to do this by hand all the time.

Issue #321, opened 2020-05-15
Ticktock should convert to/from AstroPy time


jtniehof:
Surrounding discussion of #286, we should support conversion of TickTock to/from AstroPy time. With most of our time systems we're implementing getXXX() functions and then storing that representation in a .XXX attribute. We can do the same thing with astropy.time; I suggest "APT" but would be up for other options.

We could also have explicit conversion functions that get called differently, but I'd rather stick with doing it as similar to the other systems as possible. (In the future I might want to replace some of those with properties, but that's a different issue.)

I can probably handle this one.

Issue #317, opened 2020-05-14
Ticktock: ability to add other than timedelta


jtniehof:
Ticktock add/subtract methods operate on timedeltas and thus fall through to Python datetimes, and can't properly handle leapseconds. Should add ability to use other types (either seconds or Julian days as floats?).

balarsen:
Can the numpy datetime64 do leap seconds? Would really love to find another
implementation that we can use or copy.

jtniehof:
Nope, it can't. I've argued with them about it but they've said they won't support it unless I can tell them what (2050, 1, 1) - (2049, 12, 31, 23, 59, 59) is, right now. They don't get that UTC isn't defined in the future.

balarsen:
That is frustrating. We need to think hard on how we handle it. For differences a conversion to TAI and add/subtract is great and should work. Our implementation of checking for leap seconds in the UTC->TAI conversion is a huge bottleneck currently. 
https://github.com/spacepy/spacepy/blob/0fbb879a5d9f50fa2d22bf5b056ba60f1edac7b6/spacepy/time.py#L1281

jtniehof:
My first thought would be to allow adding numerical only for numerical representations (TAI, MJD, GPS, CDF Epoch) and have it operate on the "native" representation of the Ticktock. So if you add 1 to MJD it's a day, 1 to TAI is a second, 1 to Epoch is a millisecond. Alternatively I'm okay with saying it's always seconds.

balarsen:
My take would be to decide on an internal time base (I vote TAI) and just always convert to that and do all calculations there then transform to the output. From there math is seconds or with timedelta objects would be pretty easy. 

This implies we need the input to TAI conversion to be as efficient as possible.

PR #390, opened 2020-08-05
time.Ticktock overhaul

### LINKED ISSUE #319: Ticktock should handle JD/MJD without datetime
Conversion in and out of JD/MJD right now involves going through datetime, which loses the ability to represent leap seconds even though JD/MJD can handle it. The implementation needs to be fixed to handle that.

### LINKED ISSUE #320: Ticktock should handle GPS without datetime
Conversion in and out of GPS time right now involves going through datetime, which loses the ability to represent leap seconds even though GPS can handle it. The implementation needs to be fixed to handle that. Should be a fairly easy matter of falling back to TAI.

### LINKED ISSUE #381: matplotlib 3.3 breaks spacepy <=0.2.1
Our cron build of master has failed three days running. Python 2.7 on bionic is fine, 3.5 on xenial is fine, but 3.6, 3.7, 3.8 on bionic have test failures. I haven't done a diff of the outputs but they seem to all be nearly (but not exactly) the same. A lot of it is related to functions that use OMNI input so maybe there's something weird with the test db there.

All of this has succeeded before so there's probably some change in the Travis environment exposing some sort of fragility. There are also some numpy warnings showing up, and I got all the Python 3.8 and  numpy warnings squashed in #372, so our version of numpy may have shifted.

jtniehof:
This PR closes #319, closes #320, and closes #381, and lays the groundwork for #321. Basically it's a substantial overhaul (not really rearchitecting) of Ticktock to keep it from always falling through to datetime, particularly when not appropriate.

Note: When I say "UTC" here I mean "Ticktock.UTC", i.e. datetime representing UTC.

## Completed changes

There are some more tests of JD/MJD and some general cleanup of that code; basically to orient myself to it and prepare to rework (c4707880b16ef0d967c81cb1f3e642d688f3cfa9).

`update_items` lost the superfluous `cls` argument and passing in anything other than `data` as the argument now actually works, plus a few related tests and fixes (e.g. eDOY wasn't updated) (30ff6e98a3668539005fdae5ddf72baaa66bd095).

There's a big update of documentation just on how all the various timebases are computed from others (cdcfa13a451117928c854e6ddcbb21603f09b8a1).

Then a whole series of commits where, if you call getX, and X was the dtype, it actually preserves the data attribute rather than smashing it from UTC. This basically maintains the data passed in as the key reference. The exception is ISO, since the passed-in data might not be a full ISO date + time string. This series also includes little related bugfixes--for instance, ISO was supposed to properly handle leapsecond output but didn't (596036272e03cf3cbfbec242aacf090a7065fc28) and this was only apparent because leapseconds were smashed, and you couldn't use Unix times as input on Python 3 (f20cee73b507f2c4098aacdd256aaf234f1e7a5b).

I changed the sorting to be by TAI (a300af2d367715b8fa269e208c8a85c532c3ccb2).

There was a huge error in calculating leap seconds before 1961 (61bb56d3a17a6ab51d17923eb7b6f74219519686).

There are a whole lot of other commits that aren't specifically linked, so browing through the history is recommended. https://github.com/jtniehof/spacepy/commits/mpl_33_fixes

On master, leap seconds "bounce" into the next day. E.g. TAI 1514764832 is UTC 2005-12-31T23:59:60 but master renders in Ticktock UTC as 2006-01-01T00:00:00. I updated this to truncate, e.g. 2005-12-31T23:59:59.999999; this is the way pycdf handles it for TT2000. Note that:
`datetime.datetime(2005, 12, 31, 23, 59, 59, 999999).strftime('%Y-%m-%d:%H:%M:%S')`
renders as `2005-12-31T23:59:59`, i.e. it doesn't round, so I think this gives the correct impression. The old way also didn't handle sub-second very well; I didn't explicitly verify this but TAI 1514764832.5 would probably have been 2006-01-01T00:00:00.5 which would come AFTER TAI 1514764833 (2006-01-01T00:00:00). I've tested this case as well. This change can be backed out, but it's going to affect stuff going forward. (0b2c0bbb6b0dac37d188a27822320e4c71957efe)

Ticktock used to truncate fractional leap seconds. E.g. on 1965 1 September TAI-UTC is actually 3.84 but Ticktock treated it as 3. I changed this to rounding (treat this as 4) (b9e557f2303944ee493227e5db5282d74d59a742). Actually handling them properly isn't in the cards, at least for now, because in the fractional leap second era the TAI second and UTC second aren't the same length, so doing it 100% right requires the implementation of a scaling factor in addition to fractional offsets, a much bigger change. (#393)

ISO input and output now support dates before 1900 on Python before 3.3 (42b128d86a47a07f4ed07829c4b54768e4a29359). This fix can of course go away with #26.

TAI can now parse directly from ISO input so it can get leapseconds.

JD, MJD, GPS, CDF, UNX are now TAI-based. UTC now converts from TAI if the inputs are anything but ISO. TAI can now convert from anything.

RDT is TAI-based. UTC converts from TAI on RDT input now. Interestingly enough, this is the first case where I've run into precision limitations...converting TAI directly from RDT actually results in the output being off by a little bit compared to converting RDT to UTC to TAI. There's probably some sort of protection of small differences going on in that later path; I'm going to check some of my helper functions to try and improve that.

now() and today() use utcnow(). They used to use just now, which gave you something that claimed to be UTC (had the UTC label, converted to TAI as if it were UTC, etc.) but was actually in local time.

In the other matplotlib department:

pycdf's epoch_to_num is now fixed; it'll use get_epoch if matplotlib is installed but otherwise use the old values. smartTimeTicks test is also fixed (only the test had to be changed.) I've confirmed the other uses (e.g. in toolbox) are unaffected by the exact values.

## Open questions
omni does a lot of stuff via RDT. Should this change to MJD or TAI? I think that would have to be another PR, but we should open an issue so as not to miss it.

Does it make sense to start using longdoubles (128 bit) for most of the time representations? (MJD, JD, TAI at least, but probably also GPS). Going back and forth between MJD and TAI (for instance) doesn't really lose _precision_ but it does lose _representability_, since 1/86401 isn't exactly representable as a binary float. Another alternative would be to use an internal representation of TAI days and seconds-since-start-of-day, but that's a pretty core design change (having an internal representation instead of allowing ``data`` to be anything and converting among.) I played with longdoubles a bit and it doesn't seem to fix things; I wonder if it would make sense to round the output of the RDT to TAI conversion to get more expected results. E.g. if we only have .01s precision, actually round to .01 so we don't have a mess of extra meaningless digits.

One JD and MJD support leapseconds, we can either have MJD always start at midnight, or we can always have `MJD = JD - 2 400 000.5`. This is because on days with leapseconds, midnight is not halfway between noon and noon. [Resolution B1](https://www.iers.org/IERS/EN/Science/Recommendations/resolutionB1.html) seems to strictly define MJD as JD - 2 400 000.5, but it also gives the wiggle room of "care be exercised to state [MJD] definition specifically."

I implemented truncating leap seconds to the end of day rather than pushing to the start of the next day, and rounding TAI-UTC instead of truncating. Since these are behavior changes it would help if people screamed sooner rather than later!

## Changes to come
This is done: ~~The main thing to finish here is actually getting away from UTC as the baseline. Right now everything derives from UTC. This makes sense for some things (e.g. RDT, CDF) but not for others (TAI, GPS). The plan is that everything will convert from the thing that makes the most sense, rather than directy from UTC, and I'll make sure that there's a path from anything to everything else. So GPS will convert from TAI but TAI will know how to convert from a lot more things.~~

- [x] GPS now converts to/from TAI but the commit isn't tagged as closing #320, so should fix that.
- [x] Get JD/MJD compatible with leap seconds; this might be working from TAI directly or UTC + leapsecs (#319)
- [x] Reject UTC and ISO inputs between 1582-10-5 through 1582-10-14, as they don't exist. (Punt, #397).
- [x] Support dates before 1900 in ISO input, as strptime doesn't allow them. (This might have to be deferred to #395).
- [x] Change RDT not to use matplotlib (part of #381)
- [x] Fix smartTimeTicks tests (#381)
- [x] Convert TAI straight from ISO, so that ISO inputs with leapseconds get the correct TAI.
- [x] Convert Ticktock.now to UTC instead of local.

Not going to add AstroPy time (#321) in this PR; it's massive enough already.

## PR Checklist
This is going to be updated as things are actually finished.


jtniehof:
Regarding the fractional seconds, note our handling here inserts six leap seconds at the end of 1971, the transition from "slow UTC seconds" to purely integral leap seconds. It won't be fixed by rounding and, again, I kind of think is out of scope. Another side effect is that we'll ultimately accept times like 1963-12-31T23:59:60, which didn't actually exist. I just want to be clear on what will and will not get fixed with this change!

jtniehof:
(And yes, I'll get more documentation in place on these issues.)

jtniehof:
Just to call out something in the description above: because there were no comments here, I made the decision to truncate leap seconds to end of day rather than pushing them into the next day.

balarsen:
I would love to see a C backend for ISO formats of a few varieties. I have started using pandas for this task as it has a nice C backend. Could use theirs but I am not sure the dependency is worth it as it is not that hard. Maybe round two can hit that. I’ll put in a feature request so it doesn’t get lost.

balarsen:
On the fractional leap seconds. That is super ugly and not our fault. In the docs can you make statements like “accurate to within ...X” after sone date or the like?

jtniehof:
Similarly, now rounding TAI-UTC rather than truncating.

jtniehof:
@balarsen : I added a pretty extensive discussion of conversions, leap seconds, etc. to the time module, and just added a note on the magnitude of differences through that time (of order 5sec.)

This is making that docstring pretty heavy, but that's in keeping with our style elsewhere. I have ideas for that but it's out of scope for this PR.

jtniehof:
Calling out the "open question" about definition of MJD (midnight-to-midnight or JD - 2400000.5)

balarsen:
I am not into time enough to have an opinion here. Anyone else make a decision we like that we can reference and use?

drsteve:
So isn't this just a question about the timescale we're representing with JD? I.e., are JD and MJD on the UTC scale or the TT scale?
What I mean is that if we're representing JD(UTC) then a day is a UTC day, and midday UTC is always integer JD(UTC) and midnight UTC is always integer MJD.
If we're representing JD(TT) then it would need to line up with the TT midday/midnight.
So in either case both should be true: MJD is JD-2400000.5 **_and_** MJD is always integer at midnight. On the UTC scale that implies that day length varies on leap second days.

IIRC we had decided that we were representing JD as JD(UTC). Perhaps I'm misremembering? Of course, if I'm getting something wrong here LMK.

jtniehof:
It's JD(UTC), and yes, the day is longer, but it's not symmetric about midnight (which is the dividing line of MJD days).

On UTC day N with a leapsecond, it's 43201 seconds from noon (start of JD M) to midnight of UTC day N + 1, and then 43200 seconds from midnight to noon of UTC day N + 1. Thus the 0.5 mark of JD M comes half a second before midnight.

jtniehof:
This is finally ready to go. Sorry; it's a huge one, which is why I've requested reviews from everybody. I fear @drsteve is probably going to be the stuckee on final merge, but it would appreciated if everybody would bang on it as much as possible and run any code you've got that uses Ticktock.

jtniehof:
"Open questions" is pretty much still open, should look at as part of review.

CHANGELOG
```diff
@@ -39,6 +40,22 @@ pybats
...
+   within the leap second is now converted to the end of the day rather
+   than the beginning of the next day.
+ - Fractional leapsecond counts are now rounded to the integer rather than
+   truncated.
+ - Fix conversions to ISO before 1900 on Python 2.
+ - Include leapseconds in JD/MJD calculation, so the length of a second on a
+   day with leapseconds is slightly shorter but leap second times are not lost.
+ - Most time systems are now converted to/from TAI. This may cause small
+   differences with previous versions of SpacePy, on order of a double
+   precision. Time conversions around leapseconds may also be different;
```
drsteve, line 55: On order of a double precision isn't going to mean much to a lot of users. This should be less than a millisecond, right?

jtniehof:
It depends, and I think you bring up a good point that should be
quantified, in CHANGELOG and release notes. The float precision is relative
so for e.g. RDT it's worse, more milliseconds, than MJD (and less than JD).
I'll work up the docs.

balarsen:
Related to #395 

dateutil.parse 89.25317879764546
dateutil.isoparse 11.910058648131903
ticktock 306.8038737841839
datetime.strptime 14.634390825003939

from 627432a36832ec50f3a21ca2c4e59491674203b7. Something made ticktock 3x slower... doesn't have to be fixed now but worth noting and will need some eyes.

jtniehof:
I'll look in detail, but the ISO input now does two new things. It handles
leap seconds, and it converts .ISO to a consistent format. So no matter
what weirdo set of strings comes in, it's rendered as the desired format,
and .ISO is made at initialization time.

balarsen:
> 
> I'll look in detail, but the ISO input now does two new things. It handles leap seconds, and it converts .ISO to a consistent format. So no matter what weirdo set of strings comes in, it's rendered as the desired format, and .ISO is made at initialization time.

I took a quick look and one thought is to fall back to `[dateutil.parser.isoparser](https://dateutil.readthedocs.io/en/stable/_modules/dateutil/parser/isoparser.html)` to get us into datetimes then one can do all the mucking with leaps etc. 

For me speed for well formatted (like Magephem h5 files) is more important that string format flexibility.

balarsen:
Pycharm is complaining in many spots, I culled the list to ones I thought might be worth a look. I will post them here for you to decide if they matter. The line numbers are time.py in  627432a

- Class 'Iterable' does not define '__getitem__', so the '[]' operator cannot be used on its instances: 439
   - Lots of these all over, I don't understand them but figure I will point them out
- Local variable 'dtype_func' might be referenced before assignment: 341
   - Might be true there is a if...else then a try...except to get here. Caught in a test?
- Redundant character escape '\.' in RegExp: 1861
- Unused import statement 'from spacepy import DOT_FLN': 2099
- PEP 8: W605 invalid escape sequence '\g': 1861
    - does this need the "r" before the string?
- Global variable 'secs' is undefined at the module level: 2100
    - this is true, care?


Other things I note:
- Why are you falling back to UNX for the the `__gt__`, `__lt__` etc and not TAI that seems like it is the base? Seems like extra conversions. 
- in `_tai_naive_to_real()` docs around Returns are messed up (no =====)
- in `_tai_real_to_naive()` docs around Returns are messed up (no =====)

drsteve:
@balarsen `dateutil.parser.isoparse` doesn't handle leap seconds. That is, if you give it a leap second in your ISO input it errors:
```
>>> dateutil.parser.isoparse('2015-06-30T23:59:60')
```
will raise a ValueError and complain that seconds have to lie in the range 0..59
Ticktock will handle that situation now.


It's very unlikely to help with the speed, but the regex on line 1861 could be compiled before the for loop and then the `subn` method from the compiled pattern can be used. 
It's also possible that the string representation in the MagEphem files is just a bad choice... those HDF5 files also have numeric systems, and I'd probably get a Ticktock and the `UTC` (datetime) from the `GpsTime` variable if speed was a concern.

balarsen:
I agree except almost never will we actually hit the leap second. So even checking for it then falling back will still be faster, or failing and catching. Magephem is only one example and the only public one I have. Maybe we need to move to C but that is major work. My standard data is daily files 16 samples per second and ticktock is way too slow. Pushes people away from spacepy, me included, I typically use pandas for time instead. 

Could split times on the T and cache the conversion via lru cache or other smart things.

balarsen:
What do I not understand about leap seconds here? This seems incorrect:

```
In [1]: import spacepy.time as spt
In [18]: spt.Ticktock(['2016-12-31T23:59:60.999999']).TAI -  spt.Ticktock(['2016-12-31T23:59:60']).TAI
Out[18]: dmarray([0.])
```
Seems this is a valid time and should give the answer 1. 

This seems correct but inconsistent with above. 
```
In [20]: spt.Ticktock(['2017-01-01T00:00:00']).TAI -  spt.Ticktock(['2016-12-31T23:59:60']).TAI
Out[20]: dmarray([1.])
```

drsteve:
@balarsen Has found an issue in his comment above. It looks like fractional seconds within the leap second are all truncated to the start of the leap second.

This issue and another are wrapped up in the following code snippet:
```
In [1]: import spacepy.time as spt

In [2]: tt = spt.Ticktock(['2015-06-29T23:59:59.5', '2015-06-29T23:59:60', '2015-06-30T00:00:00', '2015-06-30T23:59:59', '2015-
   ...: 06-30T23:59:60','2015-06-30T23:59:60.5', '2015-07-01'])

In [3]: tt.TAI.tolist()
Out[3]:
[1814313634.0,
 1814313635.0,
 1814313635.0,
 1814400034.0,
 1814400035.0,
 1814400035.0,
 1814400036.0]
```
There's a leap second at the end of June 30th 2015. There isn't one on the previous day.

Concerns:
1. The first time is on a half-second, but TAI (and GPS, etc.) don't reflect that. If I start from other time systems we're all good, but if I start from ISO I lose fractional seconds. I think this is the cause for @balarsen 's previous comment.
2. The erroneous ISO leap second doesn't raise an error, it silently falls to being one second later (i.e. the start of the next day). So the times that come out are largely correct (e.g., TAI is correct), but the ISO representation is wrong.
3. The ISO representation is not valid and it propagates to the UTC representation as if it were valid. That means that the UTC/datetime is also 1 microsecond off.
4. This is just point 1 again really. The last 3 times in the list are all a half second apart, but TAI (and GPS) doesn't show the fractional part.

drsteve:
Also, I realize that the `isoformat` method can be blamed on me, but it's aggravating me in this round of testing. And it's possible that it's where the TAI truncation issue is coming in under this revision...

If I make Ticktock from ISO I get the issues highlighted above. If I then change the `isoformat` to microseconds then force a recalculation using `getTAI` then my TAI is correct and has fractional seconds. `isoformat` was intended only to change the display, not the precision, but I think with the change in ISO parsing it's now forcing a change in precision.

- This can be fixed (though not cleanly) by changing the default isoformat to 'microseconds'.
- We should allow setting the display precision on `Ticktock` instantiation (new kwarg?)
- It would be nice to be able to set the default display ('seconds' or 'microseconds') in the `.spacepyrc` and/or at the module level

jtniehof:
Going to go back through all the comments, but the ISO parsing thing is going to require some discussion and design, so comments on that first:

I did add using the user/class-specified output format as the first thing to check when parsing, figuring if somebody set the format they might well be passing stuff in with that format. I'm not sure it's what's biting us, as the UTC still has the fractions, and this is the case both on this branch and on master:
```python
In [2]: spacepy.time.Ticktock('2010-01-01T00:00:01.123').UTC[0]
Out[2]: datetime.datetime(2010, 1, 1, 0, 0, 1, 123000)
In [3]: spacepy.time.Ticktock('2010-01-01T00:00:01.123').TAI[0]
Out[3]: 1640995235.0
```
So not a new problem, but one that should be fixed as part of this PR. My guess is there's a spurious cast to `int` in there somewhere; I will work on it.

Part of the reason for taking the user's specification first is the fallthrough to `dateutil.parser`. This tries so hard to get _something_ that occasionally it's ridiculous:
```python
In [6]: dateutil.parser.parse('2010-010T00:00')
Out[6]: datetime.datetime(2010, 10, 19, 0, 0)
```
Uh, no. So the user should be able to give a specification. Obviously since `isoformat` is an instance method the idea is for the change to be on output rather than input. (Incidentally, `isoformat` claims it will recalculate `ISO` but it doesn't, should it?) So that means the only way to get a different input format with what I've set up is to fandango on the class member, which is not particularly brilliant. So maybe add an argument to `__init__` that's used only for parsing, or that changes the instance's value of _isofmt for both parsing and output? I do sort of like the idea of having the output and input representations match by default; are you okay with that if it doesn't cause problems with fully parsing the input?

`isoparse` might be part of the solution but it requires a fairly new dateutil and also starts to get into the speed question. I really don't want to worry about optimizing this until we get the behaviour we want both defined and implemented.

jtniehof:
The "we'll accept anything as ISO input" is really nice but also an endless rabbit hole...you can pass in all sorts of junk, which I rather like, but has implications for speed, code nastiness, and unexpected results.

@balarsen, we can't use isoparse and then do leap seconds after, because the string has to be sanitized of leap seconds first. Also, if there are no leapseconds on the input, the fix-it-up loop is empty and basically a no-op; it's already essentially skipped.

jtniehof:
5e0448e6aa7faf3128e2e21d40462328238b6207 has updated documentation on the scale of possible changes from double precision. I put this with the "what gets calculated from where" information in Ticktock.

jtniehof:
@drsteve Re: tickrange: yes, that's a followup. I have not touched anything other than the core Ticktock. That includes not messing with the add/subtract stuff. In particular this does _not_ do anything about #317. Again, core functionality first, then we can work on other stuff. Please do open issues if there's something that you think needs to be done.

@balarsen , that applies to the gt/lt as well.

jtniehof:
@balarsen , re: pycharm: that's the very first commit in the branch, so the line numbers are going to be hideously out to lunch, and my level of caring about pycharm is pretty low. Long term I'd rather get pylint working right....

    Class 'Iterable' does not define 'getitem', so the '[]' operator cannot be used on its instances: 439
This is pycharm getting confused by the fact that we attach the timesystem-specific attributes to the instance later on.

    Local variable 'dtype_func' might be referenced before assignment: 341

Not relevant to the PR (existing code), and yes it's likely not to be hit because of the if-else but it's sloppy at best.

    Redundant character escape '.' in RegExp: 1861

I think this is bogus. Literal . needs to be escaped.

    Unused import statement 'from spacepy import DOT_FLN': 2099

Not _really_ unused but a casualty of flailing because of #391. Fixed.

    PEP 8: W605 invalid escape sequence '\g': 1861
        does this need the "r" before the string?

Yeah, this is probably clearer with the explicit raw string. 

    Global variable 'secs' is undefined at the module level: 2100
        this is true, care?

Actually not true; it's declared as global here and then defined later in the function.

jtniehof:
in _tai_naive_to_real() docs around Returns are messed up (no =====)
    in _tai_real_to_naive() docs around Returns are messed up (no =====)

Somewhat irrelevant because they're (intentionally) not rendered, but worth having right.

jtniehof:
Converting `np.integer` or `np.signedinteger` to a dtype is deprecated. The current result is `np.dtype(np.int_)` which is not strictly correct.

Ugh. I wish numpy had actual good documentation on their specific vs. generic dtypes. The end result is thrash until it stops complaining, so count another thrash.

jtniehof:
The erroneous ISO leap second doesn't raise an error, it silently falls to being one second later (i.e. the start
    of the next day). So the times that come out are largely correct (e.g., TAI is correct), but the ISO representation
    is wrong.

The question is how much we want to care about this. I can add in checks, but we don't have checks for every possible invalid entry--e.g. we don't catch `datetime.datetime(1582, 10, 9)` which doesn't exist. My focus here was primarily on getting valid output for valid input. This is still a _new_ inconsistency which wasn't there before. I've got a notion I'll try....

jtniehof:
Now errors on invalid ISO leapsecond inputs.

The ISO-to-TAI-subsecond thing is tested for the case where the input format permits it. I'm honestly not sure how the UTC is getting the subseconds when the TAI isn't; literally the same code. But the issue is parsing with the output type so until we have a design call on that (comments above), it's not worth hacking out. 

The ISO parsing check means the ISO function now needs the leapsecond record. That means it _has_ to be read on import. So I changed out the time reference in datamodel and did an obnoxious import-in-function thing for the datamodel references to toolbox. It's ugly but at least it's a step forward and it works.

drsteve:
I'm happy with functionality, and _**almost**_ everything looks like it works.

- I have found a case that doesn't seem to be working as advertised. The TAI and GPS times given are for the same instant (easy to check), but the calculated UTC doesn't match up. The values calculated starting from GPS time don't push the leap seconds into the end of the UTC second (in this case the TAI, ISO, etc. all seem to be correct). 

```
In [46]: spt.Ticktock([1861920035.5, 1861920036.5], dtype='TAI').UTC
Out[46]:
dmarray([datetime.datetime(2016, 12, 31, 23, 59, 59, 500000),
         datetime.datetime(2016, 12, 31, 23, 59, 59, 999999)],
        dtype=object)

In [47]: spt.Ticktock([1167264016.5, 1167264017.5], dtype='GPS').UTC
Out[47]:
dmarray([datetime.datetime(2016, 12, 31, 23, 59, 59, 500000),
         datetime.datetime(2016, 12, 31, 23, 59, 59, 500000)],
        dtype=object)
```

- Docs-wise I can't see anywhere that makes a recommendation for what to use when passing times to/from other software or libraries. We should probably have a note for this, since our JD/MJD are JD(UTC) and MJD(UTC), and the way we handle leapseconds for that may not be identical to other software. Folks might assume that JD is the best system to use, but I think we should explicitly recommend TAI or GPS.

- Error-wise, there's still a case that's not reporting the desired error. That is, for incorrect leapsecond times *after* the last known leapsecond we're not hitting the error saying it's not a valid leapsecond:
```
In [24]: mjd = spt.Ticktock(['2017-12-31T23:59:60.5']).MJD
IndexError                                Traceback (most recent call last)
<ipython-input-24-6210abcb82be> in <module>
----> 1 mjd = spt.Ticktock(['2017-12-31T23:59:60.5']).MJD

~/miniconda3/envs/default/lib/python3.7/site-packages/spacepy/time.py in __init__(self, data, dtype, isoformat)
    366         else:
    367             # ISO is populated by update_items
--> 368             self.update_items('data')
    369             if dtype.upper() == 'TAI':
    370                 self.TAI = self.data

~/miniconda3/envs/default/lib/python3.7/site-packages/spacepy/time.py in update_items(self, attrib, cls)
    883                     del self.ISO # Also will recalc the ISO
    884                     del keylist[keylist.index('ISO')] # So no need to calc again
--> 885             self.TAI = self.getTAI()
    886             if 'UTC' in keylist and self.data.attrs['dtype'] != 'ISO':
    887                 self.UTC = self.getUTC()

~/miniconda3/envs/default/lib/python3.7/site-packages/spacepy/time.py in getTAI(self)
   1497
   1498         if self.data.attrs['dtype'] == 'ISO':
-> 1499             isoout, UTC, offset = dtstr2iso(self.data, self._isofmt)
   1500             if 'UTC' not in dir(self):
   1501                 self.UTC = spacepy.datamodel.dmarray(

~/miniconda3/envs/default/lib/python3.7/site-packages/spacepy/time.py in dtstr2iso(dtstr, fmt)
   1947             utcday = np.frompyfunc(lambda x: x.date(), 1, 1)(UTC[leapidx])
   1948             closestday = np.searchsorted(leapsecday, utcday)
-> 1949             bad = utcday != leapsecday[closestday]
   1950             if bad.any():
   1951                 raise ValueError('{} is not a valid leapsecond.'.format(

IndexError: index 32 is out of bounds for axis 0 with size 32

```

jtniehof:
So just to be explicit, three pending issues:

- [ ] Calculation of UTC from GPS time during leapsecond `[1167264016.5, 1167264017.5]`. This is really weird since it should go to TAI first, so I'll also double-check the TAI conversion.
- [ ] Document recommendation of TAI for transferring to other environments
- [ ] Passing in ISO string with a nonexistent leapsecond after the last existing leapsecond raises an error that's an obvious implementation problem, rather than a "bad leapsecond" ValueError. I presume this is the traceback.

jtniehof:
@balarsen , should I also put some note in `isoformat` for the constructor that specifying this might speed stuff up?

jtniehof:
Inkscape is only used if building the LaTeX docs.

jtniehof:
Requested changes done. The conversion to UTC inside of leapsecond was actually broken for everything except TAI and ISO, so that was a good catch.

balarsen:
> 
> 
> Approving.
> If @balarsen or @spacecataz want to give a final look over and approve then I am happy to merge so we can get on with the rest. Or one of you can merge, either way...

I'm happy now.

PR #403, opened 2020-08-20
datamanager.rebin example: specify axis limits (Closes #402)

### LINKED ISSUE #402: axis limits on rebin example
The example for datamanager.rebin needs explicit axis limits on the plot, it's not automatically choosing a very good range.

jtniehof:
This PR adds explicit axis limits to the example code for datamanager.rebin, so closes #402. Whether or not this was an issue depended on what version of matplotlib you were using, but in some cases the axis limits were set way out and there was lots of whitespace. I didn't add the limit setting to the example code in the docstring, just the behind-the-scenes code that actually renders the plot.


balarsen:
@drsteve, @spacecataz  want to weigh in or happy?

jtniehof:
I think this has been open long enough for people to comment if they wish, and it's good to merge?

PR #410, opened 2020-08-21
toolbox: remove broken doctest

### LINKED ISSUE #414: toolbox.humansort gives wrong answer
The docstring claims:
```
    >>> import spacepy.toolbox as tb
    >>> dat = ['r1.txt', 'r10.txt', 'r2.txt']
    >>> dat.sort()
    >>> print(dat)
    ['r1.txt', 'r10.txt', 'r2.txt']
    >>> tb.human_sort(dat)
    ['r1.txt', 'r2.txt', 'r10.txt']
```

The answer actually given is:
```
In [7]:  tb.human_sort(dat)
Out[7]: ['r1.txt', 'r10.txt', 'r2.txt']
```

```
import spacepy.toolbox as tb
dat = ['r1.txt', 'r10.txt', 'r2.txt']
tb.human_sort(dat)
    ['r1.txt', 'r10.txt', 'r2.txt']
```

```
Linux-4.4.0-18362-Microsoft-x86_64-with-debian-bullseye-sid
sys.version_info(major=3, minor=7, micro=7, releaselevel='final', serial=0)
numpy=1.19.1
scipy=1.5.2
matplotlib=3.2.2
```

### Version of SpacePy
github master

balarsen:
Leftover doctest at the bottom of toolbox, hasn't worked since before 0.1.5. 

Also some code was checking for <numpy 1.3 which is a lot older than we support. 

None of the other modules have it, I propose removing it. 

(JTN: Closes #414)

### DIFF
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -174,9 +174,15 @@ def hypot(*args):
     >>> tot = 500
     >>> for num in tb.logspace(1, tot, 10):
     >>>     print num
-    >>>     num_list.append(timeit.timeit(stmt='tb.hypot(a)', setup='from spacepy import toolbox as tb; import numpy as np; a = [3]*{0}'.format(int(num)), number=10000))
-    >>>     num_np.append(timeit.timeit(stmt='tb.hypot(a)', setup='from spacepy import toolbox as tb; import numpy as np; a = np.asarray([3]*{0})'.format(int(num)), number=10000))
-    >>>     num_scalar.append(timeit.timeit(stmt='tb.hypot(*a)', setup='from spacepy import toolbox as tb; import numpy as np; a = [3]*{0}'.format(int(num)), number=10000))
+    >>>     num_list.append(timeit.timeit(stmt='tb.hypot(a)',
+                            setup='from spacepy import toolbox as tb;
+                            import numpy as np; a = [3]*{0}'.format(int(num)), number=10000))
+    >>>     num_np.append(timeit.timeit(stmt='tb.hypot(a)',
+                          setup='from spacepy import toolbox as tb;
+                          import numpy as np; a = np.asarray([3]*{0})'.format(int(num)), number=10000))
+    >>>     num_scalar.append(timeit.timeit(stmt='tb.hypot(*a)',
+                              setup='from spacepy import toolbox as tb;
+                              import numpy as np; a = [3]*{0}'.format(int(num)), number=10000))
     >>> from pylab import *
     >>> loglog(tb.logspace(1, tot, 10),  num_list, lw=2, label='list')
     >>> loglog(tb.logspace(1, tot, 10),  num_np, lw=2, label='numpy->ctypes')
@@ -189,8 +195,10 @@ def hypot(*args):
     .. image:: ../../source/images/hypot_no_extension_speeds_3cases.png
     """
     if lib.have_libspacepy:
-        if len(args) == 1 and hasattr(args, 'ndim'): # it is an array
-                ans = lib.hypot_tb(args[0], np.product(args[0].shape))
+        if len(args) == 1 and isinstance(args[0], np.ndarray):  # it is an array
+                # make sure everything is C-ready
+                ans = lib.hypot_tb(np.require(args[0], dtype=np.float64, requirements='C'),
+                                   np.product(args[0].shape))
                 return ans
     ans = 0.0
     for arg in args:
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -196,10 +196,10 @@ def hypot(*args):
     """
     if lib.have_libspacepy:
         if len(args) == 1 and isinstance(args[0], np.ndarray):  # it is an array
-                # make sure everything is C-ready
-                ans = lib.hypot_tb(np.require(args[0], dtype=np.float64, requirements='C'),
-                                   np.product(args[0].shape))
-                return ans
+            # make sure everything is C-ready
+            ans = lib.hypot_tb(np.require(args[0], dtype=np.float64, requirements='C'),
+                               np.product(args[0].shape))
+            return ans
     ans = 0.0
     for arg in args:
         if hasattr(arg, '__iter__'):
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -346,13 +346,9 @@ def tCommon(ts1, ts2, mask_only=True):
 
     tn1, tn2 = date2num(ts1), date2num(ts2)
 
-    v_test = np.__version__.split('.')
-    if v_test[0] == 1 and v_test[1] <= 3:
-        el1in2 = np.setmember1d(tn1, tn2) #makes mask of present/absent
-        el2in1 = np.setmember1d(tn2, tn1)
-    else:
-        el1in2 = np.in1d(tn1, tn2, assume_unique=True) #makes mask of present/absent
-        el2in1 = np.in1d(tn2, tn1, assume_unique=True)
+    el1in2 = np.in1d(tn1, tn2, assume_unique=True)  #makes mask of present/absent
+    el1in2 = np.in1d(tn1, tn2, assume_unique=True)  #makes mask of present/absent
+    el2in1 = np.in1d(tn2, tn1, assume_unique=True)
 
     if mask_only:
         return el1in2, el2in1
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -197,8 +197,9 @@ def hypot(*args):
     if lib.have_libspacepy:
         if len(args) == 1 and isinstance(args[0], np.ndarray):  # it is an array
             # make sure everything is C-ready
-            ans = lib.hypot_tb(np.require(args[0], dtype=np.float64, requirements='C'),
-                               np.product(args[0].shape))
+            ans = lib.hypot_tb(
+                np.require(args[0], dtype=np.double, requirements='C'),
+                np.product(args[0].shape))
             return ans
     ans = 0.0
     for arg in args:
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -455,7 +455,7 @@ def savepickle(fln, dict, compress=None):
         container with stuff
     compress : bool
         write as a gzip-compressed file
-                     (.gz will be added to L{fln}).
+                     (.gz will be added to ``fln``).
                      If not specified, defaults to uncompressed, unless the
                      compressed file exists and the uncompressed does not.
 
@@ -468,12 +468,9 @@ def savepickle(fln, dict, compress=None):
     >>> d = {'grade':[1,2,3], 'name':['Mary', 'John', 'Chris']}
     >>> savepickle('test.pbin', d)
     """
-    if compress == None:
-        # TODO, what is this line meant to do? Whay not just check fln for ending in gz?
-        if not os.path.exists(fln) and os.path.exists(fln + '.gz'):
-            compress = True
-        else:
-            compress = False
+    if compress == None: # Guess at compression
+        # Assume compressed if compressed already exists (and no uncompressed)
+        compress = not os.path.exists(fln) and os.path.exists(fln + '.gz')
     if compress:
         import gzip
         with open(fln + '.gz', 'wb') as fh:
@@ -483,7 +480,6 @@ def savepickle(fln, dict, compress=None):
     else:
         with open(fln, 'wb') as fh:
             pickle.dump(dict, fh, 2) # 2 ... fast binary
-    return
 
 
 # -----------------------------------------------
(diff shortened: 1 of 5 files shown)


spacepy/toolbox/__init__.py
```diff
@@ -1467,6 +1471,7 @@ def indsFromXrange(inxrange):
     if not isinstance(inxrange, xrange): return None
     valstr = inxrange.__str__()
     if ',' not in valstr:
+        # TODO, can't figure out in what case this code is run, every case is False then off to else.
```
jtniehof, line 1474: Probably means we're lacking a test, then. `spacepy.toolbox.indsFromXrange(xrange(5))`

tests/test_toolbox.py
```diff
@@ -165,6 +172,12 @@ def test_indsFromXrange(self):
         foo = xrange(23, 39)
         self.assertEqual([23, 39], tb.indsFromXrange(foo))
 
+    @unittest.expectedFailure
+    def test_indsFromXrange_zerolen(self):
+        """indsFromXrange claims useful when zero len range, seems not true"""
+        foo = xrange(20, 10)  # empty
+        self.assertEqual([23, 39], tb.indsFromXrange(foo))
```
jtniehof, line 179: Why do you expect this value? I get:
```python
In [5]: spacepy.toolbox.indsFromXrange(xrange(20, 10))
Out[5]: [20, 20]
```
which is eminently reasonable.

tests/test_toolbox.py
```diff
@@ -601,6 +614,19 @@ def test_hypot(self):
             self.assertEqual(tb.hypot(numpy.array([3.,4.])), 5.0)
             self.assertEqual(tb.hypot(numpy.array([3,4])), 5.0)
 
+    def test_hypot_array(self):
+        """hypot when called on an array falls to C"""
```
jtniehof, line 618: How do these tests verify that this falls through to C?

tests/test_toolbox.py
```diff
@@ -601,6 +614,19 @@ def test_hypot(self):
...
+        invals = [ numpy.asarray([3, 4]), numpy.arange(3,6), numpy.arange(3,10), numpy.asarray([-1,2,3]) ]
+        ans = [ 5, 7.0710678118654755, 16.73320053068151, 3.7416573867739413 ]
+        for i, tst in enumerate(invals):
+            self.assertAlmostEqual(ans[i], tb.hypot(*tst))
+        for i, tst in enumerate(invals):
+            self.assertAlmostEqual(ans[i], tb.hypot(tst))
+        self.assertEqual(5.0, tb.hypot(5.0))
+        if spacepy.lib.have_libspacepy:
```
jtniehof, line 626: These should still give the same value even if don't have libspacepy and want to make sure they work in that case...in fact, what we need is to do them twice (once faking have_libspacepy to force python; I can code this up if necessary).

tests/test_toolbox.py
```diff
@@ -601,6 +614,19 @@ def test_hypot(self):
...
+            self.assertEqual(tb.hypot(numpy.array([3.,4.])), 5.0)
+            self.assertEqual(tb.hypot(numpy.array([3,4])), 5.0)
```
jtniehof, line 628: It's not obvious what you're doing here, but presumably it's checking the fix with the require (maybe comment on that?). There should also probably be a test with a subclass (e.g. dmarray) and some tests with non-numpy inputs.

spacepy/toolbox/__init__.py
```diff
@@ -190,9 +195,11 @@ def hypot(*args):
     .. image:: ../../source/images/hypot_no_extension_speeds_3cases.png
     """
     if lib.have_libspacepy:
-        if len(args) == 1 and hasattr(args, 'ndim'): # it is an array
-                ans = lib.hypot_tb(args[0], np.product(args[0].shape))
-                return ans
+        if len(args) == 1 and isinstance(args[0], np.ndarray):  # it is an array
+            # make sure everything is C-ready
+            ans = lib.hypot_tb(np.require(args[0], dtype=np.float64, requirements='C'),
```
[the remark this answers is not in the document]
balarsen: It was just as nasty before, the `if` was totally wrong before. And it crashed on me when I did it without the `np.float64`, but I agree `np.double` is a better answer as the C is `double`

spacepy/toolbox/__init__.py
```diff
@@ -190,9 +195,11 @@ def hypot(*args):
     .. image:: ../../source/images/hypot_no_extension_speeds_3cases.png
     """
     if lib.have_libspacepy:
-        if len(args) == 1 and hasattr(args, 'ndim'): # it is an array
-                ans = lib.hypot_tb(args[0], np.product(args[0].shape))
-                return ans
+        if len(args) == 1 and isinstance(args[0], np.ndarray):  # it is an array
```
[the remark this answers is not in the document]
balarsen: Agreed, so you are in favor of this change? Or have a different line in mind? With the `np.require` just below this the input could be a list...

spacepy/toolbox/__init__.py
```diff
@@ -339,13 +346,9 @@ def tCommon(ts1, ts2, mask_only=True):
...
-    v_test = np.__version__.split('.')
-    if v_test[0] == 1 and v_test[1] <= 3:
-        el1in2 = np.setmember1d(tn1, tn2) #makes mask of present/absent
-        el2in1 = np.setmember1d(tn2, tn1)
-    else:
-        el1in2 = np.in1d(tn1, tn2, assume_unique=True) #makes mask of present/absent
-        el2in1 = np.in1d(tn2, tn1, assume_unique=True)
+    el1in2 = np.in1d(tn1, tn2, assume_unique=True)  #makes mask of present/absent
+    el1in2 = np.in1d(tn1, tn2, assume_unique=True)  #makes mask of present/absent
+    el2in1 = np.in1d(tn2, tn1, assume_unique=True)
```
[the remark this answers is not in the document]
balarsen: Added, the commit message for this commit was verbose "toolbox.tCommon: removed a check and fix for numpy <=1.3".

spacepy/toolbox/__init__.py
```diff
@@ -465,6 +468,7 @@ def savepickle(fln, dict, compress=None):
     >>> savepickle('test.pbin', d)
     """
     if compress == None:
+        # TODO, what is this line meant to do? Whay not just check fln for ending in gz?
```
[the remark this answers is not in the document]
balarsen: I believe that code is unreachable or misleading at best. The file is over written anyway so what does it matter if it existed with a gz...

spacepy/toolbox/__init__.py
balarsen: I can;t wait to drop xrange entirely... #26

tests/test_toolbox.py
balarsen: I get the useless
```
In [2]: import spacepy.toolbox as tb
In [3]: tb.indsFromXrange(range(20, 10))
Out[3]: [20, 10]
```

out of python3, maybe it should error with `NotImplementedError` or something in py3.

tests/test_toolbox.py
balarsen: There is no way to verify it that I know of, but there we no tests that could go to C as there were no arrays.

tests/test_toolbox.py
balarsen: That would be good to do. Not really sure we need the C hypot but it is a good learning location. And it was wrong.

tests/test_toolbox.py
balarsen: This checks the `np.require` is doing its job and not passing the int thru to the C.

spacepy/toolbox/__init__.py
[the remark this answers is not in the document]
jtniehof: No, not that the fix is nasty, what you caught is nasty :)

spacepy/toolbox/__init__.py
jtniehof: But the new `range` is just old `xrange` so this function will stay around.

tests/test_toolbox.py
jtniehof: We should check that then; assuming the py2k result is "right" then fix on py3k.

tests/test_toolbox.py
jtniehof: I read the docstring as "test that hypot falls to C" and that's not really what it's testing..."hypot on array, may fall to C" perhaps?

spacepy/toolbox/__init__.py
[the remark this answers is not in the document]
jtniehof: Yes, I think the explicit type check is probably necessary, as much as I like to avoid it.

It does seem a bit weird that we're testing if it's an array, and then casting to an array if it isn't, but then not using the fast array code. Of course if we're looking at patching up this implementation we also need to evaluate in the light of `numpy.linalg.norm`...I suggest punting hypot enhancements (beyond this fix and the further tests for it) to another PR/issue.

spacepy/toolbox/__init__.py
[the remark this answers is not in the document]
jtniehof: No, it's not unreachable. If you call `savepickle('foo', data, compress=False)` it writes to `foo`. If you call `savepickle('foo', data, compress=True)` it writes to `foo.gz`. if you call `savepickle('foo', data)` it will write to `foo` unless `foo` does not exist and `foo.gz` does, in which case it writes to `foo.gz`. As the docstring says, "write as a gzip-compressed file (.gz will be added to L{fln}). If not specified, defaults to uncompressed, unless the compressed file exists and the uncompressed does not." Ignore the epydoc, please.

We don't have an explicit _test_ of the case where the compressed version exists and `compress` is not specified.

tests/test_toolbox.py
jtniehof: This whole function is primarily a copy of test_hypot; really only the fop set of invals changes (to be numpy arrays), and there was already a test with numpy arrays at the bottom of test_hypot. (Which, until you fixed, didn't actually fall to C, so it didn't show that bug, but that's something that I guess would probably be caught with coverage.) What we need to do is test with those inputs both in C and Python....you want me to just do that work?

tests/test_toolbox.py
jtniehof: Still not sure why you expect [23, 39] though.

jtniehof: I'm honestly not sure about the utility of this function; it doesn't work with a specified stride, for instance. There's a definite difference between the py2k and py3k behaviour in the way the `str` is rendered; my inclination is let's just slap this in as a regression test, have separate expected for 2 and 3, and worry about the defined correct behaviour later.

jtniehof: This was added by @drsteve in f725534f1d5a256b52d648ae7e5d6f8d541763fb to support omni. Basically it was using tOverlapHalf to find a start/end index, but since that was returned as an xrange, then omni wanted to derive the indices from it, and since it wanted to extend past the start/finish, it couldn't just cast to a list and get an empty list. The situation it was written for was only for things of form xrange(20, 20) (from which it would go grabbing 19:22), never with the second one smaller than the first.

Let's just document this as undefined if end < start or if there's a stride and call it good.

jtniehof:
@balarsen , what's your thought on this? You want to take a thwack at some of my comments or should I make a commit with some of those ideas implemented?

balarsen:
@jtniehof, I lost lock on this for a bit having to deal with budgets. I am happy to have you do commits if that works for you.

jtniehof:
Should we prioritize for 0.2.2? I'd like to push and get that out and if I can reasonably leave stuff off the list, it helps.

balarsen:
I’m thinking on stopping this one and just hitting some of the small bits in separate PRs. Thoughts? The hypot fix should be done. The rest can wait a bit.

jtniehof:
Nah, if we want some of it in 0.2.2 doing the whole thing isn't that much more work, so I added it to the pile.

tests/test_toolbox.py
balarsen: > Still not sure why you expect [23, 39] though.

Yeah I have no idea either...

jtniehof:
I'm going to be making commits to your branch and marking conversations resolved as I go there.

spacepy/toolbox/__init__.py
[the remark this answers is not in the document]
jtniehof: I think the decision on this was it's fine and punt any further hypot optimization to later.

jtniehof:
I'm good to merge this when done. I fixed #414 while I was at it...guess who broke it? :) Also added a test for the case that gave rise to the change that broke it.

@balarsen , once you're good with my changes I think you should also approve and somebody else should merge (@drsteve?) now that my fingerprints are on it.

jtniehof:
I had to rebase and force-push since there were conflicts.

jtniehof (approved):

balarsen (commented):
Thanks @jtniehof for finishing this off. As this is "my" pull request I can't approve. But I do approve.

jtniehof:
Since @balarsen and I both approve (and both have our fingerprints on this), unless @drsteve complains very soon I'm going to merge.

drsteve (approved):
All looks straightforward to me. Did the hypot questions get elevated to an issue?

Happy to merge now.

PR #417, opened 2020-09-14
Support Ticktock to/from astropy.time.Time (Closes #321)

### LINKED ISSUE #321: Ticktock should convert to/from AstroPy time

jtniehof:
This PR adds support for the AstroPy time object. It can be passed to the Ticktock constructor, and a new .getAPT() method and .APT attribute will return AstroPy time.

EDIT: I keep on forgetting the title doesn't work for this. Closes #321 

A few things to note:

1. AstroPy is added as a dependency in the documentation and the CI, but it is _not_ added to setup.py or requires.txt. Installing AstroPy adds no functionality to SpacePy; it's only of interest if you want to work with AstroPy. In which case, chances are you have AstroPy and don't need SpacePy to ask pip to install it.
2. Some combinations of Python 3 version + AstroPy version will throw a ResourceWarning from the tests. This is an AstroPy problem; merely importing AstroPy 1.1.1 on Python 3.5.2 with warnings set to 'always' will make this happen. Okay on AstroPy 3.0 on Python 3.6.9. This will reproduce:
    ```python
    import warnings
    warnings.simplefilter('always', category=ResourceWarning)
    import astropy
    ```
3. In the docstring for Ticktock it now mentions that `.APT` require AstroPy 1.0. This is the only case I know of where we mention dependencies in the docstring. On the one hand it seems reasonable to mention it to avoid surprise; on the other hand, we don't list everything that's required for every single function in SpacePy. Options here are to leave as-is, just say "AstroPy" without version, or assume that if you're looking for AstroPy objects the requirement for AstroPy is assumed.

## PR Checklist

### DIFF
--- a/spacepy/time.py
+++ b/spacepy/time.py
@@ -200,8 +200,7 @@ class Ticktock(MutableSequence):
     CDF
         CDF Epoch type: float milliseconds since 0000-1-1 ignoring leapseconds
     APT
-        AstroPy :class:`~astropy.time.Time`. Ticktock objects using AstroPy
-        time as input may not pickle properly. Requires AstroPy 1.0.
+        AstroPy :class:`~astropy.time.Time`. Requires AstroPy 1.0.
         (New in version 0.2.2.)
 
     Possible output data types: All those listed above, plus:
--- a/spacepy/time.py
+++ b/spacepy/time.py
@@ -201,7 +201,7 @@ class Ticktock(MutableSequence):
         CDF Epoch type: float milliseconds since 0000-1-1 ignoring leapseconds
     APT
         AstroPy :class:`~astropy.time.Time`. Ticktock objects using AstroPy
-        time as input may not pickle properly. Requires AstroPy 0.3.
+        time as input may not pickle properly. Requires AstroPy 1.0.
         (New in version 0.2.2.)
 
     Possible output data types: All those listed above, plus:
@@ -1460,7 +1460,7 @@ class Ticktock(MutableSequence):
         =====
         .. versionadded:: 0.2.2
 
-        Requires AstroPy 0.3.
+        Requires AstroPy 1.0.
 
         Examples
         ========
--- a/spacepy/time.py
+++ b/spacepy/time.py
@@ -155,6 +155,11 @@ import os.path
 import re
 import warnings
 
+try:
+    import astropy.time
+    HAVE_ASTROPY = True
+except ImportError:
+    HAVE_ASTROPY = False
 import dateutil.parser as dup
 import numpy as np
 
@@ -173,7 +178,7 @@ class Ticktock(MutableSequence):
     Ticktock( data, dtype )
 
     Ticktock class holding various time coordinate systems
-    (TAI, UTC, ISO, JD, MJD, UNX, RDT, CDF, DOY, eDOY)
+    (TAI, UTC, ISO, JD, MJD, UNX, RDT, CDF, DOY, eDOY, APT)
 
     Possible input data types:
 
@@ -194,6 +199,10 @@ class Ticktock(MutableSequence):
         Rata Die days elapsed since 0001-1-1
     CDF
         CDF Epoch type: float milliseconds since 0000-1-1 ignoring leapseconds
+    APT
+        AstroPy :class:`~astropy.time.Time`. Ticktock objects using AstroPy
+        time as input may not pickle properly. Requires AstroPy 0.3.
+        (New in version 0.2.2.)
 
     Possible output data types: All those listed above, plus:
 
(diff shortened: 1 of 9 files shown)


balarsen (dismissed):
Reading this code it looks good. I did not download and run it.

jtniehof:
I've updated against latest master. @balarsen , if you're still good with this, nobody else has commented so I think a merge is okay. (In theory the tests make sure it works ;) ) I'll reference @drsteve so he has a chance to object.

balarsen (dismissed):
Looking good by me

spacepy/time.py
```diff
@@ -1409,6 +1438,52 @@ def getGPS(self):
...
+        """
+        if self.data.attrs['dtype'] == 'APT':
+            # This should be the case from the constructor
+            self.APT = self.data
+            return self.APT
+        if not HAVE_ASTROPY:
+            raise RuntimeError('Import of astropy.time failed.')
+        GPS0 = 694656019
+        self.APT = astropy.time.Time(
+            self.TAI - GPS0, scale='tai', format='gps')
```
drsteve, line 1483: Do we want to output on the `tai` scale? Or the `utc` scale?
I was somewhat confused when I tried this out initially and printed an ISO time that was significantly different to the input I used. That, of course, was fixed when I changed the _scale_ to UTC. I believe that GPS time is the same on each scale (where the ISO time, JD, etc.) are not. For a user, it might be more intuitive to output an `astropy.time.Time` object on the `utc` scale...
Related question, do we still want to do this as `self.TAI - GPS0` or just `self.GPS`? They should be the same, right?

jtniehof: I agree that seeing an ISO datestamp for a TAI time is confusing to space physicists :)

The constructor for Time takes the input scale, not the output scale. The AstroPy docs claim that GPS time is intrinsically on the TAI scale. I've checked and it's possible to specify the input scale as `utc` and format as `gps`, and it seems to give equivalent answers:
```python
t = astropy.time.Time(5e8, format='gps', scale='tai')
t = t.utc
t.format='iso'
print(t) #1995-11-10 00:53:10.000
t2 = astropy.time.Time(5e8, format='gps', scale='utc')
t2.format='iso'
print(t2) #1995-11-10 00:53:10.000
t3 = spacepy.time.Ticktock(5e8, dtype='GPS')
print(t3.iSO[0]) #1995-11-10T00:53:10
```
However lacking a really good definition of what AstroPy considers to be the `utc` "scale" makes me nervous. E.g. they claim the unix format as being on the `utc` scale, which would involve pretending leapseconds are moments when time doesn't exist.

We could be sure by, for instance, doing:
```python
self.APT = astropy.time.Time(self.TAI - GPS0, scale='tai', format='gps').utc
```
but making two Time objects seems nasty, and I'm nervous about potentially losing leapseconds there. (It probably _won't_, mind you, just nervous...) Doing `gps` format, `tai` scale seemed the best way to make sure the information in the Time object 100% corresponded to the Ticktock input, and presumably a user working with AstroPy Time can perform manipulations to their desired output format (and scale).

`self.TAI - GPS0` should be the same as `self.GPS`, but by doing the conversion explicitly from TAI, I'm avoiding some possibility of loops doing the lookup. The interactions between getXXX, getattr, init, and update_items are tricky. There's special sauce that prioritizes doing TAI first and (to a lesser extent) UTC second, and everything else coming from one of those. Elevating GPS to that level seemed to invite complexity to save copy/pasting a subtraction.

drsteve:
Point taken on the `self.TAI - GPS0` versus `self.GPS`.

On the time scale, the astropy documentation is sparse in key places where I'd like to see more about their scale definitions and the conversions. However, while I believe that GPS time should be identical in either scale, they do explicitly say that GPS is intrinsically on the TAI scale. So from that perspective, it's probably a good idea to explicitly use that scale.

Given that I think we should probably note in the `getAPT` docstring that it produces an `astropy.time.Time` with the `tai` scale and `gps` format, and that the scale can be changed (see astropy docs).

Related, the example in the docstring is:
```
        >>> a = Ticktock('2002-02-02T12:00:00', 'ISO')
        >>> a.APT
        <Time object: scale='utc' format='gps' value=696686413.0>
```
Which shows the scale as `utc`, but that's not what comes out.

I think if we get the docs more explicit and consistent then I'm happy.

jtniehof:
Ew, that example was flat-out wrong. Added some details on the output as well.

drsteve (approved):

PR #420, opened 2020-09-17
Update dependency version documentation

### LINKED ISSUE #365: Many functions require numpy after 1.6
#355 uses `numpy.matmul` which was added in 1.10 (in my defense, that information was _buried_.) So that means 0.2.2 will essentially require 1.10 unless I cut that functionality out. It's probably work-around-able but not much fun. I'd like to avoid doing a subminor release that bumps dependency requirements, so I'll make a workaround...I don't care if it's slow and cruddy, it's a workaround.

See also #362.

jtniehof:
This PR updates the documentation of what dependencies are required and thus closes #365 . In the process I also removed a few workarounds for earlier versions of dependencies which are no longer supported (I didn't do an exhaustive search, just grabbed the quick ones I remembered) and updated the table of how we choose what dependencies to support.

## PR Checklist

### DIFF
--- a/spacepy/coordinates.py
+++ b/spacepy/coordinates.py
@@ -728,23 +728,13 @@ def quaternionFromMatrix(matrix, scalarPos='last'):
                            np.identity(3), atol=0.2):
             raise ValueError('Input rotation matrix{} not orthogonal.'.format(
                 '' if len(matrix.shape) == 2 else ' at ' + str(i)))
-        try:
-            det = np.linalg.det(m)
-        except AttributeError: # det new in numpy 1.8.0
-            det = m[0, 0] * m[1, 1] * m[2, 2] \
-                  + m[0, 1] * m[1, 2] * m[2, 0] \
-                  + m[0, 2] * m[1, 0] * m[2, 1] \
-                  - m[0, 2] * m[1, 1] * m[2, 0] \
-                  - m[0, 1] * m[1, 0] * m[2, 2] \
-                  - m[0, 0] * m[1, 2] * m[2, 1]
+        det = np.linalg.det(m)
         if det < 0:
             raise ValueError('Input rotation matrix at {} not proper.'
                              .format(str(i)))
     inshape = matrix.shape
     # Flatten out most dimensions, easier to index
     matrix = np.reshape(matrix, (-1, 3, 3))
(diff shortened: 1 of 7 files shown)


spacepy/coordinates.py
```diff
@@ -728,23 +728,13 @@ def quaternionFromMatrix(matrix, scalarPos='last'):
...
-        try:
-            det = np.linalg.det(m)
-        except AttributeError: # det new in numpy 1.8.0
-            det = m[0, 0] * m[1, 1] * m[2, 2] \
-                  + m[0, 1] * m[1, 2] * m[2, 0] \
-                  + m[0, 2] * m[1, 0] * m[2, 1] \
-                  - m[0, 2] * m[1, 1] * m[2, 0] \
-                  - m[0, 1] * m[1, 0] * m[2, 2] \
-                  - m[0, 0] * m[1, 2] * m[2, 1]
+        det = np.linalg.det(m)
```
balarsen, line 731: Excellent catch here!

tests/test_time.py
```diff
@@ -1358,8 +1358,7 @@ def test_UTCGPS(self):
         """testing get UTC from GPS"""
         t1 = t.Ticktock([6.93882013e+08, 6.93964813e+08], 'GPS')
         expected = t.Ticktock(['2002-01-01T01:00:00', '2002-01-02'])
-        #        numpy.testing.assert_array_equal(t1, expected)
-        self.assertTrue((t1 == expected).all())
+        numpy.testing.assert_array_equal(t1, expected)
```
balarsen, line 1361: This likely needs to be ` np.testing.assert_array_almost_equal` as these are possibly floats.

balarsen (changes requested):
I;m happy with  the one comment that needs looking at.

spacepy/coordinates.py
jtniehof: That was an easy one since I remember writing it pretty recently :)

tests/test_time.py
jtniehof: They're float type, but integral valued, and thus the calculation should carry through exactly with no loss of precision. If that ever didn't happen I would want the test to fail.

balarsen (approved):

PR #428, opened 2020-10-02
Fix handling of OMNI databases without UTC

jtniehof:
fe42d8a4e138e7a4a2cc12fabdd1d301b98d499e switched away from using RDT to determine the start/end time of an omni database, in favor of UTC. Unfortunately UTC is not stored in the OMNI2 HDF5 files, but generated on read. This PR tries (in order) UTC, Epoch/ISO string (for OMNI2), and RDT (as a fallback.)

## PR Checklist
- [X] (N/A, see below) New features and bug fixes should have unit tests

This isn't possible to unit test since we don't require a downloaded current OMNI2 database for unit tests. However, without the fix integration_omni.py should fail (and succeed with.)

### DIFF
--- a/spacepy/omni.py
+++ b/spacepy/omni.py
@@ -374,8 +374,17 @@ def omnirange(dbase='QDhourly'):
               'Test': testfln}
     if dbase not in infile:
         raise NotImplementedError('')
+    # Possible time variables in the HDF file and their ticktock dtype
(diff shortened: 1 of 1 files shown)


drsteve (approved):
Fixes the issue here, code looks fine.
Tested with `integration_omni.py` (failed before, worked after) as well as interactive testing in interpreter.

PR #430, opened 2020-10-26
toolbox.bootHisto: use RNG seed; make test deterministic.

jtniehof:
`bootHisto` accepts a `seed` argument to specify a seed for the RNG. Unfortunately this was not properly passed through to the underlying bootstrap function, which made `bootHisto` nondeterministic even on a single machine with a constant seed. The tests used this functionality to avoid random failures; since this didn't work, there was a cron job CI failure on master.

This PR passes through the `seed` argument properly. This results in a slight shift in the expected value in the tests, since the seed  is now being applied in the test.

## PR Checklist

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -69,6 +69,7 @@ toolbox
  - New functions quaternionFromMatrix/quaternionToMatrix to convert between
    quaternions and 3D rotation matrices.
  - Fix human_sort; this had been broken since 0.1.6.
+ - Fix function bootHisto to pass through RNG seed if provided
 
 Changes in Version 0.2.1 (2019-10-02)
 =====================================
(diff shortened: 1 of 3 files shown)


jtniehof:
[Failed build](https://travis-ci.com/github/spacepy/spacepy/builds/191548699)...the [one before](https://travis-ci.com/github/spacepy/spacepy/builds/191286562) and [one after](https://travis-ci.com/github/spacepy/spacepy/builds/191771288) suceeded despite being on the same commit, so it's a Heisenbug.

balarsen (approved):
Yep sets the random seed.

jtniehof:
@balarsen , I'm good with merging unless @drsteve wants to chime in...I had only requested the review from him since he's the other person who's worked on the bootstrap code.

drsteve:
What changed on the force push? Or was it just because of a rebase against master?
If so, any guesses on the "Heisenbug"?

jtniehof:
It was to handle the conflict rebase against master.

The Heisenbug arose because the seed wasn't being passed through, thus the test was not deterministic, and we happened to hit the rare case where the selected surrogates gave results that were outside the tolerance of the test. This PR fixes that, so the test is now deterministic. There is still "slop" (i.e. tolerance greater than zero) in the tests because the RNG is not always identical across platforms, but if the test passes once for a particular machine, version of Python, version of numpy, it should always pass (and, similarly, always fail.)

drsteve (approved):

Issue #173, opened 2019-06-22
pip build system support


jtniehof:
Some of the not-installing-without-numpy issues might relate to the fact that pip support for build requirements is very new (and not exploited in SpacePy). So we may want to add PEP 517 and 518 support as described [here](https://pip.pypa.io/en/stable/reference/pip/#build-system-interface).

This probably relates to #45. I don't think this will be a quick "one and done"; the installer is currently groaning under the weight of all the nonstandard stuff we have to do and the "modern" setup environment (pip + wheel + setuptools) really doesn't like anything nonstandard (very much our way or the highway.) So this is something to do after reducing the overall hackiness of `setup.py`, potentially after #112 .

PR #440, opened 2020-11-08
use pep517/518 pyproject.toml to ensure numpy installed before setup.py

scivision:
this avoids install errors for missing numpy, whether installing spacepy by itself or specifying spacepy in "install_requires" of other projects.
This is necessary in any project using f2py or where numpy is used during execution of setup.py.

## PR Checklist
Please try to follow as much of this checklist as you can for
Some additional suggestions are given below.

Go ahead and replace the text above (and this line!) with your
pull request description.

- [ X] Pull request has descriptive title
- [X ] Pull request gives overview of changes
- [ /] New code has inline comments where necessary
- [/ ] Any new modules, functions or classes have docstrings consistent with SpacePy style
- [/ ] Added an entry to CHANGELOG if fixing a major bug or providing a major new feature
- [ /] New features and bug fixes should have unit tests
- [ /] Relevant issues are linked to (e.g. `See issue #` or `Closes #`)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -9,6 +9,7 @@ toolbox
  - Fix bootHisto passing of kwargs to histogram and bar.
 pycdf
  - Add support for Sparse Records variables.
+ - Fix changing compression when creating variable from existing var/data.
 
 Changes in Version 0.2.2 (2020-12-29)
 =====================================
--- a/spacepy/pycdf/__init__.py
+++ b/spacepy/pycdf/__init__.py
@@ -2359,8 +2359,9 @@ class CDF(MutableMapping, spacepy.datamodel.MetaMixin):
             else:
                 if compress is None:
                     compress = c
-                if compress_param is None:
-                    compress_param = cp
+                # Ignore data's compress_param if using compress argument
+                    if compress_param is None:
+                        compress_param = cp
         if hasattr(data, 'sparse') and sparse is None:
             sparse = data.sparse()
         if hasattr(data, 'pad') and pad is None:
--- a/tests/test_pycdf.py
+++ b/tests/test_pycdf.py
@@ -2655,6 +2655,15 @@ class ChangeCDF(ChangeCDFBase):
                          comptype.value)
         self.assertEqual(8, compparam)
 
+    def testNewVarFromVarCompress(self):
+        """Create a new variable from a variable, change compression"""
+        zvar = self.cdf.new('newzvar1', compress=const.GZIP_COMPRESSION,
+                            data=self.cdf['SpinRateScalersCounts'])
+        comptype, compparam = zvar.compress()
+        self.assertEqual(const.GZIP_COMPRESSION.value,
+                         comptype.value)
+        self.assertEqual(5, compparam)
+
     def testNewVarFromdmarrayAssign(self):
         """Create a new variable by assigning from dmarray"""
         indata = datamodel.dmarray([1,2,3], dtype=numpy.int8,


balarsen:
Really cool. I didn’t know about. For others here is what seems like a decent write up on this technique. https://snarky.ca/what-the-heck-is-pyproject-toml/

scivision:
right, the old approach was using setuptools.setup(setup_requires) but that fails if the install is offline, even if the module is already installed. Plus the factors noted in that article.

jtniehof:
Thanks for the PR. Once ready to merge this closes #173. As noted there, the pip support for this is pretty new. I'm okay with it not auto-installing numpy on an older pip as long as the install isn't completely broken on older pip (i.e. the metadata version comes out sufficiently old.) So I'll need to check that and also the ramifications of not having [setuptools injection](https://pip.pypa.io/en/stable/reference/pip/#setuptools-injection).

scivision:
OK I see there is spacepy/setup.py logic using distutils. In Meson, which works on a broad range of platforms with system python, we removed distutils once the deprecation of distutils was announced. 
I mention that because scrubbing distutils hasn't been a problem for Meson with a vocal audience and many operating systems and archs.

scivision:
Also Meson happened to add pyproject.toml before we removed distutils some time later.

jtniehof:
Here's my checklist; I'll work it off:

- [ ] Build a wheel with a modern pip and make sure it works okay without setuptools injection, or figure out a way to detect pip and force injection
- [ ] Install that wheel with modern pip just to be sure :)
- [ ] Install that wheel on an older pip and make sure it works (although it's not going to magically get numpy for us, of course)
- [ ] Build a wheel with an older pip and install it with a newer pip; this is another check for setuptools injection and also if the old-build will result in something that a PEP518-aware pip will do the numpy build for (would be nice)

The problem with just dropping distutils is that setuptools has some really nasty behavior in multi-user managed environments that I'd like to avoid if possible.

jtniehof:
Incidentally since the PR is out of master that could cause weird things in the future.

jtniehof:
`pip install .` fails on pip 20.3.4 with this PR but succeeds without it. Basically it doesn't install any of the dependencies and throws the error "pip's dependency resolver does not currently take into account all the packages that are installed." Either the presence of `pyproject.toml` triggers the new resolver and the old one happens otherwise, or something about removing the setuptools injection when the `pyproject.toml` is present is causing the problem. I'll have to play around a bit with restoring the injection and seeing if that works. Either way, I definitely will test install straight from a commit hash, since that's probably closer to what will happen when installing from pypi. `pip install .` is so weird I don't care if it's a bit off but I'm concerned about the implications for other means of install.

jtniehof:
I found the issue, PEP 517/518 use different command line args for getting the info. I've got a potential fix but I need to run the range of tests.

@scivision , because the PR was made out of your `master` branch I'm reluctant to push to it. I'll probably have to open another PR to have it on a branch. Otherwise the rebase will be a pain for you ...you'll have a conflict trying to update your fork. That's going to happen regardless (since the one commit you have on `master` will conflict once we rebase.)

New checklist!

- [ ] Build a wheel with a modern pip (10.0 or later) and make sure it works okay without setuptools injection, or figure out a way to detect pip and force injection (`pip wheel .`)
- [ ] Install that wheel with modern pip just to be sure :)
- [ ] Install that wheel on an older pip (before 10.0) and make sure it works (although it's not going to magically get numpy for us, of course)
- [ ] Build a wheel with an older pip and install it with a newer pip; this is another check for setuptools injection and also if the old-build will result in something that a PEP518-aware pip will do the numpy build for (would be nice)
- [ ] Do a direct pip install without first building the wheel on new pip (`pip install .`)
- [ ] Do a direct pip install without first building the wheel on old pip
- [ ] pip install from a github branch, new pip
- [ ] pip install from a github branch, old pip

scivision:
It's fine for me re: master unless it's a bother for you. Thanks

jtniehof:
Oh, cute, since it's against the `master` branch it auto-closed on push. Let me try to get it reopened....

jtniehof:
Got the wires crossed on pushing to a branch with a different name and so this is up to date with `master`, assumed merged, and closed, and so I can't reopen it, and can't push to @scivision 's branch because it's no longer on an open PR. I'm just going to open another PR.

scivision:
Okay sorry for the difficulty thanks again.

jtniehof:
> Okay sorry for the difficulty thanks again.

No problem, thanks for submitting this!

PR #459, opened 2020-12-18
Correct MLT labeling in pybats.ram.RamSat flux plots

drsteve:
In `spacepy.pybats.ram.RamSat` the `add_omniflux_plot` method produces MLT labels that are offset by 12 hours (i.e. offset of pi radians in the azimuthal coordinate.

The issue was found when plotting simulation output and comparing to Van Allen Probes. It can be reproduced with the test file though.
Inspecting the test file `tests/data/pybats_test/ramsat_test.nc` the approximate SM position is [-1.78, -4.58, -1.17]. Since X is positive sunward and Y is positive duskward, this should imply an MLT in the morning. Calling `_orbit_formatter` yielded an MLT label of `16:35 MLT`.

The new fix:
1. removes the custom MLT calculation and calls `spacepy.toolbox.rad2mlt`.
2. adds a test to check that the output string for plot labeling is as expected given the test file

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -48,6 +48,7 @@ pybats
  - Bats2d.add_b_magsphere and add_stream_scatter now have kwargs for adding arrows on ray traces
  - Updates to ram.RamSat including omnidirectional flux calculation
  - ram.PressureFile fixed for Python 3.x failures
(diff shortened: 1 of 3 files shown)


spacecataz (approved):
Crisp solution; glad to see the tie-in with toolbox functions.

PR #475, opened 2021-01-06
empiricals.py: getPlasmaPause() raising a warning rather than a print

### LINKED ISSUE #474: spacepy.empiricals .getPlasmaPause prints a warning not raise/error
`spacepy.empiricals.getPlasmaPause` prints a warning, should either raise a warning or raise a `ValueError`

See:

https://github.com/spacepy/spacepy/blob/a4c0ca6895dc2c419c9cd0a71d1d85b62f7f89db/spacepy/empiricals.py#L124

import spacepy.time as spt
import spacepy.empiricals as emp
ticks = spt.tickrange('2002-01-01T12:00:00','2002-01-04T00:00:00',.25)
emp.getPlasmaPause(ticks, model='CA1992', LT=12)

### Error message/Traceback:
```
array([3.76      , 3.76      , 4.35799998, 4.35799998, 4.35799998,
       4.35799998, 4.35799998, 4.35799998, 4.35799998, 4.54200002,
       5.14      ])
```

```
Linux-4.4.0-18362-Microsoft-x86_64-with-glibc2.10
sys.version_info(major=3, minor=8, micro=5, releaselevel='final', serial=0)
numpy=1.19.4
scipy=1.6.0
matplotlib=3.3.3
```

### Version of SpacePy
76c5a35c688e7876292ce05f3bf833a793d77fd6

balarsen:
`getPlasmaPause` now raises `RuntimeWarning`, closes #474 

## PR Checklist

Unit test added, minor enough did not modify CHANGELOG

### DIFF
--- a/spacepy/empiricals.py
+++ b/spacepy/empiricals.py
@@ -123,7 +123,7 @@ def getPlasmaPause(ticks, model='M2002', LT='all', omnivals=None):
 
     if model == 'CA1992':
         if LT != 'all':
-            warnings.warn(RuntimeWarning('No LT dependence currently supported for CA1992 model'))
+            warnings.warn('No LT dependence currently supported for CA1992 model', RuntimeWarning)
     if model not in model_list:
         raise ValueError("Please specify a valid model:\n{0}".format(' or '.join(model_list)))
 
--- a/spacepy/empiricals.py
+++ b/spacepy/empiricals.py
@@ -75,12 +75,12 @@ def getPlasmaPause(ticks, model='M2002', LT='all', omnivals=None):
     """
     Plasmapause location model(s)
 
-    CA1992 -- Carpenter, D. L., and R. R. Anderson, An ISEE/whistler 
-    model of equatorial electron density in the magnetosphere, 
+    CA1992 -- Carpenter, D. L., and R. R. Anderson, An ISEE/whistler
+    model of equatorial electron density in the magnetosphere,
     J. Geophys. Res., 97, 1097, 1992.
-    M2002 -- Moldwin, M. B., L. Downward, H. K. Rassoul, R. Amin, 
-    and R. R. Anderson, A new model of the location of the plasmapause: 
-    CRRES results, J. Geophys. Res., 107(A11), 1339, 
+    M2002 -- Moldwin, M. B., L. Downward, H. K. Rassoul, R. Amin,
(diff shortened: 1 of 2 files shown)


jtniehof:
I still feel weird about making the user trap a warning for something that might be perfectly reasonable to do, and which there's no way of them finding out about it other than tripping it. I know "nobody ever reads the docs" but without reading the docs they wouldn't even know to pass in CA1992 in the first place....

A quick skim of Carpenter and Anderson suggests they do have some L dependence, and it's just we haven't implemented it. I think the docs really could use a mention of exactly which we've implemented and warn on 'all' and specific "bad" LTs. It looks like what's implemented is the plasmapause inner limit for 00-06 LT, so we should warn for LT outside of there. For that a warning makes sense: it's not returning what's actually in the model.

And then I'll open an enhancement issue to implement the MLT dependence for 06-15, which is easy. 15-24 might be harder (I'd have to actually, like, understand it.)

spacepy/empiricals.py
```diff
@@ -120,8 +122,8 @@ def calcLpp(Kpmax, A, B, power=1):
     model_list = ['CA1992', 'M2002', 'RT1970']
 
     if model == 'CA1992':
-        if LT!='all':
-            print('No LT dependence currently supported for this model')
+        if LT != 'all':
+            warnings.warn(RuntimeWarning('No LT dependence currently supported for CA1992 model'))
```
jtniehof, line 126:
```suggestion
            warnings.warn('No LT dependence currently supported for CA1992 model', RuntimeWarning)
```

This is more typical usage

balarsen: Ok, I will change that. I have never done them that way, I have always matched how errors are called e.g. `raise TypeError('windowmean error: winsize/overlap must be timedeltas')`

spacepy doesn't have a lot of warnings but the ones it does match your style.

balarsen:
> EDIT: I think a reasonable minimal doc update would be to add the "warns" section. https://numpydoc.readthedocs.io/en/latest/format.html#sections

I added this and ran the docs, looks good.

jtniehof (approved):
Looks good to me.

jtniehof:
Any reason I shouldn't open up the enhancement issue?

balarsen:
> 
> 
> Any reason I shouldn't open up the enhancement issue?

Good by me, unlikely to be done rapidly but would add some value.

PR #480, opened 2021-01-08
toolbox: documentation update for numpydoc formatting and added Raises

balarsen:

Not really a major change but I feel like a documentation entry in the CHANGELOG is appropriate, if for no other reason than as a reminder.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,8 @@
 Changes in Version 0.2.3 (2021-xx-xx)
 =====================================
+documentation
+ - toolbox cleaned for numpydoc format and Raises sections added
+
 
 Changes in Version 0.2.2 (2020-12-29)
 =====================================
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -101,22 +101,22 @@ def unique_columns(inval, axis=0):
     axis=0 is unique rows, axis=1 is unique columns
 
     Parameters
-    ==========
+    ----------
     inval :  array-like
         array to find unique columns or rows of
 
     Optional Parameters
-    ===================
+    -------------------
     axis : int
         The axis to find unique over, default: 0
 
     Returns
-    =======
+    -------
     out : array
         N-dimensional array of the unique values along the axis
 
     Examples
-    ========
+    --------
     """
     # this is a nice trick taking advantage of structed arrays where each row or column
     #   is the value, so returl unique works
@@ -135,17 +135,17 @@ def hypot(*args):
     compute the N-dimensional hypot of an iterable or many arguments
 
     Parameters
-    ==========
+    ----------
     args : many numbers or array-like
         array like or many inputs to compute from
 
     Returns
-    =======
+    -------
     out : float
         N-dimensional hypot of a number
 
     Notes
-    =====
+    -----
     This function has a complicated speed function.
      - if a numpy array of floats is input this is passed off to C
      - if iterables are passed in they are made into numpy arrays and comptaton is done local
@@ -157,7 +157,7 @@ def hypot(*args):
      - >20 elements premake them into a numpy array of doubles
 
     Examples
-    ========
+    --------
     >>> from spacepy import toolbox as tb
     >>> print tb.hypot([3,4])
     5.0
@@ -215,7 +215,7 @@ def tOverlap(ts1, ts2, *args, **kwargs):
     Finds the overlapping elements in two lists of datetime objects
 
     Parameters
-    ==========
+    ----------
     ts1 : datetime
         first set of datetime object
     ts2 : datetime
@@ -224,12 +224,12 @@ def tOverlap(ts1, ts2, *args, **kwargs):
         additional arguments passed to tOverlapHalf
 
     Returns
-    =======
+    -------
     out : list
         indices of ts1 within interval of ts2, & vice versa
 
     Examples
-    ========
+    --------
     Given two series of datetime objects, event_dates and omni['Time']:
 
     >>> import spacepy.toolbox as tb
@@ -245,7 +245,7 @@ def tOverlap(ts1, ts2, *args, **kwargs):
     ... , datetime.datetime(2000, 9, 30, 0, 0)]
 
     See Also
-    ========
+    --------
     tOverlapHalf
     tCommon
     """
@@ -266,7 +266,7 @@ def tOverlapHalf(ts1, ts2, presort=False):
     returned by tOverlap.
 
     Parameters
-    ==========
+    ----------
     ts1 : list
         first set of datetime object
     ts2 : list
@@ -278,14 +278,14 @@ def tOverlapHalf(ts1, ts2, presort=False):
                    the list if one sort can be done for many calls to tOverlap
 
     Returns
-    =======
+    -------
     out : list
         indices of ts2 within interval of ts1
 
         **note:** Returns empty list if no overlap found
 
     See Also
-    ========
+    --------
     tOverlap
     tCommon
     """
@@ -304,24 +304,24 @@ def tCommon(ts1, ts2, mask_only=True):
     Finds the elements in a list of datetime objects present in another
 
     Parameters
-    ==========
+    ----------
     ts1 : list or array-like
         first set of datetime objects
     ts2 : list or array-like
         second set of datetime objects
 
     Returns
-    =======
+    -------
     out : tuple
         Two element tuple of truth tables (of 1 present in 2, & vice versa)
 
     See Also
-    ========
+    --------
     tOverlapHalf
     tOverlap
 
     Examples
-    ========
+    --------
     >>> import spacepy.toolbox as tb
     >>> import numpy as np
     >>> import datetime as dt
@@ -387,21 +387,21 @@ def loadpickle(fln):
     load a pickle and return content as dictionary
 
     Parameters
-    ==========
+    ----------
     fln : string
         filename
 
     Returns
-    =======
+    -------
     out : dict
         dictionary with content from file
 
     See Also
-    ========
+    --------
     savepickle
 
     Examples
-    ========
+    --------
     **note**: If fln is not found, but the same filename with '.gz'
            is found, will attempt to open the .gz as a gzipped file.
 
@@ -460,11 +460,11 @@ def savepickle(fln, dict, compress=None):
                      compressed file exists and the uncompressed does not.
 
     See Also
-    ========
+    --------
     loadpickle
 
     Examples
-    ========
+    --------
     >>> d = {'grade':[1,2,3], 'name':['Mary', 'John', 'Chris']}
     >>> savepickle('test.pbin', d)
     """
@@ -498,12 +498,12 @@ def assemble(fln_pattern, outfln, sortkey='ticks', verbose=True):
         filename to save combined files to
 
     Returns
-    =======
+    -------
     out : dict
         dictionary with combined values
 
     Examples
-    ========
+    --------
     >>> import spacepy.toolbox as tb
     >>> a, b, c = {'ticks':[1,2,3]}, {'ticks':[4,5,6]}, {'ticks':[7,8,9]}
     >>> tb.savepickle('input_files_2001.pkl', a)
@@ -662,6 +662,11 @@ def dictree(in_dict, verbose=False, spaces=None, levels=True, attrs=False, **kwa
     attrs : boolean (optional)
         display information for attributes
 
+    Raises
+    ------
+    TypeError
+        Input does not have keys or attrs, cannot build tree.
+
     Examples
     --------
     >>> import spacepy.toolbox as tb
@@ -755,7 +760,7 @@ def _crawl_yearly(base_url, pattern, datadir, name=None, cached=True,
     """Crawl files in a directory-by-year structure
 
     Parameters
-    ==========
+    ----------
     base_url : str
         Base of the data. This URL should point to a directory containing
         yearly directories (YYYY).
@@ -778,7 +783,7 @@ def _crawl_yearly(base_url, pattern, datadir, name=None, cached=True,
         years prior to this which have already been downloaded.
 
     Returns
-    =======
+    -------
     list
         All the filenames that were mirrored, in order; or None if there
         were no updates. If there are any updates, all filenames are
@@ -852,7 +857,7 @@ def _get_qindenton_daily(qd_daily_url=None, cached=True, startyear=None):
     """Download the Qin-Denton OMNI-like daily files
     
     Parameters
-    ==========
+    ----------
     qd_daily_url : str (optional)
         Base of the Qin-Denton data, in hourly JSON-headed ASCII. This URL
         should point to the directory containing the yearly directories.
@@ -866,7 +871,7 @@ def _get_qindenton_daily(qd_daily_url=None, cached=True, startyear=None):
         rather than all years. This will delete older files!
 
     Returns
-    =======
+    -------
     SpaceData
         The data extracted from the Q-D dataset, fully processed for saving
         as SpacePy HDF5 OMNI data.
@@ -885,12 +890,17 @@ def _assemble_qindenton_daily(qd_daily_dir):
     """Assemble Qin-Denton daily files into OMNI structure
 
     Parameters
-    ==========
+    ----------
     qd_daily_dir : str
         Directory with Qin-Denton daily files.
 
+    Raises
+    ------
+    ValueError
+        Qin-Denton file ill formed with different lengths by variable
+
     Returns
-    =======
+    -------
     SpaceData
         The data extracted from the Q-D dataset, fully processed for saving
         as SpacePy HDF5 OMNI data.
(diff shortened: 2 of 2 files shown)


spacepy/toolbox/__init__.py
```diff
@@ -1880,27 +1913,27 @@ def logspace(min, max, num, **kwargs):
...
-    =======
+    -------
     out : array
         log-spaced bins from min to max in a numpy array
 
     Notes
-    =====
-    This function works on both numbers and datetime objects
+    -----
+    This function works on both numbers and datetime objects. This function is not leap second aware.
```
jtniehof, line 1927: Seems obvious since datetime isn't leapsecond aware, but I presume the idea is that the spacing is wrong in an absolute system (e.g. 24 bins from 2016-12-31T00 through 2017-01-01 will be 3600s apart not 3600.04 in the linspace case). Might be worded more clearly?

Total nitpick: Would just put "Works on both numbers and datetime objects. Not leapsecond aware." Sentence fragments seem fine.

spacepy/toolbox/__init__.py
```diff
@@ -1927,27 +1960,27 @@ def linspace(min, max, num, **kwargs):
...
-    =======
+    -------
     out : array
         linear-spaced bins from min to max in a numpy array
 
     Notes
-    =====
-    This function works on both numbers and datetime objects
+    -----
+    This function works on both numbers and datetime objects. This function is not leap second aware.
```
jtniehof, line 1974: See logspace notes.

spacepy/toolbox/__init__.py
```diff
@@ -2209,18 +2242,23 @@ def query_yes_no(question, default="yes"):
...
+    ----------
     question : string
         the question to ask
     default : string (optional)
 
+    Raises
+    ------
+    ValueError
+        The default answer is invalid.
+
```
jtniehof, line 2254: As above, I think it's better to document the valid defaults!

jtniehof (commented):
Big question here is whether we want to document every single exception; see detailed comments.

spacepy/toolbox/__init__.py
```diff
@@ -1536,13 +1560,22 @@ def windowMean(data, time=[], winsize=0, overlap=0, st_time=None, op=np.mean):
...
 
+    Raises
+    ------
+    ValueError
+        data and time have different lengths
+    TypeError
+        winsize or overlap are not datetime.timedelta
+    ValueError
+        The requested overlap is larger than requested window
+
```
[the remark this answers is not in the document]
drsteve: Window size and overlap can be integers (number of array elements).  You can't mix and match though, and supplying a time array through the keyword argument implies time-based binning. If you're using time-based then `timedelta`s have to be used, if you're using index-based binning then indices need to be passed.

balarsen:
> 
> Big question here is whether we want to document every single exception; see detailed comments.

I think we likely should, complete documentation.

CHANGELOG
```diff
@@ -1,5 +1,8 @@
 Changes in Version 0.2.3 (2021-xx-xx)
 =====================================
+documentation
+ - toolbox cleaned for numpydoc format and Raises sections added
+
```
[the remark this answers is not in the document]
balarsen: Can do that, seemed to make sense at the time.

spacepy/toolbox/__init__.py
```diff
@@ -1124,17 +1134,26 @@ def get_url(url, outfile=None, reporthook=None, cached=False,
...
         ``keepalive``. If not provided, will attempt to make a connection.
 
+    Raises
+    ------
+    ValueError
+        If conn is provided and keepalive is not True
+    RuntimeError
+        HTTP status >= 400 returned
+    RuntimeError
+        If cached is True and an outfile is not set.
```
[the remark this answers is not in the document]
balarsen: > 
>     1. The cached `True` without `outfile` should probably be a `ValueError`...yes that's how I wrote it but documenting it like this, it seems clearly wrong.

Should I update that here? Can without issue. Likely justifies a change entry in the docs. 

> 
>     2. The numpydoc guide isn't very clear on this, but I think it should probably be one entry per exception type, and will probably take some playing to get the formatting right. So (assuming 1) "If conn is provided and keepalive is False, or cached is True and no outfile provided." Might also be nice to have them alphabetical by exception type.

I feel like the guide examples double up on exception type and in the rendering I feel like it makes the most sense. Otherwise the doc has to say something like "bla bla or bla bla" fine for short messages but tough for long ones. 

> 
>     3. numpydoc guide says "This section should be used judiciously, i.e., only for errors that are non-obvious or have a large chance of getting raised." I think it should be obvious that putting in invalid combinations of arguments will result in errors (the parameters section should probably make invalid combinations more apparent, though.) The HTTP error is _maybe_ non-obvious, but it's pretty clear what happened if it is raised. Do we want to have a house style to always have this? At a certain point it degenerates into the Java "throws" declarations...

Fair enough. Drop this one as the message is clear?
```
    RuntimeError
        HTTP status >= 400 returned
```

spacepy/toolbox/__init__.py
```diff
@@ -1259,17 +1278,22 @@ def update(all=True, QDomni=False, omni=False, omni2=False, leapsecs=False,
         timestamp on local file (default). Set False to always
         download files.
 
+    Raises
+    ------
+    RuntimeError
+        OMNI2 zip file is not as expected
+
```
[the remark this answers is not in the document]
balarsen: Sure, I erred on the side of all but that does make decent sense. And the message is ok in the error.

spacepy/toolbox/__init__.py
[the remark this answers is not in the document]
balarsen: I'll poke here. The docstring and the code error messages got confusing on what was required.

spacepy/toolbox/__init__.py
balarsen: Updated, I think we need to be explicit here as if it fell through to `time` it could be leapsecond aware.

spacepy/toolbox/__init__.py
balarsen: DOne

spacepy/toolbox/__init__.py
balarsen: This whole function is a mess.... updated to have what is allowed.

spacepy/toolbox/__init__.py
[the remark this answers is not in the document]
balarsen: Updated, have a look after the push. I think well. Dropped the Raises in favor of notes in Parameters and did update to `ValueError`

spacepy/toolbox/__init__.py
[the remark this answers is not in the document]
balarsen: @drsteve What do I not understand here about the flow? `timedelta` is pretty clearly required. 

```
        try:
            assert type(winsize) == datetime.timedelta
            assert type(overlap) == datetime.timedelta
        except AssertionError:
            raise TypeError('windowmean error: winsize/overlap must be timedeltas')
```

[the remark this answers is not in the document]
jtniehof: Looks like we need a unit test for the non-timedelta case. I wonder something got refactored poorly at some point.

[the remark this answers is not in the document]
drsteve: @balarsen `timedelta` is very clearly required *if* a time array is supplied.
The section you're pasting from starts with:
```
if len(time) == 0:
        startpt, res = 0, 1
        time = list(range(len(data)))
        pts = True
    else:
```
and then the time-based stuff is in the `else`.
Similarly, if you come down past the error checking then the whole thing is in an `if` control block. 
```
    if pts:
        #loop for fixed number of points in window
        ...
    else:
        #loop with time-based window
```
Probably the thing that's missing is a clear statement that *if* a time array is supplied *then* the function assumes time-based windowing and requires `winsize` and `overlap` to be `timedelta`s. If a time array is not supplied the expectation is points-based windowing and `winsize` and `overlap` are expected to be integer.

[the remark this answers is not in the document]
jtniehof: To me it's not clear that `time` being of non-zero length means `time` must be `datetime` and `winsize`, `overlap` must be `timedelta`, since the docs say `time` can be numeric or `datetime`. If `time` is numeric I wouldn't expect `winsize` to be `timedelta`. Am I missing something?

spacepy/toolbox/__init__.py
jtniehof: There's an explicit format for listing available options; if this is used in the parameters then the Raises can go away:
`default : { 'yes', 'no', None} (optional)`

Also, this isn't your change but as long as you're fixing things, the type is `str` not `string`.

CHANGELOG
```diff
@@ -1,5 +1,7 @@
 Changes in Version 0.2.3 (2021-xx-xx)
 =====================================
+toolbox
+ - docstrings cleaned for numpydoc format and Raises documented where it added value
```
jtniehof, line 4: Over 80 character, wrapping would be nice (it's inconsistent in CHANGELOG, alas.)

spacepy/toolbox/__init__.py
```diff
@@ -1102,8 +1112,12 @@ def get_url(url, outfile=None, reporthook=None, cached=False,
 
     This is similar to the deprecated ``urlretrieve``.
 
+    .. versionchanged:: 0.2.3
+       In earlier versions of SpacePy invalid combinations of cached and outfile raised RuntimeError,
+       changed to ValueError.
```
jtniehof, line 1117: Line too long, docstrings should wrap at 72 per PEP8.

spacepy/toolbox/__init__.py
```diff
@@ -1114,27 +1128,27 @@ def get_url(url, outfile=None, reporthook=None, cached=False,
     cached : bool (optional)
         Compare modification time of the URL to the modification time
         of ``outfile``; do not retrieve (and return None) unless
-        the URL is newer than the file.
+        the URL is newer than the file. Is set outfile is required.
```
jtniehof, line 1131: If set, not Is set. I would think that the comparison to outfile implies outfile is required, but maybe not....

spacepy/toolbox/__init__.py
```diff
@@ -1114,27 +1128,27 @@ def get_url(url, outfile=None, reporthook=None, cached=False,
...
     keepalive : bool (optional)
         Attempt to keep the connection open to retrieve more URLs.
         The return becomes a tuple of (data, conn) to return the
         connection used so it can be used again. This mode does not
-        support proxies. (Default False)
+        support proxies. Required to be True if conn is provided. (Default False)
```
jtniehof, line 1136: Too long line again.

spacepy/toolbox/__init__.py
```diff
@@ -1520,12 +1534,12 @@ def windowMean(data, time=[], winsize=0, overlap=0, st_time=None, op=np.mean):
...
 
     Parameters
-    ==========
+    ----------
     data : array_like
         1D series of points
     time : list (optional)
         series of timestamps, optional (format as numeric or datetime)
-        For non-overlapping windows set overlap to zero.
+        For non-overlapping windows set overlap to zero. Must be same length as data.
```
jtniehof, line 1542: Too long again.

spacepy/toolbox/__init__.py
```diff
@@ -1880,27 +1894,27 @@ def logspace(min, max, num, **kwargs):
...
-    =======
+    -------
     out : array
         log-spaced bins from min to max in a numpy array
 
     Notes
-    =====
-    This function works on both numbers and datetime objects
+    -----
+    This function works on both numbers and datetime objects. Not leapsecond aware.
```
jtniehof, line 1908: Too long again.

spacepy/toolbox/__init__.py
```diff
@@ -1927,27 +1941,27 @@ def linspace(min, max, num, **kwargs):
...
-    =======
+    -------
     out : array
         linear-spaced bins from min to max in a numpy array
 
     Notes
-    =====
-    This function works on both numbers and datetime objects
+    -----
+    This function works on both numbers and datetime objects. Not leapsecond aware.
```
jtniehof, line 1955: Too long again.

jtniehof (changes requested):
In addition to the linewrap, there's a merge commit in there so the history isn't linear, and we won't be able to put it on master. Needs to be rebased.

jtniehof:
Also since this fixes #484 now, need to update the PR description to note that.

spacepy/toolbox/__init__.py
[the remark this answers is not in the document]
drsteve: The docs don't line up with the code here, and without doing archaeology I'm not sure whether this was functionality that used to exist, or whether the docs are just misleading allowing time to be numeric. However, if we're making sure the documentation lines up with behaviour then supplying a time array pushes us into checking for `winsize` and `overlap` being `datetime.timedelta`.
```
In [13]: tb.windowMean([3]*20, time=range(20))
AssertionError                            Traceback (most recent call last)
~/miniconda3/envs/default/lib/python3.8/site-packages/spacepy/toolbox/__init__.py in windowMean(data, time, winsize, overlap, st_time, op)
   1601         try:
-> 1602             assert type(winsize) == datetime.timedelta
   1603             assert type(overlap) == datetime.timedelta

AssertionError:

During handling of the above exception, another exception occurred:

<ipython-input-13-2f219d46f123> in <module>
----> 1 tb.windowMean([3]*20, time=range(20))

~/miniconda3/envs/default/lib/python3.8/site-packages/spacepy/toolbox/__init__.py in windowMean(data, time, winsize, overlap, st_time, op)
   1603             assert type(overlap) == datetime.timedelta
   1604         except AssertionError:
-> 1605             raise TypeError('windowmean error: winsize/overlap must be timedeltas')
   1606         pts = False #force time-based averaging
   1607         if (type(time[0]) != datetime.datetime):

TypeError: windowmean error: winsize/overlap must be timedeltas
```
If the time input then isn't `datetime` it'll error at trying to work add timedeltas to numeric values:
```
----> 1 tb.windowMean([3]*20, time=range(20), winsize=dt.timedelta(1), overlap=dt.timedelta(0))

~/miniconda3/envs/default/lib/python3.8/site-packages/spacepy/toolbox/__init__.py in windowMean(data, time, winsize, overlap, st_time, op)
   1641     else:
   1642         #loop with time-based window
-> 1643         lastpt = time[0] + winsize
```

Omitting the time array allows us to use numeric window size and overlap since it assumes the indices as timestamps and then indices, winsize and overlap are all on the same scale:
```
In [15]: tb.windowMean([3]*20, winsize=4, overlap=0)
Out[15]: ([3.0, 3.0, 3.0, 3.0], [2.0, 6.0, 10.0, 14.0])
```

For fixing documentation I think clarifying current behaviour is the right thing to do. Moving forward it's probably good to support non-datetime timestamps.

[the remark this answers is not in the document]
jtniehof: I checked the history. From the first CVS commit, `winsize` and `overlap` had to be `timedelta`. At various points in the history, there were some `num2date` things going on that I didn't look at too closely, but might have been trying to make matplotlib timestamps work.

So I agree, document and maybe open an enhancement issue to support non-`datetime`. `rebin` might actually be useful for that.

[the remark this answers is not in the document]
balarsen: I got this cleaned to I think an acceptable min state in the error message.

spacepy/toolbox/__init__.py
balarsen: updated

spacepy/toolbox/__init__.py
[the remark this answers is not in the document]
balarsen: good in 1aa9067e02c6d8d26ecd3023fbc349c3cd062df2

spacepy/toolbox/__init__.py
balarsen: good in 1aa9067e02c6d8d26ecd3023fbc349c3cd062df2

CHANGELOG
balarsen: @jtniehof There was a conversation here a bit ago. Are we doing 80? Seems really small, I can't remember the outcome.

spacepy/toolbox/__init__.py
balarsen: Updated.

spacepy/toolbox/__init__.py
balarsen: Fixed, I have messed this up on a few occasions and feel like explicit is good.

spacepy/toolbox/__init__.py
balarsen: fixed

spacepy/toolbox/__init__.py
balarsen: Fixed, I note there are dozens of too ling lines that were not touched in this PR.

spacepy/toolbox/__init__.py
balarsen: updated

jtniehof (approved):
Okay, I'm good with this now. @balarsen , can you do a quick check that I didn't do terrible things to it before I merge?

jtniehof:
Unless someone objects, this will be merged no earlier than Friday 23 September

Issue #497, opened 2021-01-27
Find uncaught warnings in CI


jtniehof:
The CI really should fail on uncaught warnings. I'm not sure exactly the best way to do this; `-Werror` seems to raise the exception immediately so `catch_warnings` is ineffective and there are lots of false positives (although that's from memory.) Maybe setting a really clever warnings filter? Or we can grep the output of the tests and look for uncaught warnings. It would just be nice to have something other than "@balarsen upgraded numpy and opened an issue" to find these things.

Obviously #455, #425, #461, and any other pending warnings need to be cleaned up before this happens.

jtniehof:
Couple of possible approaches that come to mind:

We could wrap the call to `unittest.main` in `test_all.py` with a `catch_warnings` to catch uncaught warnings and then fail. But I don't know if `unittest.main` returns. We could also get fancy and extend `unittest.main` or subclass `TestCase` to see if there are uncaught warnings...this could even extend to having a different status reported (e.g. W instead of F for fail, E for error).

drsteve:
So I'm torn on having the CI actually _fail_ on uncaught warnings. The CI indicates to the wider world whether we're passing tests, but a reasonable user could easily infer, from seeing that banner saying CI failed, that there are unresolved (known) bugs affecting functionality. Uncaught warnings from upstream don't necessarily rise to that level.

I do see the need to trap them automatically somehow so we get notified, and obviously having the tests fail is an easy way to do that.

My ideal solution - and I don't know if this is possible - would be to have our CI open an issue (or comment on a PR, whatever) when there are uncaught warnings and assign to devs.
It looks to me like GH Actions can be used to open an issue (https://github.com/marketplace/actions/open-github-issue). Not sure about how we'd set that workflow up, but it avoids the appearance of SpacePy being "broken" when it's an upstream `DeprecationWarning` that doesn't immediately impact functionality or behaviour.

Subclassing `TestCase` to report a different status could help with a workflow like that.

jtniehof:
I think it would be possible for each test job to parse the output and look
for warnings, then pass the information out to the "all tests" job which
could take action. We wouldn't want the test job itself to take the action
since that would result in opening an issue for every Python version,
dependency version, etc. There's a concept of "output" from the job that
other jobs can look at and we could use that.

I'm thinking the right thing would be for a test to fail if it's a PR
build, and pass but open an issue if it's a cron job build. The trick would
be to not open an issue every single time the cron job runs. I don't know
[...]

PR #505, opened 2021-02-04
Fix various tests for specific numbers of warnings

jtniehof:
We had several tests that either checked code which used to issue a warning and wanted to make sure it was fixed so it didn't issue that warning, or where we were verifying that the code was issuing a warning it was supposed to. Unfortunately they were implemented as "issues no warnings at all" or "issues exactly one warning that matches what we expect." numpy 1.20 deprecates a bunch of stuff so suddenly there are a lot more warnings and these tests fail. This is why the cron CI run fails.

This PR tightens those tests so they check for only what we really care about: a particular warning is or is not issued.

## PR Checklist

### DIFF
--- a/tests/test_pycdf.py
+++ b/tests/test_pycdf.py
@@ -2680,7 +2680,7 @@ class ChangeCDF(ChangeCDFBase):
             [[1, 2, 3], [4, 5, 6]], self.cdf['newzVar'][...])
 
     def testNewVarTime(self):
-        with warnings.catch_warnings(record=True) as w:
+        with warnings.catch_warnings(record=True) as cm:
             if sys.version_info[0:2] == (2, 7)\
                and hasattr(cdf, '__warningregistry__'):
                 # filter 'always' is broken in Python 2.7
@@ -2693,14 +2693,18 @@ class ChangeCDF(ChangeCDFBase):
                         break
             warnings.simplefilter('always')
             self.cdf['newzVar'] = [datetime.datetime(2010, 1, 1)]
-        self.assertEqual(1, len(w))
-        self.assertEqual(
-            w[0].category, DeprecationWarning)
-        self.assertEqual(
-            'No type specified for time input; assuming CDF_EPOCH.'
-            ' This will change to TT2000 in the future, on systems which'
-            ' support it.',
-            str(w[0].message))
+        n_match = 0
+        for w in cm:
+            if w.category is DeprecationWarning\
+               and str(w.message) \
+               == 'No type specified for time input; assuming CDF_EPOCH.'\
+               ' This will change to TT2000 in the future, on systems which'\
+               ' support it.':
+                n_match += 1
+            else:
+                warnings.showwarning(
+                    w.message, w.category, w.filename, w.lineno)
+        self.assertEqual(1, n_match)
         # For future
         #expected = cdf.const.CDF_TIME_TT2000.value if cdf.lib.supports_int8 \
         #           else cdf.const.CDF_EPOCH.value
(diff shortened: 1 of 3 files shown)


balarsen:
I'm happy for the answer to be yes, but does all this working work really add value to our lives? I worry that in chasing warnings we are not fixing issues or adding new functionality or burning the list of open issues (like docs).

drsteve:
> I'm happy for the answer to be yes, but does all this working work really add value to our lives? I worry that in chasing warnings we are not fixing issues or adding new functionality or burning the list of open issues (like docs).

I'm going to say yes.
- Upstream warnings shouldn't make our tests fail (tests tell the wider world whether we have a problem).
- Upstream warnings like the latest set from the numpy 1.20 release affect nothing _now_ but could become breaking later. So we need to be aware and fix (see #497).
- If we deprecate something in SpacePy we probably should make sure the that we're actually getting a `DeprecationWarning`.
- When we finally remove the deprecated function (or function version) the deprecation test will fail, so that helps us maintain the test suite.

I don't think this sets the bar too high for contributions either. There's no need for most contributions to deal with warnings, and the dev team can usually add commits to a PR if we need to. So if there's a contribution and it covers everything but trapped warnings? Well, I say we can review and either suggest a patch in comments or just add a commit to the PR. 

Having said that, yeah, it'd be nice to spend the time working on features or docs instead. But getting this straight now should actually save time later.

jtniehof:
> I'm happy for the answer to be yes, but does all this working work really add value to our lives?

Define "all this working." Do you mean this PR? Without it tests are failing. Do you mean fixing the numpy warnings? They're usually stuff that is going to eventually bit us (deprecations). Do you mean testing that the fixes for numpy warnings work? I can maybe see ditching that, but we do have a general principle of unit tests for fixes and I think we have had them regress before. Do you mean testing that code where we're supposed to issue a warning actually does? Again, I'm so-so, but would prefer to.

To both issues of our time and the bar for contribution: it's time to start with a SpacePy test helper, like `numpy.testing` or [dbp_testing.py](https://github.com/spacepy/dbprocessing/blob/master/unit_tests/dbp_testing.py) In that context I think I can genericize what I've been doing to something like:
```python
with spacepy_testing.assertWarns(self, DeprecationWarning, '^spacepy.pycdf', 'No time type specified'):
    do_stuff()

with spacepy_testing.assertNoWarns(self, DeprecationWarning, '^numpy', 'Indexing with non-tuple sequence'):
    do_stuff()
```

balarsen:
> To both issues of our time and the bar for contribution: it's time to start with a SpacePy test helper, like `numpy.testing` or [dbp_testing.py](https://github.com/spacepy/dbprocessing/blob/master/unit_tests/dbp_testing.py) In that context I think I can genericize what I've been doing to something like:

There is a `numpy.testing.assert_warns` that hopefully we can use.

jtniehof:
> There is a `numpy.testing.assert_warns` that hopefully we can use.

It doesn't distinguish between one `DeprecationWarning` and another, and I haven't checked that it passes other warnings through.

drsteve (commented):
Looks like this covers it. 

Question:
The quaternion deprecation test in test_toolbox fails for me in Python 2.7 (when I run `test_toolbox`, but not when I run `test_all` (odd, but so is `warnings` in 2.7).  Should those tests be updated to follow this approach? If so, is that appropriate for this PR?

jtniehof:
> The quaternion deprecation test in test_toolbox fails for me in Python 2.7 (when I run `test_toolbox`, but not when I run `test_all` (odd, but so is `warnings` in 2.7). Should those tests be updated to follow this approach? If so, is that appropriate for this PR?

What I'd suggest is merging this to stop the bleeding and then I can genericize and go on a little romp through the test suite in a different PR.

drsteve (approved):

PR #506, opened 2021-02-06
Unittest helper module; test warnings; test and document against built SpacePy

jtniehof:
This PR combines a number of quality-of-life improvements for the unit test suite and docs:

- Adds a new test support module, `spacepy_testing`.
- Adds `assertWarns`/`assertDoesntWarn` context managers to make testing of warnings easier per conversation in #505.
- Uses these new context managers to clean up existing warning tests
- Does some further warnings cleanup where we had tests that were enabling all warnings with `simplefilter always` and then eating them with `catch_warnings` and doing nothing with the output.
- Changes the unit tests to run against the _built_ SpacePy instead of the _installed_ SpacePy, so you can test without installing. This will make it much easier for us to cache dependencies in the future, although that is left out here.
- Similarly changes the documentation to run against the built SpacePy instead of the source. This allows the removal of some workarounds in the code where compiled code wasn't available to the documentation builder, and should make cleaning up docs warnings easier in the future.

## Open questions

1. This uses `assertWarns` and `assertDoesntWarn` as two different classes (`DoesntWarn` is an almost trivial subclass). I think this is easier from the user perspective, although it probably could be an argument to `__init__` instead. Also I'm not thrilled with `assertDoesntWarn` as a name.
2. `assertX` reproduces the warnings filter syntax exactly. I don't think `lineno` is very useful, but it's in the warnings filter so I left it in. Also the `action` should almost always be `always`; I made the default be to add no filter, but maybe this should be the default. Or maybe `action` should be left out and just forced to `always`.
3. Do we want the ability to specify the number of times a warning should be issued instead of just lock to one-or-none? For warnings that are raised on successive lines, I just used the context manager twice. This would complicate the argument list but would also be a way to back away from having two different classes (just specify 0 if the warning shouldn't be raised.)

## PR Checklist

### DIFF
--- a/tests/test_empiricals.py
+++ b/tests/test_empiricals.py
@@ -6,6 +6,7 @@ import warnings
 
 import dateutil.parser as dup
 import numpy as np
+import spacepy_testing
 import spacepy.time as spt
 import spacepy.toolbox as tb
 import spacepy.empiricals as em
@@ -79,14 +80,11 @@ class empFunctionTests(unittest.TestCase):
         self.assertRaises(KeyError, spam1)
 
     def test_getPlasmaPauseCA1992warn(self):
-        with warnings.catch_warnings(record=True) as w:
-            warnings.simplefilter('always', category=RuntimeWarning)
+        with spacepy_testing.assertWarns(
+                self, 'always',
+                r'No LT dependence currently supported for CA1992 model',
+                RuntimeWarning, r'spacepy\.empiricals$'):
             em.getPlasmaPause(self.ticks, model='CA1992', LT=12, omnivals=self.omnivals)
-        self.assertEqual(1, len(w))
-        self.assertEqual(RuntimeWarning, w[0].category)
-        self.assertEqual(
-            'No LT dependence currently supported for CA1992 model',
-            str(w[0].message))
 
     def test_getLmax(self):
         """getLmax should give known results (regression)"""
--- a/tests/test_pycdf.py
+++ b/tests/test_pycdf.py
@@ -31,6 +31,7 @@ import warnings
 import matplotlib.dates
 import numpy
 import numpy.testing
+import spacepy_testing
 from spacepy import datamodel
 import spacepy.pycdf as cdf
 import spacepy.pycdf.const as const
@@ -690,8 +691,8 @@ class MakeCDF(unittest.TestCase):
     def setUp(self):
         self.testdir = tempfile.mkdtemp()
         self.testfspec = os.path.join(self.testdir, 'foo.cdf')
-        pth = os.path.dirname(os.path.abspath(__file__))
-        self.testmaster = os.path.join(pth, 'po_l1_cam_testc.cdf')
+        self.testmaster = os.path.join(spacepy_testing.testsdir,
+                                       'po_l1_cam_testc.cdf')
 
     def tearDown(self):
         shutil.rmtree(self.testdir)
@@ -1138,16 +1139,14 @@ class CDFTestsBase(unittest.TestCase):
 
 class CDFTests(CDFTestsBase):
     """Tests that involve an existing CDF, read or write"""
-    pth = os.path.dirname(os.path.abspath(__file__))
-    testmaster = os.path.join(pth, 'po_l1_cam_test.cdf')
+    testmaster = os.path.join(spacepy_testing.testsdir, 'po_l1_cam_test.cdf')
     testbase = 'test.cdf'
     expected_digest = '94515e62d38a31ad02f6d435274cbfe7'
 
 
 class ColCDFTests(CDFTestsBase):
     """Tests that involve an existing column-major CDF, read or write"""
-    pth = os.path.dirname(os.path.abspath(__file__))
-    testmaster = os.path.join(pth, 'po_l1_cam_testc.cdf')
+    testmaster = os.path.join(spacepy_testing.testsdir, 'po_l1_cam_testc.cdf')
     testbase = 'testc.cdf'
     expected_digest = '7728439e20bece4c0962a125373345bf'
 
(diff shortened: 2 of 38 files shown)


jtniehof:
Force-push just to resolve the conflict rebase.

tests/test_pycdf.py
```diff
@@ -690,8 +693,8 @@ class MakeCDF(unittest.TestCase):
     def setUp(self):
         self.testdir = tempfile.mkdtemp()
         self.testfspec = os.path.join(self.testdir, 'foo.cdf')
-        pth = os.path.dirname(os.path.abspath(__file__))
-        self.testmaster = os.path.join(pth, 'po_l1_cam_testc.cdf')
+        self.testmaster = os.path.join(spacepy_testing.testsdir,
```
drsteve, line 696: Is there any particular reason to keep the test cdfs in the `tests` directory rather than `tests/data`?

tests/test_empiricals.py
```diff
@@ -79,14 +80,11 @@ def test_getPlasmaPauseErrors2(self):
         self.assertRaises(KeyError, spam1)
 
     def test_getPlasmaPauseCA1992warn(self):
-        with warnings.catch_warnings(record=True) as w:
-            warnings.simplefilter('always', category=RuntimeWarning)
+        with spacepy_testing.assertWarns(
+                self, 'always',
+                r'No LT dependence currently supported for CA1992 model',
```
drsteve, line 85: Just to confirm, these need to be raw strings because they get passed to a regex in the class? I mean, it all seems to work, but I haven't made sure yet that they're all actually raw strings. Likewise, I don't think I saw docs saying the strings passed here need to be passed as raw strings.

tests/test_time.py
```diff
@@ -130,17 +131,13 @@ def test_sec2hms(self):
                     [0, 0, 1],
                     [0, 0, 30],
                     [0, 59, 59])
-        with warnings.catch_warnings(record=True) as w:
-            warnings.filterwarnings(
-                'always', 'Number of seconds > seconds in day.*',
-                UserWarning, '^spacepy\\.time')
+        with spacepy_testing.assertWarns(
+                self, 'always',
+                'Number of seconds > seconds in day\. Try days keyword\.$',
```
drsteve, line 136: I guess my earlier question about raw strings applies here.  It might be good to highlight (somewhere) when and why raw strings need to be used in calling this.

drsteve (commented):
To respond to the questions/comments in the PR:
1. _This uses assertWarns and assertDoesntWarn as two different classes ... although it probably could be an argument to `__init__` instead_.
I prefer having them separate to selecting by argument. The name doesn't worry me. It's clear enough.
2. _assertX reproduces the warnings filter syntax exactly. I don't think lineno is very useful, but it's in the warnings filter so I left it in. Also the action should almost always be always; I made the default be to add no filter, but maybe this should be the default. Or maybe action should be left out and just forced to always._
Reproducing filter syntax is nice here and is one less thing to remember for folks who are familiar enough with that syntax already. Having the default action as always is probably good, but I also probably would keep it set explicitly in the calls (for clarity).
3. _Do we want the ability to specify the number of times a warning should be issued instead of just lock to one-or-none? For warnings that are raised on successive lines, I just used the context manager twice. This would complicate the argument list but would also be a way to back away from having two different classes (just specify 0 if the warning shouldn't be raised.)_
Current behavior is good with me.

tests/test_pycdf.py
jtniehof: > Is there any particular reason to keep the test cdfs in the `tests` directory rather than `tests/data`?

Because they were the first test data we had and I wasn't organized enough to have a "data" directory yet. (Actually it's probably because they were the first _tests_ and predate the inclusion of pycdf in spacepy.) I'm happy to move them but probably out of scope for this PR; I can open another.

tests/test_empiricals.py
jtniehof: [There's no such thing as a raw string](https://stackoverflow.com/a/2081708); "raw string" is sloppy colloquialism for "[raw string literal](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals)":
```python
raw = r'\foo'
print(type(raw)) #<type 'str'>
print(repr(raw)) #'\\foo'
cooked = '\\foo'
print(type(cooked)) #<type 'str'>
print(repr(cooked)) #'\\foo'
print(raw == cooked) #True
```

I tend to use raw literals whenver I do a regular expression just so I only have to worry about escaping the regex and not escaping the string as well. Even if there's nothing that technically needs escaping in there, I might come back and add something later and get bit. I also find it's a bit of a hint to the reader that this thing is a regex.

But the caller can construct their strings however they like.

tests/test_time.py
jtniehof: Same answer as above. I _am_ being inconsistent with myself here, partly because I was copying it from an equality test not a regex test. I can make it a raw if it reads better.

jtniehof:
>        Reproducing filter syntax is nice here and is one less thing to remember for folks who are familiar enough with that syntax already. Having the default action as always is probably good, but I also probably would keep it set explicitly in the calls (for clarity).

Sounds reasonable. If we change the default to `always` then people would have to pass in an explicit `None` if they don't want to change the warnings filter. (Note the change only happens within the context manager block, not globally.) But it's not like the default matters a lot: since it's so early in the argument list, you're almost always specifying it or having to use keyword syntax after, doesn't save on a lot.

So if explicit `None` passing in doesn't make you feel squicky, I'll go ahead and make that change and push it.

Thanks on the other questions, too.

drsteve:
Explicit `None` is fine by me.

drsteve (approved):

PR #510, opened 2021-02-19
PyBats: Support for binary "outs" IDL formatted files.

spacecataz:
## PR Checklist
IdlFile objects can now handle "outs" files - files that contain several "frames" (data entries for a single epoch) combined into a single file.  Users can navigate the frames using the "switch_frame" method to move to an arbitrary frame.

A new function decorator is introduced to help create calculation functions that update when frames are changed.

New functionality is documented in docstrings for IdlFile and Bats2d objects.

ASCII *.outs files are not supported yet; but binary is the most pressing need for this oft-requested feature.


drsteve:
I've started looking over this, and hopefully I'll make it through fairly quickly.

Initial comments:
1. Commit d281d41113be1b21f4c4eb53eb5c6e6838448508 added a `mag_grid[...].outs` file to the test data directory. That's not used at all, or even readable by `spacepy.pybats.bats.MagGridFile`. I'm guessing it was added in anticipation? If this PR only handles binary `.outs`, and at that only binary 2D MHD `.outs` files, then that file doesn't really need to be here. (I'd suggest amending that commit to remove the file. It should probably be added when it's used/usable)
2. Since the support is only for the binary 2D MHD slices ("IDL" files), should that be made clearer in the `CHANGELOG`? Related, what's the current mode of failure if an ASCII `.outs` is read (and is it tested)?
3. My flake8 plugin is freaking out about these files... can you do a quick pass to clean up `pybats/__init__.py` and `pybats/bats.py`? At least hit some of the easier things like trailing whitespace and tabs? And maybe bring the new code into closer alignment with PEP8 (whitespace around operators, after commas, etc.)

I'll get to a full review in the next day or two.

spacecataz:
> 1. Commit [d281d41](https://github.com/spacepy/spacepy/commit/d281d41113be1b21f4c4eb53eb5c6e6838448508) added a `mag_grid[...].outs` file to the test data directory. That's not used at all, or even readable by `spacepy.pybats.bats.MagGridFile`. I'm guessing it was added in anticipation? If this PR only handles binary `.outs`, and at that only binary 2D MHD `.outs` files, then that file doesn't really need to be here. (I'd suggest amending that commit to remove the file. It should probably be added when it's used/usable)
> 2. Since the support is only for the binary 2D MHD slices ("IDL" files), should that be made clearer in the `CHANGELOG`? Related, what's the current mode of failure if an ASCII `.outs` is read (and is it tested)?

Taking these two together:

IDL files are not 2D MHD slices, but any output that uses the IDL format (MagGrid files, shell slices, 3D files, etc.).  It is ubiquitous.  Changelog is fine as written.  

The mag_grid[...].outs file is an ascii .outs file added in anticipation of expanding this capability to ascii files.  This file is definitely going to be used as a test point in the near future.  If you insist, I can amend the commit, but this file is definitely coming in a near-future commit.

> 3. My flake8 plugin is freaking out about these files... can you do a quick pass to clean up `pybats/__init__.py` and `pybats/bats.py`? At least hit some of the easier things like trailing whitespace and tabs? And maybe bring the new code into closer alignment with PEP8 (whitespace around operators, after commas, etc.)

I don't know what a flake8 plugin is or why it is a point of reference for Spacepy PRs.
Happy to do a PEP8 pass with a slight request: if you are speaking about those two files _in general_, I kindly ask that this pass wait until the PR associated with my _tests_ branch, which installs more PyBats tests and already has a bunch of PEP8 cleanup done.  If its specifically associated with the changes made in this PR, I can go back and do it now.

spacecataz:
Okay, gave it a _hard_ linting, more than I thought I would do.  Covered everything I touched plus more, but finishing this off is definitely a future PR.  Will include more in push on *tests* branch.

spacecataz:
Found an issue with AMR (grid is not constant) simulations- calculations are failing because array sizes aren't correct.  Arrays should be over-written all together; working on this now.

spacecataz:
Fixed issues, ready for review again!

drsteve:
The derived quantities don't appear to be  updating in binary `MagGridFile` objects when the frame is changed.

```
In [1]: from spacepy.pybats import bats

In [2]: test = bats.MagGridFile('mag_grid_e19890312-140000_19890314-005900.outs')

In [3]: test.calc_h()

In [4]: test['dBn'].max(), test['dBh'].max()
Out[4]: (74.86824418211852, 104.28978066243008)

In [5]: test.switch_frame(99)

In [6]: test['dBn'].max(), test['dBh'].max()
Out[6]: (137.61417470026848, 104.28978066243008)

In [7]: test.switch_frame(1053)

In [8]: test['dBn'].max(), test['dBh'].max()
Out[8]: (482.3769964040968, 104.28978066243008)
```

I assume the fix here is to decorate `calc_h` with `@calc_wrapper`?

On a similar note, on the `Bats2d` objects, the `calc_E` method isn't decorated and, I assume, would also not get updated on switching frames.

spacecataz:
> The derived quantities don't appear to be updating in binary `MagGridFile` objects when the frame is changed.
[...]
> I assume the fix here is to decorate `calc_h` with `@calc_wrapper`?

Thanks, this is a good catch.  Fixing now and pushing. 

> On a similar note, on the `Bats2d` objects, the `calc_E` method isn't decorated and, I assume, would also not get updated on switching frames.

Good catch.  There were a couple that needed decorating, including in the shell file objects, too.  Changed and updated.

spacepy/pybats/__init__.py
```diff
@@ -110,30 +110,60 @@ class methods with the *calc_* prefix and call them.
...
+# Pybats-related decorators:
+def calc_wrapper(meth):
+    '''
+    This is a decorator for `self.calc_*` object methods.  It establishes
+    `self._calcs`, a list of calcuation functions that have been called, and
+    adds the called function to that list to keep track of which calculations
+    have already been performed for the object.  If a calculation has already
+    been performed, it is skipped in order to reduce effort.
+    '''
+
```
drsteve, line 131: I know it's neither here nor there, but the whitespace lines surrounding the  `@wraps(meth)` make thing harder to read for me.

spacepy/pybats/__init__.py
```diff
@@ -554,42 +597,235 @@ def readarray(f, dtype=np.float32, inttype=np.int32):
...
+    Given a binary IDL-formmatted file opened as a file object, *f*, 
+    and whose file pointer is positioned at the start of an entry, 
+    skip to the end of the entry and return the total number of bytes skipped.
+
+    Fortran90+ binary files wrap entries with integers that state the number
+    of bytes of the entry.  This function reads the number of bytes before
+    the data entry, skips over the entry, and reads the trailing byte wrapper
+    before returning.
+
+    f : Binary file object
```
drsteve, line 642: Should this be in a "parameters" block?

spacepy/pybats/__init__.py
```diff
@@ -877,49 +1093,249 @@ class IdlFile(PbData):
...
+    these files, but only one data frame can be made available at a time.  This
+    prevents very large memory requirements.
+
+    These files are opened and handled similarly to regular IDL-formatted
+    files  with some important differences.  Upon instantiation, the user
+    may select which data frame to open with the *iframe* kwarg.  The default
+    action is to open the first frame.  The user may learn more about the
+    number of frames within the file and their associated epoch/iteration
+    information by examining the top level *attrs* (see below.)
+    The user may switch to any arbitrary frame using the `switch_frame(iframe)`
```
drsteve, line 1117: Just to note that I really would like to have the ability to iterate over frames. I'm happy to merge without, since there are additional updates coming and it could be argued to be an extra feature request...
But iteration would be super-handy, as would a simple `.next_frame()` or something to help navigation (maybe also `.prev_frame()`). Currently getting the next frame involves looking up the `iframe` in the attrs and passing that through to `switch_frame` which feels a bit messy. Again, noting as a _please put something like this in the upcoming set of revisions_.

spacepy/pybats/__init__.py
```diff
@@ -877,49 +1093,249 @@ class IdlFile(PbData):
...
+    | *_range              | The range of iterations/epochs covered in the file |
+    | ndim                 | Number of spatial dimensions covered by the data   |
+    | nframe               | The total number of data frames within the file    |
+    | iframe               | The current frame loaded (zero-based)              |
+    | format               | The format of the file, either binary or ascii     |
+    | header               | The raw string header of the file                  |
+
+    Notes
+    -----
+
```
drsteve, line 1146: Do blank lines here affect how the docs render? I'm not sure, but would probably remove it.

spacepy/pybats/__init__.py
```diff
@@ -877,49 +1093,249 @@ class IdlFile(PbData):
...
+            for name in calcs:
+                args, kwargs = calcs[name]
+                getattr(self, name)(*args, **kwargs)
+                
     def __repr__(self):
         return 'SWMF IDL-Binary file "%s"' % (self.attrs['file'])
-    
-    def read(self, header, keep_case):
+
+    def read(self, iframe=0):
```
drsteve, line 1318: While I'm thinking about feature requests to make this easier to use - it might be useful to be able to specify a time and get the nearest frame (or just have a convenience function for frame index lookup).

spacepy/pybats/bats.py
```diff
@@ -640,40 +646,122 @@ def trace(self, bats):
...
 
+
 class Bats2d(IdlFile):
     '''
-    A child class of :class:`~pybats.IdlFile` taylored to BATS-R-US output.
+    A child class of :class:`~pybats.IdlFile` tailored to 2D BATS-R-US output.
+
+    Calculations
+    ------------
+
```
drsteve, line 667: Same question about whether blank lines under section headings affect the rendering in docs

spacepy/pybats/bats.py
```diff
@@ -640,40 +646,122 @@ def trace(self, bats):
...
+    updated.
+
+    An exception to this is built-in `calc_*` methods, which perform common
+    MHD/fluid dynamic calculations (i.e., Alfven wave speed, vorticity, and
+    more.)  These values are updated when the data frame is switched (see the
+    `switch_frame` method).
+
+    Plotting
+    --------
+
```
drsteve, line 687: Same question about whether blank lines under section headings affect the rendering in docs

spacepy/pybats/bats.py
```diff
@@ -640,40 +646,122 @@ def trace(self, bats):
...
+    | add_contour        | Create a contour plot of a given variable      |
+    | add_pcolor         | Add a p-color (no-interpolation contour) plot  |
+    | add_stream_scatter | Scatter stream traces (any vector field)       |
+    | add_b_magsphere    | Add magnetic field lines for X-Z plane cuts    |
+    | add_planet         | Add a simple black/white planet at the origin  |
+    | add_body           | Add an inner boundary at the origin            |
+
+    Extracting and Stream Tracing
+    -----------------------------
+
```
drsteve, line 723: Same question about whether blank lines under section headings affect the rendering in docs

spacepy/pybats/bats.py
```diff
@@ -1073,27 +1167,26 @@ def _calc_Ekin(self, units='eV'):
...
-                                           self[s+'uy']**2+
-                                           self[s+'uz']**2)
+            self[s+'Ekin'] = dmarray(sqrt(self[s+'ux']**2 +
+                                          self[s+'uy']**2 +
+                                          self[s+'uz']**2)
                                      * conv * mass[s.lower()],
                                      attrs={'units':units})
 
-
+
```
drsteve, line 1189: Trailing whitespace

spacepy/pybats/__init__.py
spacecataz: Indeed; fixed.

spacepy/pybats/__init__.py
spacecataz: Indeed.  Fixed.

spacepy/pybats/__init__.py
spacecataz: These are all excellent suggestions.  Only hitch is ensuring we can still iterate over variables (a la dictionary keys/datamodel keys), which is used reasonably often enough.  I'm sure we can work around that, however.  The next and prev frame suggestions are great and are on my to-do list.

spacepy/pybats/__init__.py
spacecataz: Not sure, but removing whitespace to be consistent.  Same goes with suggestions below.

spacepy/pybats/__init__.py
spacecataz: Speaking from experience, the latter would be most useful.  Will take a bit of work/thought.  It's also a pandora's box to interpolating to an arbitrary time between two frames -- super useful and something I've considered, but for a later date.

spacepy/pybats/bats.py
spacecataz: That sneaky bugger.  Fixed.

spacepy/pybats/bats.py
drsteve: Was this one being left as-is or did you want to amend your last commit and force push?

spacecataz: Derp.
Fixed.

drsteve (approved):

PR #520, opened 2021-03-17
Handle leapseconds pre-1972

### LINKED ISSUE #393: Ticktock: Handle time when TAI second != UTC second
From 1960-1972, the UTC second was not equal in length to the TAI (and SI) second. This involved both fractional leap seconds and a UTC second being longer than the SI second. In Ticktock, we currently handle this by simply truncating TAI-UTC to the nearest integer wherever there's an update in the USNO tai-utc file, and maintain TAI second == SI second.

There are a few ways this could be handled:

1. The "right way", read the appropriate rate from the file and maintain a non-integral offset within this period.
2. Implement "stepped atomic time," which has more frequent fractional leapseconds but no rate adjustment.
3. Fudge the leap second data between 1960 and 1972, where we insert leap seconds according to the current rules to keep UTC-UT1 within 0.9s. Basically "playing IERS." This sounds terrible but it's actually not very different from what we're doing: some of the records in tai-utc result in no leap second treatment from Ticktock (pure rate changes and/or small fractional leaps), whereas some result in two (end of 1963: TAI-UTC is 3.2, previous record is 1.9, truncation leads to two leap seconds; see #390)
4. Document that if you want absolute time precision before 1972 *and* perfect conversion between UTC and TAI, you're barking up the wrong tree. Note that before 1960 UTC technically didn't exist so it's already somewhat hypothetical...we really don't want to handle "railroad time" in SpacePy.

### LINKED ISSUE #435: Deleting leapseconds file makes it impossible to download
This is hilarious...importing toolbox imports time, which reads the leapseconds file. So if the leapseconds file doesn't exist, importing toolbox to run `update` fails. Ooops.

This is probably going to require that time gracefully assumes no leapseconds and reloads itself after `update` downloads leapseconds.

```sh
rm ~/.spacepy/data/tai-utc.dat
python -e "import spacepy.toolbox; spacepy.toolbox.update(leapsecs=True)"
```

### Error message/Traceback:

```
  File "<string>", line 1, in <module>
  File "/home/jtniehof/.local//lib//python2.7/site-packages/spacepy/toolbox/__init__.py", line 67, in <module>
    from spacepy import time as spt
  File "/home/jtniehof/.local//lib//python2.7/site-packages/spacepy/time.py", line 2573, in <module>
    _read_leaps()
  File "/home/jtniehof/.local//lib//python2.7/site-packages/spacepy/time.py", line 2259, in _read_leaps
    with open(fname) as fh:
IOError: [Errno 2] No such file or directory: '/home/jtniehof/.spacepy/data/tai-utc.dat
```

```
Linux-4.4.0-98-generic-x86_64-with-Ubuntu-16.04-xenial
sys.version_info(major=2, minor=7, micro=12, releaselevel='final', serial=0)
numpy=1.11.0
scipy=0.17.0
matplotlib=1.5.1
```

### Version of SpacePy
git master (568c0df6a38aabf800bf371adda28471ff4177e4)

jtniehof:
This PR applies the post-1972 leapsecond rules from 1958 to 1972 and thus closes #393. Background discussion of the problem is in that issue. It also gracefully handles a missing leapsecond file and closes #435.

## PR Checklist

The fix for #435 isn't easily unit-testable so it was confirmed by hand.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -6,6 +6,7 @@ tests
  - Add assertWarns and assertDoesntWarn context managers.
 time
  - Fix warning for out-of-date leapseconds.
+ - Apply post-1972 leapsecond rules to pre-1972.
 toolbox
  - Fix bootHisto passing of kwargs to histogram and bar.
 pycdf
--- a/Doc/source/release_notes.rst
+++ b/Doc/source/release_notes.rst
@@ -30,6 +30,12 @@ to :func:`numpy.histogram` and :func:`matplotlib.pyplot.bar` has been fixed.
 The check for out-of-date leapseconds in :mod:`~spacepy.time` has been
 fixed (previously warned even when the file was up to date.)
 
+Other changes
+*************
+Modern leapsecond rules are applied from 1958-1972 rather than
+rounding fractional leapseconds. See :mod:`~spacepy.time` for full
+discussion of leap seconds and other conversion considerations.
+
 0.2.2 (2020-12-29)
 ------------------
 
--- a/developer/scripts/rubber_seconds.py
+++ b/developer/scripts/rubber_seconds.py
@@ -1,0 +1,154 @@
+#!/usr/bin/env python
+
+"""Calculate various approaches to UTC - TAI during rubber second era"""
+
+import datetime
+
+import matplotlib
+matplotlib.use('Agg')
+import matplotlib.pyplot as plt
+import numpy
+import spacepy.time
+
+# These are values from USNO tai-utc during rubber-second era
+# Piece-wise linear, calculated as a constant plus a rate times
+# days since a particular day (so seconds/day). Days are UTC days
+# The rate can change on 1 Jan and the constant can change any month;
+# the days-since day is relative to the last rate change, not the
+# last record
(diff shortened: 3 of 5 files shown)


jtniehof:
I just made a commit that adds the TAI-UTC from the CDF library to the plot. At this point there are too many lines on the plot so I recommend anybody playing with this just comment out the bits they're not interested in. The upshot is that CDF exactly matches the actual rubber-second UTC line. I honestly don't know...maybe we should do this right, but it's going to involve another rearchitecture of time, hopefully not as drastic as before.

drsteve:
What's the status here? Is this waiting for my review of the "back-dated" leap seconds rule? Or are we intending to match the "rubber seconds"? In the latter case I guess we can either incorporate this as a step on the way to the rubber second, or we can put the PR on hold until rubber seconds are done. In the former case, we just incorporate this...
I've totally forgotten where the discussion had got to though, and I apologize for dropping this.

jtniehof:
I'd say let's call this good enough for now, assuming you're happy with the back-dated leap second approach as applied. It's definitely better than what we have, fixes another bug, and maybe we don't care about rubber.

drsteve (approved):

PR #539, opened 2021-07-20
Fix bug in pybats.ram.PressureFile that gives zero electron pressure

drsteve:
The `pybats.ram.PressureFile` reader erroneously assumes that if electron pressures are present the variable is listed in the file header as `PPAR_e` or `PPER_e`, where the species is actually given in upper case (`PPAR_E`).

The current reader makes empty arrays (zero-filled) for all expected species. The case error means that the electrons aren't detected as present so the arrays are all zero. This also means that the total pressure is read from the wrong column and is incorrect.

The present pull request fixes this issue and generalizes the reader so that it will correctly read whichever species are present (RAM now allows a specified list of named species). Test files are added to allow testing the reader. The PR also removes all trailing whitespace in files touched by the PR.


spacepy/pybats/ram.py
```diff
@@ -1158,99 +1158,92 @@ def __init__(self, filename, *args, **kwargs):
...
-        f = open(filename, 'r')
-        lines = f.readlines()
-        f.close()
+        with open(filename, 'r') as fh:
+            lines = fh.readlines()
+        head_entries = lines[1].strip().split()
+        variables = head_entries[:-1]
+        p_unit = head_entries[-1].replace('[', '').replace(']', '')
+
+        def name_map(namein):
```
spacecataz, line 1167: Small request: Include, either in comments or docstring, example of an entry to be converted.  Seeing that in-line is super helpful when returning to this code to debug an edge case or new case.

spacecataz (changes requested):
Great update to clean up some *very old* and rather ugly bits of code along with a helpful fix.
Only request is to expand the docstring on the internal function to parse variables; these types of functions have a way of running into issues as output files change. 

Very small request; everything else is fine and can be merged.

spacecataz (approved):
*chef kiss*

PR #541, opened 2021-08-11
Fix contour plot method (add_cont_press) in pybats.ram.PressureFile

drsteve:
The `spacepy.pybats.ram.PressureFile.add_cont_press` method was failing as the plot routines called internally make boolean masks from the array passed in. This is one of the cases where dmarray will not behave correctly (broken numpy subclass behavior that's not trapped by our wrapper), so the plot fails.

So this PR:
1. Adds basic tests for the `add_cont_press` and `add_pcol_press` methods to ensure they'll actually make plots without erroring. It also tests the return types.
2. Makes a couple of fixes for new matplotlib warnings
  - `zmin` and `zmax` kwargs are moved into the norm for the `add_pcol_press` plot call.
  - `np.int` is a deprecated alias for `int` and this was changed on line 1915 of `pybats/__init__.py`
3. Trailing whitespace stripped from `pybats/__init__.py`
4. I haven't added a `CHANGELOG` entry as I think this was broken between releases, so shouldn't need to be in the release notes.

- [N/A] Added an entry to CHANGELOG if fixing a major bug or providing a major new feature

### DIFF
--- a/spacepy/pybats/__init__.py
+++ b/spacepy/pybats/__init__.py
@@ -3,15 +3,15 @@
 PyBats!  An open source Python-based interface for reading, manipulating,
 and visualizing BATS-R-US and SWMF output.
 For more information on the SWMF, please visit the
-`Center for Space Environment Modeling <http://csem.engin.umich.edu>`_. 
+`Center for Space Environment Modeling <http://csem.engin.umich.edu>`_.
 
 Introduction
 ------------
 
 At its most fundamental level, PyBats provides access to output files written
-by the Space Weather Modeling Framework and the codes contained within.  
+by the Space Weather Modeling Framework and the codes contained within.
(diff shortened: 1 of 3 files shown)


jtniehof:
I checked and the new tests pass on 0.2.2, and this PR, and on master. Maybe this broke with certain versions of numpy or Python?

jtniehof:
On python 3.9, numpy 1.19.5, master, test_pcolor_plot works and test_contour_plot fails. However, both work on python 3.9, numpy 1.19.5, and 0.2.2. So this does look like what you fixed was indeed a regression in between releases, and also version-dependent. I would agree that doesn't need a changelog.

jtniehof (approved):
Looks good to me. Only question is whether it's time to move the pyplot import to the top of test_pybats since it's pulled in by so many tests now.

PR #553, opened 2021-09-16
Update CI to 20.04 (Closes #550)

### LINKED ISSUE #550: Ubuntu 16.04 to be removed from GH Actions (impacts tests for 3.5)
We just got hit by one of the "brown outs" for Ubuntu 16.04 on the GH actions test runner. https://github.com/spacepy/spacepy/actions/runs/1233777812

Basically, Ubuntu 16.04 is being removed from GH actions (https://github.com/actions/virtual-environments/issues/3287) on September 20th. Both of our Python 3.5 builds are on 16.04 images.

[Our dependency chart](https://spacepy.github.io/dep_versions.html) suggests we could drop explicit 3.5 testing, which I'd be happy enough with. But we do need to figure out what our solution is fairly quickly...

jtniehof:
This PR updates all our CI jobs to use Ubuntu 20.04, and closes #550.

The first commit is all that really matters: Python 3.5 was added to the GHA environment on Ubuntu 18.04, and so all that was needed was to move to 18.04 instead of 16.04.

That, however, got me down a deep rabbit hole, much of which has been rebased away. 18.04 also supported Python 3.4 in GHA, so I added that to the CI. Finally, I moved the CI to 20.04, which works fine except it doesn't support 3.4, so that was dropped. I could have kept it on 18.04, I guess, but this seemed like more fun.

A lot of the flailing that was rebased away was trying to use the deadsnakes PPA so we had that in our back pocket for CI on older versions. It's actually really hard to use directly in GHA. There's an action for using it directly, but it only works on 20.04, and the deadsnakes PPA only provides up to 3.5 on 20.04. Despite the name, it's now more oriented towards newer Pythons instead of older.

I should also note...we still have the OS specified for each entry in the matrix, because the cacheing is OS dependent, and it's actually hard to get that information otherwise.

## PR Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,85 +30,73 @@ jobs:
       matrix:
         include:
           - python-version: 2.7
-            os: ubuntu-18.04
+            os: ubuntu-20.04
             numpy-version: ">=1.10.0,<1.11.0"
             piplist: "scipy>=0.11.0,<0.12.0 matplotlib>=1.5.0,<1.6.0 networkx>=1.0,<1.1 h5py>=2.6,<2.7 ffnet>=0.7.0,<0.8 astropy>=1.0,<1.1"
             cflags: "-Wno-error=format-security"
             dep-strategy: "oldest"
           - python-version: 2.7
-            os: ubuntu-18.04
+            os: ubuntu-20.04
             numpy-version: ">=1.16.0,<1.17.0"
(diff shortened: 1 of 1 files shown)


drsteve (approved):
I've been trying to think of a compelling reason to keep 18.04 in the test matrix and I just can't. This is pretty much the minimal change to close #550 and hopefully push out the horizon on similar issues until after we've dropped support for older versions.

PR #556, opened 2021-09-29
pybats/ram: Fix fallback pitch angle binning in RamSat

drsteve:
## The Issue
It was pointed out to me that there were missing parentheses in the fallback calculation of `dMu` for calculating omnidirectional flux if the `pa_width` array is not present in the loaded `RamSat` object.
```
dMu[i] = 4*np.pi*self['pa_grid'][i] -self['pa_grid'][i-1]
```
where the _width_ is the subtraction...

This code, because it's a fallback calculation, shouldn't be called by someone using `RamSat` to work with RAM output satellite files. That is, the `pa_width` should always be present. However, if it's there then it should work, so...

## The Updates
- The original fallback calculation didn't actually populate the `pa_width` array, so this now does.
- The original implementation only multiplied indices [1, end] by 4pi (missing
index 0), so would be incorrect even with the parentheses.
- The update removes a loop in the fallback and simplifies the logic check to see whether this needs calling.
- I added comments to document the assumptions of the fallback calculation
- There's now a test to hit the new code:
  - Uses the recalculated `pa_width`, so will error if it's not stored
  - Checks that the new `pa_width` is consistent with that in the test file
  - Check for similarity of derived omni flux using new widths (noting that this won't be the same because the fallback calculation doesn't have enough info to ensure identical bin widths to what should have been in the file)
- I did a sweep of the full `RamSat` class to deal with PEP8 stuff (whitespace, `x not in y` vs. `not x in y`, etc.)

- [N/A] Added an entry to CHANGELOG if fixing a major bug or providing a major new feature

### DIFF
--- a/spacepy/pybats/ram.py
+++ b/spacepy/pybats/ram.py
@@ -616,14 +616,14 @@ class RamSat(SpaceData):
         omnikeys = [('omni{0}'.format(get_species_label(kk)), kk)
                     for kk in self if get_species_label(kk) is not None]
         # Create delta mu, where mu = cos(pitch angle)
-        dMu = np.zeros(nPa)
         if 'pa_width' in self:
             dMu = 4*np.pi*self['pa_width']
         else:
-            dMu[0] = self['pa_grid'][1]
-            for i in range(1, nPa):
-                # Factor of 4*pi here so we don't need it later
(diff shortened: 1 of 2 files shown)


spacecataz (approved):
Looks good; great PEP 8 changes, ready to pull.

PR #568, opened 2021-11-01
Remove previously-deprecated code

### LINKED ISSUE #336: Remove deprecated pycdf.AttrList.from_dict
`from_dict` was deprecated in 0.1.5 and should be removed at next minor (not subminor) release.

### LINKED ISSUE #337: Remove deprecated pybats.ram.read_ram_dst
`read_ram_dst` was deprecated in 0.1.6 and should be removed in next minor (not subminor) release, unless @spacecataz has comments?

### LINKED ISSUE #358: spectrogram class hides spectrogram module
`plot.spectrogram` contains a class `spectrogram`. However since top-level `plot` does `from .spectrogram import *` it's impossible to get at the `plot.spectrogram` module. This makes the autodocs not work correctly.

My preference would be to not bring everything into the `plot` namespace, but that's not the smallest possible change to fix this issue. Otherwise we need to make the names not clash, and I can see two options:

1. Rename the `spectrogram` module to `spect` (or similar).
2. Rename the `spectrogram` class to `Spectrogram`.

The advantage of 1 is that it doesn't change the current behavior. 2 is more consistent with convention (classes are usually capitalized) but it's just not going to be possible to have lowercase `spectrogram` refer to both the class and the module for a transition period. If we do 2, my suggestion would be to add the `Spectrogram` name for the class to both `plot` and `plot.spectrogram` and then make any instantiation of `spectrogram` raise a DeprecationWarning so people know to change the caps. Then on a later release we can remove the `spectrogram` class name, exposing the `spectrogram` module.

I'm bringing this up now so we can do the deprecation in 0.2.2 and possibly make the change in 0.3.0, so then I can start working on getting the docs happier.

jtniehof:
This PR removes a lot of code that has been deprecated in 0.1 or 0.2, as well as some similar behavior changes that need to be handled as a migration.

 * Switches main change documentation from CHANGELOG to the release notes
 * Remove deprecated `pycdf.AttrList.from_dict` (Closes #336).
 * Make v3 CDFs and TIME_TT2000 the default in CDFs (#198 ...this does not completely close, but completes the 0.3.x portion)
 * Remove deprecated pybats `read_ram_dst` (Closes #337) and `fix_format`.
 * Remove lower-case alias for `Spectrogram` (Closes #358).
 * Remove deprecated colourmaps.
 * Remove deprecated `toolbox.feq`.
 * Remove quaternion functions from `toolbox` (they're all in `coordinates`).

## PR Checklist
- [X] (See below) Added an entry to CHANGELOG if fixing a major bug or providing a major new feature

The PR template still has CHANGELOG; that's changed in here to reference the release notes. The release notes are updated in this PR.


spacecataz (approved):
Looks good to me- esp. the pybats removals, which I'm happy to see.

Def. get feedback from others to make sure!

balarsen:
Thanks for being on top of these. Looks good to me also.

PR #575, opened 2022-02-18
Added sort_unstructured_data keyword, defaulting to False. 

lrastaet:
This optional keyword is needed to disable sorting in _read_idl_bin in a case when the full 3D grid is being read and the block structure needs to be preserved.
This will change the behavior of _read_idl_bin() where sorting was done all the time. 
Sorting in _read_idl_ascii is a new option but has not been tested.
To preserve the original behavior the default may be set to True and the new sorting code removed from _read_idl_ascii().

## PR Checklist
Please try to follow as much of this checklist as you can for
Some additional suggestions are given below.

Go ahead and replace the text above (and this line!) with your
pull request description.

### DIFF
--- a/spacepy/pybats/__init__.py
+++ b/spacepy/pybats/__init__.py
@@ -392,7 +392,7 @@ def add_body(ax, rad=2.5, facecolor='lightgrey', show_planet=True,
     ax.add_artist(body)
 
 
-def _read_idl_ascii(pbdat, header='units', start_loc=0, keep_case=True):
+def _read_idl_ascii(pbdat, header='units', start_loc=0, keep_case=True, sort_unstructured_data=False):
     '''
     Load a SWMF IDL ascii output file and load into a pre-existing PbData
     object.  This should only be called by :class:`IdlFile`.
@@ -539,7 +539,18 @@ def _read_idl_ascii(pbdat, header='units', start_loc=0, keep_case=True):
                 pbdat[v] = dmarray(np.reshape(pbdat[v], pbdat['grid'],
                                               order='F'), attrs=pbdat[v].attrs)
 
-
+    # Unstructured data can be in any order, so let's sort it.
+    if gtyp == 'Unstructured' and sort_unstructured_data:
+        gridtotal = np.zeros(npts)
+        offset = 0.0  # The offset ensures no repeating vals while sorting.
+        for key in pbdat['grid'].attrs['dims']:
+            gridtotal = gridtotal + offset + pbdat[key]
+            offset = offset + np.pi/2.0
+            SortIndex = np.argsort(gridtotal)
+        for key in list(pbdat.keys()):
+            if key=='grid': continue
+            pbdat[key] = pbdat[key][SortIndex]
+                
 def readarray(f, dtype=np.float32, inttype=np.int32):
     '''
     Read an array from an unformatted binary file written out by a
@@ -819,7 +830,7 @@ def _probe_idlfile(filename):
 
 
 def _read_idl_bin(pbdat, header='units', start_loc=0, keep_case=True,
-                  headeronly=False):
+                  headeronly=False, sort_unstructured_data=False):
     '''
     Load a SWMF IDL binary output file and load into a pre-existing PbData
     object.  This should only be called by :class:`IdlFile`, which will
@@ -983,7 +994,7 @@ def _read_idl_bin(pbdat, header='units', start_loc=0, keep_case=True,
                 pbdat[names[i]] = pbdat[names[i]].reshape(pbdat['grid'], order='F')
 
         # Unstructured data can be in any order, so let's sort it.
-        if gtyp == 'Unstructured':
+        if gtyp == 'Unstructured' and sort_unstructured_data:
             gridtotal = np.zeros(npts)
             offset = 0.0  # The offset ensures no repeating vals while sorting.
             for key in pbdat['grid'].attrs['dims']:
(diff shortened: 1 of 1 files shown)


drsteve (changes requested):
Thanks for the PR @lrastaet!
First, we appreciate the contribution and this seems like it'll be a great extension to the current `pybats` capability. I do have a few comments and requests though, and I imagine others (@spacecataz, probably), will have some further comments too.

- Can you explain the use case a little more clearly? I understand your one-sentence description about preserving block structure, but that doesn't say _why_ preserving block structure is important. My guess is that it's for calculating the connectivity so the 3D output can be formatted for VTK-based (or similar) tools? It'd be really helpful to have some extra information about what the code change is enabling and/or fixing.
- It doesn't really matter from our end, but I note that you've developed on `master`. After merge the commit(s) would probably have different hashes (because we use a rebase strategy), so unless you're comfortable rolling back your repo history before you pull from upstream again then I'd suggest moving this into a feature branch and doing the PR from that.
- The PR changes default behavior, and that's generally not advised. Anything that changes default behavior should initially preserve the default behavior but raise a deprecation warning (unless there's a good reason not to). Then we can change the default behavior a release later. This way no users get surprised by any unintended consequences.
- Unless the old behavior is not desired (and/or a bug, which I don't think is the case here), then changing between the current behavior and the new behavior should _all_ be switchable, and your PR says it's not (although a read through the code suggests it's all switchable). Can you make sure that's all in alignment (and if it is, then just clarify in the PR so I don't misinterpret).
- There's no test for the new code. You might be able to contrive a test using one of the BATS IDL files in `/tests/data/pybats_test`, but if not then a *minimal* test data set needs to be constructed (please keep file size as small as possible). What should be tested here? I'd probably start with ensuring that the old behavior is encoded in a test, and then also testing that the expected block structure is preserved with the new kwarg enabled. Line 545 of `pybats/__init__.py` also suggests that there's code to ensure no duplicate data points, so there should probably be a test to show that.
- Line 1180 has a typo (_postions_) and should probably also note that setting the flag to `False` preserves the block ordering (since this is the aim of the PR).
- Line 1318 needs a space after the second comma
- This should probably also have an entry in the release notes. The PR checklist should also be filled out (which you can either do with `x`s in the markdown, or just click on the boxes without going in to edit mode)
- Finally, for general reference, I'd like to note that we have some documentation for developers at https://spacepy.github.io/index.html#spacepy-documents (and it's probably in need of some work, so if you have suggestions or think something is missing, please let us know)

We're always happy to have conversations about any of this, and if there's anything you wanted help with do let us know.

lrastaet:
@drsteve 

Thanks for the review. 

The preservation of the block structure is currently used by a BATSRUS model reader I am developing for Kamodo (https://github.com/nasa/kamodo/ -- a package of readers and analysis software for space science models). 
The BATSRUS reader/wrapper for Kamodo turns BATSRUS magnetosphere model outputs into functions and uses an interpolator that takes advantage of the grid structure to enable data-model comparisons and visualizations.

I plan to rework the PR as suggested in your comments. 
I certainly will have questions as I work through the steps.
What is the best way to submit questions?

I guess this PR should be closed and when I have the new request ready I could refer to this one if that is desired.

spacecataz:
Some quick notes:

1) The tests pass, so I'm comfortable with not adding new tests at this point.  Old behavior is tested already, so I think we're safe there.  A better test case will be to implement a 3d file test that relies on the original output file ordering. I have a 3d file branch that will include this in the feature, so I'm okay forgoing new tests at this point.

We could quickly add a test that uses the existing files, however.  It wouldn't be particularly enlightening, but it would exercise the functionality.  @lrastaet, if you create a new PR, I can add that test pretty quickly and save us both some time.

2) Changing the default behavior as @lrastaet has done I believe is the right thing to do; just need a better description and an update in the changelog and PR description.

3) Regarding creating a new PR: if you plan to keep working from master on your fork, we don't have to do this- we can just update the PR and associated files no problem.  However, if you want to move this to a development branch, a new PR is probably the correct way to move forward.

drsteve:
Now that github is back up...

- Discussion is probably best kept in the PR so we can all chip in and it's available for reference later. If you do make a new PR then we can have any additional discussion there
- If you want to move to a feature branch then you'd need a new PR, and in that case I'd suggest making a new branch from master (which will include your commit here) then resetting your master to the previous commit.
- If you're happy to continue in `master` and fix that later then no need for a new PR
- The fact that the tests pass with a change in behavior means that the sort is a) not important to the tests _including however the data are used in tests_ and b) is not tested for. So the sort should be tested for. This should be easy though.
- Since the code doesn't officially support 3D (in practice it does, but it's not exploited in any way) it's probably okay to leave testing for the 3D case for now.
- While the unsorted read might be a better default, and it does pass our current (not-very-exhaustive) tests, I'm still uncomfortable with an immediate change to default read behavior. Anyone who is reading files currently could well be relying on the sorted data so a change without a deprecation could break analyses without warning (this is probably unlikely, but it's the general principle). I will still recommend having the sort be the default but adding a deprecation warning - anyone who wants the new behavior will be able to supply a flag, which isn't a burden. Then after our next release we can switch the default to unsorted and remove the deprecation warning (and this won't break code using the new flag). I'm happy to hear reasons for pushback on this.

lrastaet:
I created a new branch and have reworked the proposed modification tot eh code.
I ran the test_pybats.py myself on a clone of that branch and all tests but one succeed (assert(lib/have_libspacepy) fails).
How do I build the library after installing the local clone in development mode (pip install -e spacepy/)?

drsteve:
> How do I build the library after installing the local clone in development mode (pip install -e spacepy/)?

I don't use pip for this, so I'd have to look at documentation to understand why the library isn't being either properly built, or installed (whatever the underlying cause of the test failure is). If you've cloned your fork of the repo and want to build, I'd just run `python setup.py install` (with `--user` if you aren't in a virtual environment _a la_ conda).

PR #589, opened 2022-04-01
Add cache-expiration to CI

jtniehof:
CI is currently failing due to AstroPy and I think it's because there was a glitch at exactly the time our dependencies were rebuilt for the week. There's no way to force-expire the dependency cache to check that. This PR modifies the cache keys to include a reference to a repository-wide secret. We can change that secret to force-expire all dependency caches.

While I was diagnosing this I also found that `test_coordinates` imports `spacepy_testing` (which sets up the pathing) after `spacepy.coordinates`. This meant the `coordinates` import would not come from the build directory. This wasn't showing up in CI because `test_all` fixes the path immediately, but I fixed that for testing `coordinates` on its own.

## PR Checklist
- [X] (N/A)Any new modules, functions or classes have docstrings consistent with SpacePy style
- [X] (N/A)Added an entry to release notes if fixing a significant bug or providing a new feature
- [X] (N/A)New features and bug fixes should have unit tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,19 +125,23 @@ jobs:
       shell: bash
     - name: Cache pip
       uses: actions/cache@v2
+      env: # Need to tickle this to get it to unlock secrets at all
+        CACHE_VERSION: ${{ secrets.CACHE_VERSION }}
       with:
         path: ~/.cache/pip
         # Force-expire the pip cache weekly
-        key: pip-v${{ env.pip-cache-version}}-${{ steps.get-week.outputs.week }}-${{ matrix.os }}-${{ matrix.python-version }}-${{ matrix.dep-strategy }}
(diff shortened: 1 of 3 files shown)


jtniehof:
Looks like, in addition to this not fixing the AstroPy problem, the secret is not passing through--the cache isn't expiring when the secret is changed.

jtniehof:
It looks like this may be a fool's errand, because secrets don't get passed to pull requests (not a surprise), and there's no way to set a non-secret environment variable. Supposedly PRs by maintainers are supposed to work, but I'm guessing it doesn't because none of us have permission to push directly...

jtniehof:
I am leaving the original description for historical purposes, but because using the secrets for cache expiration doesn't work, I have removed that functionality. So this PR only fixes the pathing for `test_coordinates` and documents how the cacheing works.

I didn't rebase out the commit that tried to set up the secret-based cache expiration, because I suspect we might want to come back to it, and even if not it's useful to have a record of what didn't work. But I can certainly change that.

The original problem, that astropy blew up, has fixed itself. I still feel we're in a bit of a fragile spot there but it would be tricky to get right (caching on that specifically might help, but might also hide problems..and it's hard to do a "cache but update").

drsteve (approved):

PR #594, opened 2022-04-22
pycdf enhancements: UTF-8, truthiness, document slow readonly closing

### LINKED ISSUE #555: pycdf fails with non-unicode characters in CDF attribute
Loading ARASE ephemeris CDFs with `spacepy.datamodel.fromCDF` fails with a `UnicodeDecodeError`.
This arises because `fromCDF` is reading _everything_ at once, but the actual error is in `pycdf`.

Basically, there is a non-unicode byte sequence at the end of one of the metadata entries. There's no failsafe on conversion (e.g., a replace mode, or a skip-if-invalid), so the attribute read fails.

```
wget https://ergsc.isee.nagoya-u.ac.jp/data/ergsc/satellite/erg/orb/def/2017/erg_orb_l2_20170908_v03.cdf
```
then in python
```
import spacepy.pycdf as cdf
data = cdf.CDF('erg_orb_l2_20170908_v03.cdf')
data.attrs
```

### Error message/Traceback:
```
Python 3.8.2 (default, Mar 26 2020, 15:53:00) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.26.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import spacepy.pycdf as cdf
/home/smorley/miniconda3/envs/default/lib/python3.8/site-packages/spacepy/time.py:2365: UserWarning: Leapseconds may be out of date. Use spacepy.toolbox.update(leapsecs=True)
  warnings.warn('Leapseconds may be out of date.'

In [2]: data = cdf.CDF('erg_orb_l2_20170908_v03.cdf')

In [3]: data.attrs
Out[3]: ---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
~/miniconda3/envs/default/lib/python3.8/site-packages/IPython/core/formatters.py in __call__(self, obj)
    700                 type_pprinters=self.type_printers,
    701                 deferred_pprinters=self.deferred_printers)
--> 702             printer.pretty(obj)
    703             printer.flush()
    704             return stream.getvalue()

~/miniconda3/envs/default/lib/python3.8/site-packages/IPython/lib/pretty.py in pretty(self, obj)
    392                         if cls is not object \
    393                                 and callable(cls.__dict__.get('__repr__')):
--> 394                             return _repr_pprint(obj, self, cycle)
    395 
    396             return _default_pprint(obj, self, cycle)

~/miniconda3/envs/default/lib/python3.8/site-packages/IPython/lib/pretty.py in _repr_pprint(obj, p, cycle)
    698     """A pprint that just redirects to the normal repr function."""
    699     # Find newlines and replace them with p.break_()
--> 700     output = repr(obj)
    701     lines = output.splitlines()
    702     with p.group():

~/miniconda3/envs/default/lib/python3.8/site-packages/spacepy/pycdf/__init__.py in __repr__(self)
   5580         @rtype: str
   5581         """
-> 5582         return '<' + self.__class__.__name__ + ':\n' + str(self) + '\n>'
   5583 
   5584     def __str__(self):

~/miniconda3/envs/default/lib/python3.8/site-packages/spacepy/pycdf/__init__.py in __str__(self)
   5592         """
   5593         if self._cdf_file._opened:
-> 5594             return '\n'.join([key + ': ' + (
   5595                 ('\n' + ' ' * (len(key) + 2)).join(
   5596                 [str(value[i]) + ' [' + lib.cdftypenames[value.type(i)] + ']'

~/miniconda3/envs/default/lib/python3.8/site-packages/spacepy/pycdf/__init__.py in <listcomp>(.0)
   5594             return '\n'.join([key + ': ' + (
   5595                 ('\n' + ' ' * (len(key) + 2)).join(
-> 5596                 [str(value[i]) + ' [' + lib.cdftypenames[value.type(i)] + ']'
   5597                  for i in range(value.max_idx() + 1) if value.has_entry(i)])
   5598                 if isinstance(value, Attr)

~/miniconda3/envs/default/lib/python3.8/site-packages/spacepy/pycdf/__init__.py in <listcomp>(.0)
   5594             return '\n'.join([key + ': ' + (
   5595                 ('\n' + ' ' * (len(key) + 2)).join(
-> 5596                 [str(value[i]) + ' [' + lib.cdftypenames[value.type(i)] + ']'
   5597                  for i in range(value.max_idx() + 1) if value.has_entry(i)])
   5598                 if isinstance(value, Attr)

~/miniconda3/envs/default/lib/python3.8/site-packages/spacepy/pycdf/__init__.py in __getitem__(self, key)
   4751         else:
   4752             if self.has_entry(key):
-> 4753                 return self._get_entry(key)
   4754             else:
   4755                 raise IndexError('list index ' + str(key) + ' out of range.')

~/miniconda3/envs/default/lib/python3.8/site-packages/spacepy/pycdf/__init__.py in _get_entry(self, number)
   5230                 result = bytes(buff)
   5231             else: #Py3k, make unicode
-> 5232                 result = str(numpy.char.array(buff).decode())
   5233         else:
   5234             if not self._raw:

~/.local/lib/python3.8/site-packages/numpy/core/defchararray.py in decode(self, encoding, errors)
   2185 
   2186         """
-> 2187         return decode(self, encoding, errors)
   2188 
   2189     def encode(self, encoding=None, errors=None):

<__array_function__ internals> in decode(*args, **kwargs)

~/.local/lib/python3.8/site-packages/numpy/core/defchararray.py in decode(a, encoding, errors)
    555     """
    556     return _to_string_or_unicode_array(
--> 557         _vec_string(a, object_, 'decode', _clean_args(encoding, errors)))
    558 
    559 

UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa1 in position 407: invalid start byte
```

Linux-4.19.128-microsoft-standard-x86_64-with-glibc2.10
sys.version_info(major=3, minor=8, micro=2, releaselevel='final', serial=0)
numpy=1.21.2
scipy=1.6.2
matplotlib=3.4.3
CDF=3.8.0.1

(note that this was my reproduction of someone else's bug report - all I know about their system is Python 3.9, Mac, SpacePy from github fairly recently)

### Version of SpacePy
HEAD of master from repo

### LINKED ISSUE #560: Define __bool__ for pycdf.Var and/or document truthiness
pycdf doesn't explicitly define "truthiness" for a variable, so I'm often a little more verbose than I should be to see if records are written. I expected `Var` to be always truthy, but it isn't....maybe because it subclasses `MutableSequence` and that might call `len` for us?

I need to check that this is robust and maybe defining `__bool__` (or `__nonzero__` for Py2, but let's not talk about that) would be more efficient/easier. This is a cool enough trick it's also worth documenting once I'm sure it's robust (and thus testing!)

```python
import spacepy.pycdf
f = spacepy.pycdf.CDF('foo.cdf', create=True)
v = f.new('variable', data=[], type=spacepy.pycdf.const.CDF_INT1)
print('full' if v else 'empty')  # empty, but I expected full
print('full' if len(v) else 'empty')  # empty, as expected
```

```
Linux-4.15.0-159-generic-x86_64-with-Ubuntu-18.04-bionic
sys.version_info(major=2, minor=7, micro=17, releaselevel='final', serial=0)
numpy=1.13.3
scipy=0.19.1
matplotlib=2.1.1
```

### Version of SpacePy
Current master of github.

### LINKED ISSUE #586: Support UTF-8 in pycdf CHAR/UCHAR values
NASA CDF C library 3.8.1 adds support for UTF-8 in variable records and Entries of type CDF_CHAR/CDF_UCHAR. This is primarily in the tools, not in the base library. Names (of attributes and variables) must still be in ASCII.

With raw variables, pycdf does no translation between bytes and strings. With "normal" variables and Python 3, the bytes are decoded to strings. This currently assumes ASCII, and failure raises a UnicodeDecodeError.

Proposed changes are:

- Set the default encoding to utf-8 instead of ascii, based on the assumption that people will now start making CDFs with UTF-8. This will not be CDF library dependent (a user still may have, say, 3.7, and there's no reason not to open newer files).
- Instead of raising an exception on failure to decode, use the "replacement" strategy, which puts Unicode fffd "Replacement Character" in the string (usually renders as a funky question mark).
- EDIT: Replaced with next option ~~Provide an "encoding" argument to the CDF constructor, to be used on both read and write. I don't _like_ doing this on write, it's more important people be able to read files in other encodings (e.g. I think Arase has some), but probably best for symmetry. Note that bytes (vs. strings) will still be written without any change. So it would be possible for "encoding" to only apply on read and assume, when writing, that everything should go UTF-8, and if the user doesn't like that they can encode the bytes themselves: `cdf.attr['Entry'][0] = entry.encode('shift-jis')`~~
- EDIT: Provide an "encoding" argument to the CDF constructor, to be used on both read and write. This allows people to read files in other encodings (e.g. I think Arase has some) and will maintain symmetry between read and write. However, as soon as a CDF is open read-write, and the encoding is not UTF-8, issue a warning of using a nonstandard encoding, but still use that encoding on writes. If the user wants to use a different encoding on write than on read, bytes are still written without change. So they can encode the bytes themselves: `cdf.attr['Entry'][0] = entry.encode('shift-jis')`. Warning is issued when switch to read-write, not on every write.
- EDIT: Continue to raise UnicodeEncodeError on _write_ if a unicode (str) value is passed in which cannot be encoded in the file's encoding. However, do not check if bytes are passed (i.e. do not try to encode the bytes just to make sure they're valid, if you pass bytes, they go directly).

jtniehof:
This PR lumps together several small pycdf improvements:

- In connection with the new NASA CDF interpretation, support for UTF-8 encoding and documentation of how the encoding works. In addition to switching to utf-8 instead of ascii encoding, this also changes the behavior on decoder fail: silent replacement instead of raising exception. Closes #555. Closes #586.
- Tests truthiness of variables and documents this trick. Closes #560.
- Updates the documentation on slow closing of readonly CDFs, since 3.8.1 fixed this.

## PR Checklist

### DIFF
--- a/Doc/source/pycdf.rst
+++ b/Doc/source/pycdf.rst
@@ -1,28 +1,16 @@
-
 ######################################
 pycdf - Python interface to CDF files
 ######################################
 
 .. automodule:: spacepy.pycdf
 
-
 Contents
 --------
 
-- `Quickstart`_
-    - `Create a CDF`_
-    - `Read a CDF`_
-    - `Modify a CDF`_
-    - `Non record-varying`_
-    - `Slicing and indexing`_
-    - `String handling`_
-- `Classes`_
-- `Functions`_
-- `Submodules`_
-- `Data`_
+.. contents::
+   :depth: 2
+   :local:
 
-Quickstart
-----------
 Create a CDF
 ============
 This example presents the entire sequence of creating a CDF and populating
@@ -302,7 +290,7 @@ The underlying C library is represented by the :attr:`~spacepy.pycdf.lib`
 variable.
 
 Classes
--------
+=======
 
 .. autosummary::
     :template: clean_class.rst
@@ -325,7 +313,7 @@ Classes
     EpochError
 
 Functions
----------
+=========
 
 .. autosummary::
     :template: clean_function.rst
@@ -334,7 +322,7 @@ Functions
     concatCDF
 
 Submodules
-----------
+==========
 
 .. autosummary::
     :toctree: autosummary  
@@ -344,7 +332,7 @@ Submodules
     istp
 
 Data
-----
+====
 
 .. attribute:: lib
 
--- a/Doc/source/pycdf.rst
+++ b/Doc/source/pycdf.rst
@@ -15,6 +15,7 @@ Contents
     - `Modify a CDF`_
     - `Non record-varying`_
     - `Slicing and indexing`_
+    - `String handling`_
 - `Classes`_
 - `Functions`_
 - `Submodules`_
@@ -248,6 +249,49 @@ hourly data file created in earlier examples.
 >>> cdf.close()
 
 The :class:`Var` documentation has several additional examples.
+
+.. _pycdf_string_handling:
+
+String handling
+===============
+
+   .. versionchanged:: 0.3.0
+
+   Prior to SpacePy 0.3.0, pycdf treated all strings as ASCII-encoded, and
+   would raise errors when writing or reading strings that were not valid
+   ASCII.
+
+Per the NASA CDF library, variable and attribute names must be in
+ASCII. The contents of ``CDF_CHAR`` and ``CDF_UCHAR`` were redefined
+to be UTF-8 as of CDF 3.8.1. As of SpacePy 0.3.0, pycdf treats all
+``CHAR`` variables with a default encoding of UTF-8. This is true
+regardless of the version of the underlying CDF library.
+
+UTF-8 is a variable-length encoding, so the number of elements in the
+variable may not correspond to the number of characters if data are
+not restricted to the ASCII range.
+
+A different encoding can be specified with the ``encoding`` argument
+to :class:`~spacepy.pycdf.CDF.open` and this encoding will be used on
+all reads and writes to that file. Opening a CDF read-write with
+``encoding`` other than ``utf-8`` or ``ascii`` will issue a warning.
+
+Writing strings which cannot be represented in the desired encoding
+will raise an error. When reading from a CDF, characters which cannot
+be decoded will be replaced with the Unicode "replacement character"
+U+FFFD, which usually displays as a question mark.
+
+It is always possible to write raw bytes data to a variable, if it is
+desired to use a different encoding for one time. For arrays of data,
+this will usually involve :func:`numpy.char.encode`:
+
+   >>> cdf['Variable'] = data.encode('latin-1')
+   >>> cdf['Variable'] = numpy.char.encode(data, encoding='latin-1')
+
+All encoding and decoding can also be skipped using the
+:meth:`~spacepy.pycdf.CDF.raw_var` method to access a variable;
+however, without encoding, only :class:`bytes` can be written to
+string variables.
 
 Access to CDF constants and the C library
 =========================================
--- a/Doc/source/release_notes.rst
+++ b/Doc/source/release_notes.rst
@@ -25,6 +25,11 @@ built at installation time. The default backend remains IRBEM; in
 backend. The new :mod:`~spacepy.igrf` module is part of this support
 but may be of interest on its own.
 
+In accordance with a change from NASA, :mod:`~spacepy.pycdf` now
+assumes strings in CDFs are UTF-8. It will no longer raise errors on
+reading non-ASCII data from a CDF. See :ref:`pycdf_string_handling` in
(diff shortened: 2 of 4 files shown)


jtniehof:
I did hand-test the fix of #555 but chose not to include that specific file as a unit test. The tests do include reading from an attribute that isn't valid UTF-8.

drsteve (approved):

Issue #472, opened 2021-01-04
Consider vectorising LANLstar calculations


aaronhendry:
The current calculation of LANLstar values for a large number of input points/epochs is rather more inefficient than it needs to be. Specifically, the for-loop in the _LANLcommon, from LANLstar.py:

https://github.com/spacepy/spacepy/blob/76c5a35c688e7876292ce05f3bf833a793d77fd6/spacepy/LANLstar.py#L144-L148

The actual neural-network calculation that takes place under the ffnet hood in this loop is a simple series of matrix/vector operations, but there is a significant level of overhead involved in each call to ffnet. It is possible to extract the various weights and biases needed to manually calculate the ANN value from the ffnet pickle object, and then convert the for-loop above into a series of numpy vector operations. This results in significantly increased throughput for the LANLstar functions -- it also has the added benefit of removing the soft-dependency on ffnet+networkx.

I have a working proof-of-concept for this idea, which accurately reproduces the results from ffnet, however unfortunately it is currently written in MATLAB script (for various reasons it was easier for me to do this at the time in MATLAB than modify SpacePy to do this). It shouldn't be too arduous to convert it to work within SpacePy, but I guess the first question is whether this is even something that the SpacePy maintainers are interested in?

jtniehof:
Removing the dependency on ffnet? That alone has me interested!

drsteve:
Hi @aaronhendry 
Thanks for that. Yes, we're interested and porting your MatLab proof-of-concept will be much faster than starting the process from scratch, so thanks for your effort!

Just as a historical note, the dependence on `ffnet` is because that was (back in '09) pretty much the best package for training feed-forward neural networks in Python, and more importantly because very few people in space physics used Python (at the time) `ffnet` would export the trained network to a fortran subroutine so it was easy to distribute and call from other languages.
So the "least work" to include it in SpacePy once we became well-established and folks were actually using Python, was to just use ffnet to load/call.

Since there's no external development of LANLstar any more, and any new version is likely to be done from scratch using a more modern framework, it's high time that we simplified how it's included in SpacePy. Just pulling out the weights and solving directly will be faster, easier to maintain, and will simplify distribution (`networkx` is easy as a dependency, but `ffnet` isn't always easy)

[...]

PR #598, opened 2022-04-25
Mark LANLstar is unsupported and don't install ffnet/networkx

jtniehof:
ffnet will build, but not import, for Python 3.7 and later. We didn't notice this problem before because the CI `test_all` script didn't include `test_lanlstar`. This doesn't seem worth putting any more effort into, rather we should continue with #472 to make LANLstar function without ffnet.

This PR, then:

 * Removes most references to ffnet and networkx
 * Does not test LANLstar for python > 3.6 (but does test for <= 3.6 in CI)
 * Notes lack of support in release notes
 * Removes links to LANLstar docs from other docs

The underlying model is that `ffnet` will not be coming back, but `LANLstar` probably will, so most references to ffnet and networkx are purged, and LANLstar is made somewhat hidden.

## PR Checklist


drsteve (approved):
Given the spot we're in in this one this is probably the right thing to do, but it does feel awkward. Still, the clean up is appropriate. I probably wouldn't have removed the LANLstar page from the docs, but I won't fight it. I guess this means I need to prioritize the refactor to evaluate the NN as matrix operations using numpy.

jtniehof:
> I probably wouldn't have removed the LANLstar page from the docs, but I won't fight it.

Without ffnet, the module won't import and the docs won't build. This actually happened [with the last release](https://spacepy.github.io/lanlstar.html) but I didn't notice--I'm honestly not sure the LANLstar docs have ever built properly. It didn't seem worth a lot of effort to try and make documentation work for a module we're not currently supporting (try to figure workarounds, etc.)

I build the docs for release with latest of everything, so including those docs would require (at least) rolling back to an earlier Python and possibly earlier Sphinx, and that makes the website diff (and change to the look of the website) pretty jarring. Or maybe big workarounds and again...work.

drsteve:
> Without ffnet, the module won't import and the docs won't build. This actually happened [with the last release](https://spacepy.github.io/lanlstar.html) but I didn't notice--I'm honestly not sure the LANLstar docs have ever built properly. 

I'm convinced!

PR #614, opened 2022-06-10
pycdf 64-bit ARM (M1, 64-bit Pi) support

### LINKED ISSUE #198: pycdf: make v3 CDFs and TT2000 the default

### LINKED ISSUE #546: pycdf argument types for M1 Mac
The M1 Mac uses a different convention for variadic functions and the resulting interaction with ctypes is...weird. The result is that any call to the CDF library `CDFlib` function fails.

This is described in [Python#42880](https://bugs.python.org/issue42880) (which also has some stuff that might be useful for general ctypes troubleshooting on M1). The upshot appears to be that the first argument *must* have a type specified. There are a few other pure-variadic functions, notably EPOCHbreakdown. This may be more complicated, as the first argument is a double. I may have to type-pun this function like in #20 (which reminds me to also look at CDF_TT2000_from_UTC_parts).

```python
import spacepy.pycdf
```

### Error message/Traceback:
See #545, but that's complicated by the actual error not being thrown. What should really get thrown is `CDFError: BAD_FNC_OR_ITEM: The specified function or item is illegal.`

Python 3.9, M1 Mac

### Version of SpacePy
Current `master`

jtniehof:
This PR update the C function argument handling for pycdf, which fixes 64-bit ARM (Apple Silicon/M1 Mac, but also 64-bit RaspPi). Closes #546.

  * Types of fixed arguments are always specified; types of non-variadic arguments are never specified. ctypes is smart enough to treat specified arguments as fixed and the rest as variadic (the typepunning is still required on 32-bit ARM though).
  * Argument/return value handling is generalized to make this easier to change in the future.
  * Where function arguments are specified, a lot of explicit casts have been removed, relying instead on the ctypes automatic casting. This should make the code easier to read.

In order to do this, I wrote a very simple script that just tests loading and calling the C library. The full test suite isn't very informative when the problem is that the C library isn't being called correctly, so that script is included.

Since things were being reworked to support the different setup for the C function arguments, I also removed some deprecated code and warnings: the `_nelems` method of Var (replaced by public `nelems`) and the warnings about new defaults for backward compatibility and time type. This closes #198.

pycdf unit test suite has been run on:
- [X] Intel Linux
- [X] 32-bit Pi
- [X] 64-bit Pi
- [X] Intel Mac
- [X] Apple Silicon Mac (thanks @spacecataz)
- [X] Windows

## PR Checklist

### DIFF
--- a/Doc/source/release_notes.rst
+++ b/Doc/source/release_notes.rst
@@ -9,6 +9,20 @@ This document presents user-visible changes in each release of SpacePy.
    :local:
 
 
+0.4 Series
+==========
+0.4.0 (2022-xx-xx)
+------------------
+
+Other changes
+*************
+:mod:`~spacepy.pycdf` no longer warns when defaulting to version 3 CDFs
+and TIME_TT2000 time type if not specified; the warning was added in
+0.2.2 and the default changed in 0.3.0. Use
+:meth:`~spacepy.pycdf.Library.set_backward` to create version 2 CDFs and
+explicitly specify a time type (e.g. with :meth:`~spacepy.pycdf.CDF.new`)
+if TT2000 is not desired.
+
 0.3 Series
 ==========
 0.3.0 (2022-04-27)
--- a/Doc/source/release_notes.rst
+++ b/Doc/source/release_notes.rst
@@ -20,6 +20,11 @@ The ``_nelems`` method of :class:`~spacepy.pycdf.Var` has been removed;
 use the public interface :meth:`~spacepy.pycdf.Var.nelems`. (Deprecated
 in 0.2.2).
 
+Major bugfixes
+**************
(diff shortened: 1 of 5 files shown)


spacecataz (approved):
This all looks good and works as expected on an M1 mac.

drsteve (approved):
Looks fine to me. I ran an additional build on Windows (Win11, conda w/ 3.8) and pycdf is working great.

PR #623, opened 2022-06-29
datamodel to/fromHDF5 improvements (gzip scalars, unicode strings)

jtniehof:
This PR makes three changes to the `toHDF5` and `fronHDF5` functions in `datamodel`:

1. Does not attempt to compress scalar variables if compression is requested. Previously, the entire toHDF5 function would fail if a single element of the `SpaceData` were scalar.
2. Supports writing Unicode arrays, i.e. Python 3 strings, and also reading them as such.
3. Some light refactoring/cleanup. Mostly this is consolidating lines but I also changed the `import h5py as hdf` to just `import h5py`...it doesn't look like the renaming is a convention anywhere else, and it confused the heck out of me.

The Unicode writing is a bit messy, but it maintains the fallback to "just try it as a 35-element array of bytes" and passes all the tests...this code is somewhat messy anyhow, alas. Another wrinkle is that Unicode variables are inherently variable-length in HDF5, whereas numpy has fixed-length Unicode string arrays (U) or can permit variable-length by using object arrays. This code will, on read, always convert to a fixed-length Unicode string array (which will be as long as the longest string).

All this conversion isn't great from a performance perspective, but it's a function of Python, and thus numpy, using UCS-2 or USC-4, whereas HDF5 uses UTF-8.

The new tests demonstrate what failed before.

## PR Checklist

I didn't add anything to the release notes, because this felt like very inside baseball. I did note the change to the compress argument. I didn't see an obvious place to note the Unicode array writing; maybe that one does belong in release notes?

### DIFF
--- a/spacepy/datamodel.py
+++ b/spacepy/datamodel.py
@@ -984,46 +984,38 @@ def fromHDF5(fname, **kwargs):
                                     'value = {0} ({1})'.format(value, type(value)), DMWarning)
 
     try:
-        import h5py as hdf
+        import h5py
     except ImportError:
         raise ImportError('HDF5 converter requires h5py')
 
     if type(fname) in str_classes:
-        hfile = hdf.File(fname, mode='r')
+        hfile = h5py.File(fname, mode='r')
     else:
         hfile = fname
         #should test here for HDF file object
-
-    if 'path' not in kwargs:
-        path = '/'
-    else:
-        path = kwargs['path']
+    path = kwargs.get('path', '/')
 
     SDobject = SpaceData()
-    allowed_elems = [hdf.Group, hdf.Dataset]
+    allowed_elems = [h5py.Group, h5py.Dataset]
     ##carry over the attributes
     hdfcarryattrs(SDobject, hfile, path)
     ##carry over the groups and datasets
     for key, value in hfile[path].items():
-        #try:
-            if type(value) is allowed_elems[0]: #if a group
-                SDobject[key] = SpaceData()
-                SDobject[key] = fromHDF5(hfile, path=path+'/'+key)
-            elif type(value) is allowed_elems[1]: #if a dataset
-                isuni = (hdf.check_vlen_dtype(value.dtype)  # h5py 3+
-                         if hasattr(hdf, 'check_vlen_dtype')
-                         else hdf.check_dtype(vlen=value.dtype)) is unicode
-                try:
-                    if isuni:
-                        if hasattr(value, 'asstr'):  # h5py 3+
-                            value = value.asstr()
-                        value = numpy.require(value[...], dtype=unicode)
-                    SDobject[key] = dmarray(value)
-                except (TypeError, ZeroDivisionError): #ZeroDivisionError catches zero-sized DataSets
-                    SDobject[key] = dmarray(None)
-                hdfcarryattrs(SDobject[key], hfile, path+'/'+key)
-        #except:
-        #    raise ValueError('HDF5 file contains type other than Group or Dataset')
+        if type(value) is allowed_elems[0]: #if a group
+            SDobject[key] = fromHDF5(hfile, path=path+'/'+key)
+        elif type(value) is allowed_elems[1]: #if a dataset
+            isuni = (h5py.check_vlen_dtype(value.dtype)  # h5py 3+
+                     if hasattr(h5py, 'check_vlen_dtype')
+                     else h5py.check_dtype(vlen=value.dtype)) is unicode
+            try:
+                if isuni:
+                    if hasattr(value, 'asstr'):  # h5py 3+
+                        value = value.asstr()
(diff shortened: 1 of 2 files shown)


jtniehof:
Test failures are due to something that changed between HDF5 2.10 and 3.1. It looks like a failure to decode UTF-8 strings on read although oddly enough they wind up in object arrays, not bytes arrays.

jtniehof:
Trying to isolate if this is a problem in writing or in reading, I tried writing and reading the same file with different versions of h5py. This is not promising.

| | Write 2.10 | Write 3.1 |
|--- | --- | --- |
| Read 2.10 | Success | Fail |
| Read 3.1 | Fail | Fail |

h5dump from 2.10:
```
HDF5 "foo.h5" {
GROUP "/" {
   DATASET "dat" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_UTF8;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SIMPLE { ( 2 ) / ( 2 ) }
      DATA {
      (0): "foo", "bar"
      }
   }
   DATASET "scalar" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_UTF8;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SCALAR
      DATA {
      (0): "foo"
      }
   }
}
}
```

h5dump from 3.1:
```
HDF5 "foo.h5" {
GROUP "/" {
   DATASET "dat" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_UTF8;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SIMPLE { ( 2 ) / ( 2 ) }
      DATA {
      (0): "foo", "bar"
      }
   }
   DATASET "scalar" {
      DATATYPE  H5T_STRING {
         STRSIZE H5T_VARIABLE;
         STRPAD H5T_STR_NULLTERM;
         CSET H5T_CSET_UTF8;
         CTYPE H5T_C_S1;
      }
      DATASPACE  SCALAR
      DATA {
      (0): "foo"
      }
   }
}
}
```

The HDF5 from 2.10 and 3.1 are, at least, different (i.e. `diff` shows not the same), although identical length in bytes.

jtniehof:
`check_vlen_dtype` when run on either the 2.10 or 3.1 output succeeds:
```python
>>> h5py.check_vlen_dtype(hfile['/']['dat'].dtype)
<class 'str'>
```

It looks like h5py v3 requires being explicit if reading strings as strings: https://docs.h5py.org/en/stable/high/dataset.html#h5py.Dataset.asstr

So we probably have to change the reader to detect string types and explicitly decode on read for h5py 3+. Unfortunately even using `asstr` returns an object array, not U.

I'm still confused how writing in 3.1 makes it not readable in 2.10, if the read is what changed.

jtniehof:
Clearing up a little confusion: on 2.10, the read succeeds in that we don't get arrays of bytes type and the test passes. However, the dtype of the array is still object, not U. This threw me in the initial comparisons because I was looking briefly at the dtype not the types of its contents.

I do think we should do the cast back over to numpy U if reading UTF-8 string arrays. I think byte arrays still work fine. So there's more work to do here, sigh.

jtniehof:
Unicode issues have been resolved and the main description updated to reflect that.

spacepy/datamodel.py
```diff
@@ -1174,13 +1165,24 @@ def SDcarryattrs(SDobject, hfile, path, allowed_attrs):
                 hfile[path].create_group(key)
                 toHDF5(hfile, SDobject[key], path=path+'/'+key, compression=h5_compr_type, compression_opts=h5_compr_opts)
             elif isinstance(value, allowed_elems[1]):
+                comptype, compopts = (None, None) if value.shape is ()\
```
drsteve, line 1168: Checking whether the shape `is ()` raises a syntax warning (`"is" with a literal. Did you mean "=="?`)

If `value` is a scalar then `len(value.shape)` will be zero. If `value` is an empty array, then len still returns non-zero. Should this be checking the len (or similar)?

drsteve (changes requested):
Looks good with the exception of using `value.shape is ()`. That evaluates correctly - at least it does for now - but gives a syntax warning (and also warns on install).

spacepy/datamodel.py
jtniehof: It should be `==` not `is`...I'm not sure why I keep on doing that. Not looking for an empty array, looking for an array scalar, which will always have a shape of an empty tuple.

drsteve (approved):

PR #624, opened 2022-06-30
Support VarBundle output to SpaceData

jtniehof:
The VarBundle class takes a variable from a CDF file, optionally performs operations on it, and writes it to a new CDF. This PR updates to allow output to a SpaceData object instead of just a CDF.

Most of the nastiness is in the tests, and the diffs look worse than they are--basically the existing tests are refactored into two classes, one of which runs with CDF output, and one which runs with SpaceData output.

## PR Checklist


spacecataz (approved):
Looks good- like the overloading of output() and the subsequent wrapping as a solution for this.

PR #631, opened 2022-07-14
Update IRBEM library version to include recent bugfixes

### LINKED ISSUE #577: Coordinates: Nonsense value returned in conversion to GDZ spherical coordinates
Another example for an error resulting from a conversion _to_ GDZ spherical _from_ GEO spherical
```py
cvals = Coords([[1.063871355909377, -90.0, -180.0], [1.063871355909377, 0.0, 0.0], [1.063871355909377, 90.0, 180.0]], 'GEO', 'sph')
tt = Ticktock(ts_to_spacepystr([1068865200 1068865200 1068865200]), 'ISO')
cvals.ticks = tt
newC = cvals.convert('GDZ','sph')
```
output for newC:
Coords( [[-13119.519017475252, -90.0, -180.0], [400.00018276982473, 3.530478404585919e-15, 0.0], [406.0143894752504, 90.0, 180.0]] , 'GDZ', 'sph')
Note the same times and radii are given for each coordinate set, but different latitudes and longitudes are used. The first value returned for altitude in km is obviously nonsensical. The same value is returned when a longitude of 180 is used instead of -180 (equivalent position). This causes errors in code layers built on top of this in Kamodo (https://github.com/nasa/Kamodo).

Request: a layer be added to check for and correct these types of errors.

This also begs the question of why the returned altitudes differ at all for the same time and radius with differing latitude and longitudes. My initial guess is that the Earth is assumed to be non-spherical?

drsteve:
This PR includes the upstream changes to IRBEM (see PRBEM/IRBEM#32) to close #577 and fix some other coordinate system transform errors (and to just bring us back in sync with IRBEM).

Additional related changes:
- TEME is now fully supported by both coordinate system backends
- Testing for IRBEM-based transforms has been expanded (one key test added here tests all input/output combinations of natively Cartesian systems and checks that a roundtrip yields the expected result), as well as TEME consistency between IRBEM/SpacePy/astropy

Now PRBEM/IRBEM#32 has been merged ~there are a few things to do~:
- ~update the IRBEM source directory name to reflect the date/commit hash~
- ~update `setup.py` to capture the new IRBEM source directory name~
- update release notes

PR Checklist:


drsteve:
I've marked as ready for review with the caveat that the release notes haven't been updated. This will probably need conflicts resolving as soon as #605 is merged, so ideally I'd do the release notes after the conflicts (to reduce the number of conflicts).

tests/test_coordinates.py
```diff
@@ -524,6 +525,26 @@ def test_roundtrip_GEO_GEI(self):
...
+        for insys in irbframes:
+            for outsys in irbframes:
+                expected = spc.Coords([1, 2, 4], insys, 'car',
+                                      use_irbem=True, ticks=tt)
+                stage1 = expected.convert(outsys, 'car')
+                got = stage1.convert(insys, 'car')
+                np.testing.assert_allclose(got.data, expected.data, atol=0.001)
+                if insys in spc.SYS_EQUIV:
+                    options = [insys, spc.SYS_EQUIV[insys]]
+                    assert insys in options
```
jtniehof, line 544: It's not clear what this test is doing. I can see not doing the assert_equal is the input is in our equivalent list (i.e. test just the base system). But wouldn't the assert always pass? If it's meant to be part of the test, using assertIn might give a more useful failure message...but I'm not sure what this means, so grasping at straws here.

tests/test_coordinates.py
```diff
@@ -524,6 +525,26 @@ def test_roundtrip_GEO_GEI(self):
...
+        # exclude frames either not supported by IRBEM or that aren't cartesian
+        for key in ['ECIMOD', 'GDZ', 'RLL', 'CDMAG', 'SPH']:
+            irbframes.pop(irbframes.index(key))
+        for insys in irbframes:
+            for outsys in irbframes:
+                expected = spc.Coords([1, 2, 4], insys, 'car',
+                                      use_irbem=True, ticks=tt)
+                stage1 = expected.convert(outsys, 'car')
+                got = stage1.convert(insys, 'car')
+                np.testing.assert_allclose(got.data, expected.data, atol=0.001)
```
jtniehof, line 541: Here (and on line 546), using err_msg and verbose would give more information about what in/out combination failed--otherwise it's going to be really opaque.

jtniehof (commented):
Just getting ahead of the game here. Good enhancements; I'm not going to bother checking the irbem library itself. I've got two questions about the tests though.

tests/test_coordinates.py
drsteve: Typo... it should've been `assert got.dtype in options`

The context was the `if`/`else` block. It was supposed to be testing the system name was correct after the roundtrip. Because there are some "equivalent" names (i.e., GEI and TOD in IRBEM are both ECITOD in SpacePy backend; IRBEM's SPH is just spherical GEO and so we use GEO (and SPH doesn't exist in ctrans)) the options list and the assert was supposed to check that the ouput system name was either the input name or the system equivalent name.

tests/test_coordinates.py
drsteve: Do we have an example for that or do I have to burn brain cycles and think about it?

tests/test_coordinates.py
jtniehof: Any problem with `self.assertIn(got.dtype, options)` then? Tends to give a better error message on failure.

tests/test_coordinates.py
jtniehof: Try this for line 541:

```python
np.testing.assert_allclose(got.data, expected.data, atol=0.001, verbose=True,
    err_msg='insys: {} outsys: {}'.format(insys, outsys))
```

And for line 546:
```python
np.testing.assert_equal(got.dtype, insys, verbose=True,
    err_msg='insys: {} outsys: {}'.format(insys, outsys))
```

drsteve: Ah, gotcha. That looks like a worthwhile update. Will push shortly.

jtniehof (approved):
Looks good, thanks!

PR #632, opened 2022-07-14
Fixes and code improvements in pybats.ram 

drsteve:
This PR will make sure that the `spacepy.pybats.ram` functionality is up-to-date and compatible with output files from the current version of RAM.

- Fix time parsing in E field files to handle fractional seconds _if present_.
- Updated `LogFile` reader to handle milliseconds and properly read the last line of the log
- `IonoPotScb` updated
  - Now inherits from `PbData` instead of `object`
  - PyNIO dependence replaced with `scipy.io` netcdf reader
- Plot routines work with recent spacepy and matplotlib
  - `EfieldFile.add_potplot` fixed so it plots with recent spacepy and matplotlib
  - `LogFile.add_dst_quicklook` fixed to remove duplicate axis handling leading to bad tick placement
  - `GeoMltFile.plot_epoch_flux` extended to give user control over plot title
- `GeoMltFile` updated to use context manager for file open/read and enabled `__add__` to go with the existing `__iadd__` method for appending data files
- Lots of PEP8 linting cleanup, namespacing, more compact use of datamodel, etc.

Checklist:


drsteve:
I've expanded testing here, and while it's not comprehensive, it's still an improvement in coverage. Part of the issue is that some of the file types are hard to make minimal versions of, and adding larger files to the test data directory isn't ideal.

Also, I haven't added anything to release notes because it didn't feel like anything here rose to that level. Maybe it needs a quick sentence, but this is mostly updates that are user-transparent (unless they're on new versions of matplotlib, in which case things will work again).

jtniehof (commented):
FWIW, coverage went from 28% to 33%, so going to call that a win. Everything looks good to me.

I'm horrified that we had a hidden pynio dependency.

If the latest RAM output doesn't work without this, I think that's worth a release note--I'd put that in "new features" (is RAM versioned so it can be expressed what version it works with?) and also a note in "dependency requirements" that we don't need pynio anymore.

I will leave formal approval to @spacecataz

drsteve:
Supported RAM file types haven't changed in a way that breaks anything. It depends whether fractional second support (rather than just ignoring them) is worth adding to release notes.

Also, there's still one file type (`Currents`, not currently supported/produced by RAM-SCB) that uses PyNIO to read in there. But I can't make a test file without archaeology, so I just swept it under the rug for now.

jtniehof:
I expect fractional seconds probably won't significantly change people's results, will leave to @spacecataz . The pynio thing....we probably should mention. It might also be good to document the requirement more clearly in the Currents docstring (not essential to this PR)...I read "For quick parsing of these files, PyNIO is required" to mean "without PyNIO parsing will be slow" not "it'll fail."

spacecataz (approved):
This is great- lots of helpful improvements, linting, and overall upkeep.

PR #634, opened 2022-08-19
Support VarBundle input from SpaceData objects

jtniehof:
This PR supports VarBundle operations on SpaceData objects (it's the complement of #624). There's still a bit of CDF-specific sugar where using CDF inputs or outputs, but from the user's perspective a CDF with ISTP metadata and a SpaceData with ISTP metadata will be treated basically the same. Over time I expect to move the ISTP stuff into datamodel and make this all a lot smoother...this is a baby step.

I also pulled out some numpy 1.6 code while I was at it.

## PR Checklist

### DIFF
--- a/Doc/source/release_notes.rst
+++ b/Doc/source/release_notes.rst
@@ -27,7 +27,7 @@ temporary removal of support for this module in SpacePy 0.3.0 has therefore
 been lifted. The new implementation provides a slight performance increase
 with no change in results or accuracy.
 
-:class:`~spacepy.pycdf.istp.VarBundle` now supports output to
+:class:`~spacepy.pycdf.istp.VarBundle` now supports output to and input from
 :class:`~spacepy.datamodel.SpaceData` objects as well as
 :class:`~spacepy.pycdf.CDF`.
 
--- a/spacepy/pycdf/istp.py
+++ b/spacepy/pycdf/istp.py
@@ -1175,13 +1175,23 @@ class VarBundle(object):
     chaining operations, or to load just the relevant data from a CDF
     into a :class:`~spacepy.datamodel.SpaceData`.
 
+    ``VarBundle`` operates on a single variable within a file and
+    its various dependencies, uncertainties, labels, etc. That variable
+    can be specified one of two ways. A :class:`~spacepy.pycdf.Var`
+    can be passed as the only parameter, which implicitly defines
+    the input file (the CDF containing that variable). Or, the file
+    can be passed as the first parameter, and the name of a variable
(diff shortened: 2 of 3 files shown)


drsteve (dismissed):
I haven't managed to break it, and it looks pretty good _and useful_ to me.

The one thing I would suggest could be improved is the error message when trying `VarBundle(data['varname'])` where `data` is a `SpaceData` instead of a CDF. Currently it raises an exception with
```AttributeError: 'SpaceData' object has no attribute 'cdf_file'```
but I'm not sure how well that communicates what the user did wrong.

However, as this is an initial move on the way to having support based in datamodel I'm not sure it _needs_ fixing.

jtniehof:
That's weird...that error should complain about ``dmarray`` having no such attribute, not ``SpaceData``. Can you provide a fuller example? I could possibly require the ``source`` to be a ``Var`` if using the single-parameter form, but that feels overly restrictive if someone does something else with the same pycdf-like API. The user can always throw weird stuff at the function and I'm reluctant to do too much in the way of guessing games....

drsteve:
I was trying various (object, entry) combinations, including a datamodel representation of a (nested) HDF5 file. That's probably why it says `SpaceData` instead of `dmarray`. However, seeing `dmarray` in the same error message still doesn't tell me what I did wrong...
I also don't like the requirement for `source` to be a `Var` when using the single-parameter form. I was thinking it should explicitly raise an error saying that if the single-parameter form is used then the object passed in has to be CDF or CDF-like... (where it currently raises an error that is related, but may not make clear what the user should've done)

jtniehof:
Single-parameter has to be a variable, not a CDF...because if you pass in a CDF, you don't know which variable in that form. It's a very weird setup because the contained object (the Var) has the reference to the container (the CDF), whereas a SpaceData the container has the reference to the contained (as in dicts).

It's just a question of how much it's worth trapping one error to raise another, which tries to guess what the user is up to. You can do bizarre stuff like passing in a string and an integer and will probably get other weird errors. I guess do you expect to be able to pass in a single dmarray and have it work? The cdf_file is just the first thing to trip over in that case, and something else might show up.

The question of what to do with more deeply nested structures, supported by SpaceData and HDF5 but not CDF (and not ISTP) makes this worse...that's definitely a matter for another day.

jtniehof:
The documentation didn't make that clear, so that's updated. I'm not thrilled about the error message, but we're accepting imperfect here and it's probably less likely to lead to problems, so I put it in.

drsteve (approved):

PR #639, opened 2022-08-26
Mac installer and documentation updates

### LINKED ISSUE #638: pip builds against different numpy than installed.
From #604 

On install, pip installs numpy in a separate environment to do the build. That can be a newer version of numpy (1.23, in my test case) than conda has installed (1.21), so the built module won't import in that environment. This will manifest as a warning of a numpy API mismatch, followed by a failure to import numpy.core.multiarray. Note 1.21 is API version 0x0e and 1.23 is API version 0x10.

This will only happen when installing with pip instead of setup.py, so there's a workaround. Another is to explicitly do pip install numpy before installing spacepy, which is of course a pain.

If I remove the numpy requirement from pyproject.toml, the build fails because pip is still using an isolated environment. Same if I have an empty pyproject.toml. Completely removing pyproject.toml fixes the problem, but that's not going to work with the migration to the New Way Of Installing Things.

Bottom line: pip makes no attempt to check if the version used to build a package is the same version that it will be installed under, even though it does check that the requirement exists.

I don't think this is Mac specific, it's just a question of what's available in conda vs. pip on any particular platform.

jtniehof:
This PR makes some smaller installer changes to make life easier on the Mac, but more importantly it completely rewrites the Mac documentation to reflect the current (onerous) process of getting a working environment on the Mac.

Part of this is changing the build requirement to the minimum version of numpy available, thus closes #638 (which manifested more frequently on Mac.) This also addresses the ``System`` linking, which is related to #583 but I'm not ready to close it based just on this PR.

Other changes made in the course of this (largely coming about from testing install on various ARM platforms):

 * pycdf searches more widely on Mac. In addition to looking in /Applications/cdf-*version*, it will look in /Applications/cdf/cdf-*version*. It will also look in both locations relative to the user's home directory, as this is where the "install only for me" option puts the distribution.
 * The irbem build has always just raised a warning on failure and moved on, which results in a message printed to console. The only case where this didn't happen is if there was no Fortran compiler installed at all. This case is now caught and treated just like any other irbem failure to build. Note that pip eats these warnings, so people with no Fortran compiler installing via pip would silently wind up with a broken irbem. This is, honestly, the situation in so many cases anyhow, that "just giving a working SpacePy without a Fortran compiler" seems better, particularly with coordinates and lanlstar no longer requiring Fortran.

To test pip installs of this branch, use:
```
pip install git+https://github.com/jtniehof/spacepy.git@mac_install
```

Or, on conda with the correct SDK in place:
```
SDKROOT=/opt/MacOSX10.9.sdk pip install git+https://github.com/jtniehof/spacepy.git@mac_install
```

## PR Checklist

### DIFF
--- a/Doc/source/install.rst
+++ b/Doc/source/install.rst
@@ -39,6 +39,9 @@ If you need further assistance, you can `open an issue
     install_mac
     install_windows
 
+.. contents::
+   :local:
+
 SpacePy installs with the common Python distutils and pip.
 
 The latest stable release is provided via `PyPI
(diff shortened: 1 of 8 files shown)


drsteve (commented):
Read-through looks good. I'll leave final review and merge to @spacecataz as the designated mac tester...

jtniehof:
Rebased just to bring up to date with main.

spacecataz (approved):
Looks good!

PR #643, opened 2022-09-06
Installer-related fixes: Windows dll names; document pip cache clearing

jtniehof:
This PR fixes two issues found in testing of 0.4.0rc1.

First, it documents clearing the pip cache. The pip cache is stored at the user level, so it doesn't separate conda and macports builds. This causes problems if mixing them, which is another manifestation of our irbem "circular import" bug.

Second, something in the Windows build chain has changed the name of shared libraries since the 0.3.0 release. Previously libspacepy was `spacepy.dll`. Now, on Python 3.6 it's still `spacepy.dll`; on 3.7+, it's `libspacepy.dll.a`. I don't know what changed; the same version of mingw32 (currently 5.3.0) is used for all current builds, so the difference isn't there. This comes from `compiler.shared_lib_extension` in distutils. The solution is to update `lib.py` to look for the `libspacepy.dll.a` name as well as `spacepy.dll`. Otherwise libspacepy doesn't load.

### DIFF
--- a/spacepy/lib.py
+++ b/spacepy/lib.py
@@ -98,13 +98,11 @@ def load_lib():
     @rtype: ctypes.CDLL or ctypes.WinDLL
     """
     libdir = os.path.dirname(os.path.abspath(__file__))
-    if sys.platform == 'win32':
-        libnames = ['spacepy.dll', 'libspacepy.dll.a']
-    elif sys.platform == 'darwin':
-        libnames = ['libspacepy.dylib', 'libspacepy.so',
-                  'spacepy.dylib', 'spacepy.so']
-    else:
-        libnames = ['libspacepy.so']
+    libnames = {
+        'win32': ['libspacepy.dll.a', 'spacepy.dll'],
+        'darwin':  ['libspacepy.dylib', 'libspacepy.so',
+                    'spacepy.dylib', 'spacepy.so'],
+        }.get(sys.platform, ['libspacepy.so'])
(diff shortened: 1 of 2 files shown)


spacepy/lib.py
```diff
@@ -99,7 +99,7 @@ def load_lib():
     """
     libdir = os.path.dirname(os.path.abspath(__file__))
     if sys.platform == 'win32':
-        libnames = ['spacepy.dll']
+        libnames = ['spacepy.dll', 'libspacepy.dll.a']
```
drsteve, line 102: Does list order imply priority order? And if so, is there a possibility that an old dll could be found (say, if spacepy was installed without cleanup of a previous install)?

jtniehof: They are checked in that order. It doesn't look like either setup.py or pip removes files if installing over an existing version, which doesn't make me happy.

I guess we _probably_ should put a version symbol in there to be sure it matches (although we're not doing that for anything else and I can see it maybe possibly being a problem for, say, irbem.) I don't necessarily want to jump to that.

How about I 1. do like with pycdf, and test-load each one found, so if there are two and one is (say) wrong architecture, it just won't load and move on to the next; 2. put the dll.a first, because even though _in general_ I'm not convinced that's more likely to be the newer one, _in this particular case_ it probably will be.

And then think about this issue later.

drsteve (approved):

Issue #310, opened 2020-05-11
test coverage: poppy


jtniehof:
45%. 

- Huge chunks of stuff when have_libspacepy=False, should be easy (test twice and twiddle have_livespacepy between)
- assoc_mult
- applyRefractory

More difficult:
- plot_mult, maybe integration test
- plot, same thing
- plot_two_ppro, again

PR #646, opened 2022-09-07
PoPPy bugfixes and test coverage

### LINKED ISSUE #608: poppy.assoc fails when lags is numpy array
There's some gnarly 2010-era code in the `assoc` method of `PoPPy` that doesn't work with current numpy "The truth value of an array with more than one element is ambiguous." This gets tripped if `lags` is a numpy array instead of list.

Should be fairly easy to fix but requires a test (embarrassing that it has gone this long) and is a good opportunity to fix up a small mess.

```python
import numpy
import spacepy.poppy

lags = list(range(-5, 5))
pop = spacepy.poppy.PPro([1, 2, 3], [2.5], lags=lags, winhalf=1, verbose=True)
pop.assoc()
# calculating association for series of length [3, 1] at 10 lags
```

```python
import numpy
import spacepy.poppy

lags = numpy.arange(-5, 5)
pop = spacepy.poppy.PPro([1, 2, 3], [2.5], lags=lags, winhalf=1, verbose=True)
pop.assoc()
# assoc error: attributes lags and winhalf must be populated
```

### Error message/Traceback:
No traceback because this is in an over-eager try/except, but taking that out:

```python
Input In [71], in <cell line: 1>()
----> 1 pop.assoc()

File ~/miniconda/envs/panhelio/lib/python3.9/site-packages/spacepy/poppy.py:180, in PPro.assoc(self, u, h)
    178 if u:
    179     self.lags = u
--> 180 assert self.lags
    181 if h != None:
    182     self.winhalf = h

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

```

### Version of SpacePy
Current master from github

jtniehof:
This PR makes a few improvements to PoPPy:

- The Python backend was completely broken, so added tests using the Python backend instead of C (assuming C is present), as suggested in #310. Fixed Python backend so that these tests pass. This bumps the coverage from 45% to 52%, so still much to do on #310.
- Fixes (and tests) passing lags as a numpy array (Closes #608). There are some slight behavior changes associated. One, it could be argued that the previous checks would have failed for an _empty_ window and lags. However, the error message clearly states "must be populated" not "cannot be empty". So this is an easier check and seems consistent. Second, the error message was previously returned as a string, not printed. PoPPy is fairly inconsistent on print vs. return vs. warnings vs. exceptions. Changing to print seemed to make more sense here, pending an actual overhaul of that interaction.
- A number of style cleanups, particularly comparisons to None

## PR Checklist


jtniehof:
I guess the question is whether this merits a release note. The Python backend has been broken for a LONG time so clearly nobody was using it, and it was always kind of a fallback. The lags as array thing seems minor. OTOH, if someone is for some reason stuck with an old SpacePy and wondering why this stuff is broken, the release note might help...

Note also that, because of the way numpy random seeds are structured, the bootstrap will give different results for a given seed between the Python and C backends. I think that's okay (pretty sure we don't guarantee those to be consistent between machines, for instance.)

drsteve (approved):
Noting that the PR description says "_The Python backend was completely broken_" but doesn't explicitly say that it's now fixed... however, lines 725-728 in the revised version should fix the issue. So I'm approving but would like confirmation and/or a better description in the PR for future reference.

Everything else is good and boy is it rough reading some of this very old code...

jtniehof:
> Noting that the PR description says "_The Python backend was completely broken_" but doesn't explicitly say that it's now fixed...

Updated...since tests were added and the CI passes, that seemed implicit, but PEP 20...

Issue #544, opened 2021-08-24
pycdf can't see CDF library on M1 mac


johncoxon:
My issue is that I can't import the `pycdf` submodule.

### Minimal example to reproduce issue:
I installed spacepy by cloning the repo to get around issue #530, checking out the 0.2.2 release tag, and running `python setup.py install`. Then I ran `from spacepy import pycdf` and I get an error as described below. Running `os.environ[CDF_LIB]` returns `/opt/cdf/cdf/lib` which is the location of the CDF library on my machine, as expected.

### Error message/Traceback:
```
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-1-a80d87add280> in <module>
----> 1 from spacepy import pycdf

/opt/homebrew/Caskroom/miniforge/base/lib/python3.9/site-packages/spacepy/pycdf/__init__.py in <module>
   1284 _libpath, _library = Library._find_lib()
   1285 if _library is None:
-> 1286     raise Exception(('Cannot load CDF C library; checked {0}. '
   1287                      'Try \'os.environ["CDF_LIB"] = library_directory\' '
   1288                      'before import.').format(', '.join(_libpath)))

Exception: Cannot load CDF C library; checked /opt/cdf/cdf/lib/libcdf.dylib, /opt/cdf/cdf/lib/libcdf.dylib, /opt/cdf/cdf/lib/libcdf.dylib. Try 'os.environ["CDF_LIB"] = library_directory' before import.
```

### OS, Python version, and dependency version information:
```
macOS-11.5.2-arm64-arm-64bit
sys.version_info(major=3, minor=9, micro=6, releaselevel='final', serial=0)
numpy=1.21.2
scipy=1.7.0
matplotlib=3.4.3
```

### Version of SpacePy
I cloned the repo, and checked out release 0.2.2 on the `master` branch.

jtniehof:
Thanks for the details. I see you're on an ARM Mac. Was the CDF library compiled for the same architecture as your Python? (Either ARM or Intel under Rosetta). If those don't match, it won't be able to load.

Incidentally if you're grabbing a release you don't need to clone and check out the tag; there are [downloads available for the releases](https://github.com/spacepy/spacepy/releases).

johncoxon:
[...]

PR #649, opened 2022-09-12
Update minimum numpy version on Apple Silicon

### LINKED ISSUE #638: pip builds against different numpy than installed.

jtniehof:
#639 closed #638 by specifying the minimum-possible version of numpy in the build requirements, so that it would run on that and all subsequent versions of numpy. Unfortunately the Python version is not the only thing impacting the minimum working numpy version; Apple Silicon requires newer numpy in some cases (and it's not just that one version of numpy is the minimum for Apple Silicon; it's also still a function of Python version.)

This PR updates the required numpy version. The version numbers were determined by trial-and-error (i.e. "what version of numpy can I install on this version of Python?"). Python 3.6 and earlier don't work on Apple Silicon at all, so those are left out.

To test, use:
```
pip install git+https://github.com/jtniehof/spacepy.git@build_numpy_min_m1
```

(of course also setting `SDKROOT` if on conda on Mac).

Further testing can be done by explicitly installing an older numpy after SpacePy is installed, e.g. `pip install numpy~=1.20.0`.

## PR Checklist


jtniehof:
I've tested this on Linux (conda + Python 3.9), Intel Mac (conda + Python 3.8 and macports + Python 3.9), and ARM Mac (conda + Python 3.9 and macports + Python 3.9). Should be nothing different with Windows and I hate dealing with compilers on Windows, so can be tested more fully with the release.

@spacecataz , can you check on the failures that you were seeing earlier and verify they're fixed there? Then we can merge this, close #544, and get going on 0.4.1 (with the other PRs).

spacecataz:
Tested on M1 mac, looks good/passes tests.

spacecataz (approved):

PR #652, opened 2022-09-21
Update minimum dependencies for 0.5.0

### LINKED ISSUE #26: Remove Python 2 support

### LINKED ISSUE #423: Test pycdf against CDF 2.7 in CI
#422 extends our CI to test against the earliest supported version of our dependencies, but we're still only testing against the latest CDF library. It's probably worth at least considering testing against an earlier version of the CDF library. This could be a bit of work (even building such an old library on a modern machine could be interesting).

jtniehof:
This PR takes advantage of 0.5.0 and the removal of Python 2 support to check our minimum dependency versions and, where we have convoluted workarounds, remove those workarounds and update the minimum version. CI is updated for the new minimums. This PR updates the version dependency chart to help in this. Because we don't promise to support Ubuntu 18.04LTS anymore, it was removed from the chart.

The commit messages include the rationale for each update, but they're summarized here for each dependency, and updated based on discussion here:

 * **Python 3.6**. This is the first version with fstrings and type hints, which we definitely want to be able to use. This PR does *not* remove all the Python 2.7 and early 3.x workarounds; those should be handled over time in their own PRs (probably one per module). (Closes #26).
 * **Astropy 1.0**, no change, not seeing an immediate reason. I know some things were worked around in the initial writing but there are no ugly if statements or anything; I'd be fine with updating if we, say, change some code to use a newer API.
 * **CDF 3.5**. 3.4.0 is the first version with INT8/TT2000 support, and 3.4.1 fixes a bug that had a fairly complicated workaround. Requiring 3.5 so we don't have to draw the line between 3.4 and 3.4.1, and also because 3.5.1 is the earliest easily available from NASA. Added 3.5.1 to the CI (and fixed one minor test with 3.5.1). Closes #423.
 * **dateutil 2.1**. matplotlib stopped shipping this a long time ago, so treating this like any other dependency. 2.0 was the first version with Python 3 support. It isn't available in pip, and 2.1 was a rework of the code to combine the Python 2 and 3 code, so 2.1 is worth it for easy testing. I'm not seeing any real features in later versions.
 * **h5py 2.10**. The variable-length handling (i.e. UTF-8 strings) changed in 2.10 (the docs say 2.9 in places, but they're wrong). We can go to 2.10 (it's in Ubuntu 20.04), but there were some further changes in 3.0, so there are still some workarounds...just fewer.
 * **matplotlib 3.1**. We definitely need 2.1 for things in pybats (#653). 3.1 adds the ability to change backend after import, which should simplify our testing considerably. Also every version of matplotlib is adding more useful stuff. The epoch shift is in 3.4, so can't get rid of that without losing 20.04 support.
 * **numpy 1.15.1**. Needed 1.12 for Python 3.6 support. (Python 3.7 requires 1.15). 1.15.0 has a bug that we've been avoiding; I forget what it is but it basically makes numpy uninstallable. 1.16 is in Ubuntu 20.04 so it's the newest possible minimum. (FWIW, 1.13 is in 18.04). I have not found obvious workarounds in the SpacePy code for the maintenance of compatibility with older numpy. There are some nice-to-have functions added:
   * 1.13: [isin](https://numpy.org/doc/stable/reference/generated/numpy.isin.html)
   * 1.15: [stack for masked arrays](https://numpy.org/devdocs/reference/generated/numpy.ma.stack.html#numpy.ma.stack); [gcd](https://numpy.org/devdocs/reference/generated/numpy.gcd.html#numpy.gcd) (and lcm); starting values for many aggregate functions.
 * **scipy 1.0** 0.19.0 is required for Python 3.6 support. (Python 3.7 requires 1.0.) We can go up to 1.3 (Ubuntu 20.04). 1.0 is considered a stable API so it seems a better target. The ODE solver updates in 1.0 may enabling rework of the pybats tracing using those solvers.
 * **sphinx 1.8**. Ubuntu 20.04 has 1.8. This isn't a big deal for SpacePy right now, but if I bring in some of the doc improvements from dbprocessing, there are changes in workarounds at 1.4 and 1.8. Obviously this doesn't affect users much.

## PR Checklist


jtniehof:
This is ready. I'm particularly interested in rationale for going to a later numpy or scipy version...

drsteve:
### General thoughts on minimum version
Our dependency support rule 2 is annoying here, as we say we won't drop support for older versions without cause... If we're just applying the moving wall of Ubuntu LTS support then I'm good applying that more broadly.

Given our stated dependency policy I think the list you have above is mostly good.
I don't think we have a compelling reason to bump the minimum supported version of Matplotlib past 1.5 beyond the fact that _it'd be nice_.

I'd also like to bump the versions of numpy and scipy as you mention, but again, I don't see a reason we can cite.

### Some thoughts on the numpy version...
Numpy 1.15 also allows:
- initial values to be specified for `np.ufunc.reduce`, `np.sum`, `np.prod`, `np.max` and `np.min` (I will use this)
- NaN in histograms (as long as bins are explicit), this will be great for robustness in `Spectrogram` and `seapy` (and `seapy` is on my list to improve)
- and it brings the updates you mentioned from 1.13, plus the Heaviside function (added in 1.14)

1.14 also allowed `np.loadtxt` (and others) to read files with different encodings and that's one fewer avenues for breakage in `seapy` and `ae9ap9`

And if we take 1.15 then we can specify `>1.15.1` so we don't have to say 1.15 is cool as long as it's not 1.15.0.

**None of these are _strong_ reasons to pick 1.15. But 1.15 is still earlier than the version indicated by the _second-most recent_ Ubuntu LTS. ** So if the LTS moving wall, combined with "we will want to use this", makes a strong enough reason then sure, let's bump to the 20.04 LTS version (1.16)

**Otherwise I could anticipate having some minor fixes for robustness in a PR fairly soon that would use some of these features.**

### Thoughts on scipy version
Realistically, 1.0 is fairly old and they made a lot of API changes. As they note in the 1.0 release notes, moving to 1.0 (for them) indicates stability and backwards-compatibility. I'm happy to take them at their word and call it the first "mature" release, so that might be an exception to our release dependency rule 2.

### Thoughts on Matplotlib version
Without using features or dropping workarounds I don't think we have cause to update, unless the Ubuntu LTS moving wall (+ "we will use it") is considered enough. 1.5 let us drop the colormaps, so that was good. I'd love a reason to at least move to 2.0, but our workarounds are for much later changes. So either we: 1) move to something recent and drop the workarounds; 2) stay where we are; or 3) plan updates for 0.5 that will necessitate a move. I don't like option 1.

jtniehof:
I don't want rule 2 to tie our hands--if we want to rephrase it, we should. What I want to avoid is "we should drop support just because it's old." But (and details below), "there's a new feature that I expect to use generally even if I'm not writing the code *right now*" seems like a good reason to me. That helps avoid the bind of "oh I need this thing but we don't support that version"--again, we can make the next release a .0 and update it then, but since I'm doing a big cleanup now, it seems to make sense to do some cleanup.

So your "I would use this" sounds like a reason for numpy 1.15.

The 1.0 sounds like a decent reason for scipy 1.0, and I also do think we want to look at those integrators.

For matplotlib, I looked through the release notes and had lots of "I'd probably use that." I also *suspect* (and this is projection) that if we start testing our matplotlib code more robustly, we're probably going to find API stuff that's changed and will be a pain to keep compatible with 1.5. Can we fix it later? Sure. But maybe good to move on now.

Note these updates will irritate me, because I still have production machines on 18.04, but that's my problem....

jtniehof:
Incidentally I'll update this PR once we seem to have some consensus on the versions.

drsteve:
Updated version list looks good given the pybats updates, testing simplification, etc. 

I'll pull the PR and test on some HDF5-heavy workflows just to convince myself that the datamodel updates are as solid as they look. Do we also need to touch anything in `fromCDF` or `toCDF` to make sure it's in line with more recent versions of CDF?

I think that'll cover everything I can think of here.

jtniehof:
Good question. Nothing in this affects ``toCDF``/``fromCDF``. However, #568 made the default CDF time type TT2000. That means people will get TT2000 variables regardless of the setting of the ``TT2000`` kwarg. #568 also made non-backward-compatible CDFs the default. That actually doesn't affect ``toCDF``, as it forces non-backward mode regardless of the ``backward`` kwarg.

So I think we should probably have a separate PR to clean up ``toCDF``, documenting default of TT2000 but allow ``TT2000=False`` and also either making the ``backward`` kwarg work or removing it (people can call ``pycdf.lib.set_backward`` before ``toCDF``).

That PR would be unrelated to this one.

jtniehof:
The update I just pushed was because I found another case where pycdf was doing something different for library version 2, which we don't support as of this PR. So I removed that code.

jtniehof:
#654 fixes the toCDF issues raised above, but probably shouldn't be merged until after this one.

drsteve (commented):
I'll circle back to this later today, as so far I've only looked through docs and CI (datamodel/pycdf/tests will be next). Initial comments:
- I assume that updating tests so we can drop the `xvfb-run` in the CI is slated for a later PR?
- Release notes, line 17. I know it says it just after (and elsewhere), but I'd prefer to see this say "requires Python 3.6 (or later)."

jtniehof:
>     * I assume that updating tests so we can drop the `xvfb-run` in the CI is slated for a later PR?

Yep.

>     * Release notes, line 17. I know it says it just after (and elsewhere), but I'd prefer to see this say "requires Python 3.6 (or later)."

That's indeed confusing...I'll even leave out the parens.

drsteve (approved):
Everything makes sense from here. Some good clean-up.

spacecataz (approved):

PR #653, opened 2022-09-23
pybats:Increase Pybats test coverage & quality-of-life improvements

spacecataz:
This update increases test coverage for pybats while applying several quality-of-life updates.

The biggest item here is expanded testing for pybats, including for many of the plotting features. Plotting features now test against returning the correct MPL class. Most Bats2d calculations are now tested.

Some important QoL improvements: 

- New handling of multi-species calculations with a unified function that is shared to different classes (e.g., Bats2d and VirtSats).
- New calculations, including curl that can handle multifluid output, calculation of U_perp and U_parallel, and total current calculation for RIM output.
- Big improvement to ImfFile class in terms of variable naming, reading, and writing.
- ImfFile.quicklook() is now far more flexible for creating customized solar wind plots.
- Lots of linting!

## PR Checklist

- [x] Pull request has descriptive title
- [x] Pull request gives overview of changes
- [x] New code has inline comments where necessary
- [x] Any new modules, functions or classes have docstrings consistent with SpacePy style
- [x] Added an entry to release notes if fixing a significant bug or providing a new feature
- [x] New features and bug fixes should have unit tests
- [ ] Relevant issues are linked to (e.g. `See issue #` or `Closes #`)

PR #656, opened 2022-09-30
Fix Sphinx warnings; build docs in CI

jtniehof:
This PR fixes all our current Sphinx warnings and adds checks to CI. If the docs fail to build without warnings, the CI will fail.

Most of the Sphinx warnings are fairly straightforward. Some are a matter of providing temporary objects for the autosummary to find, where we have attributes built at run time. I am NOT trying to fix anything other than the Sphinx warnings here--there are some things not in proper numpydoc format, etc.

A note on the CI: this is testing with oldest and newest set of dependencies, not a whole mess. It also is heavily copy/pasted from the unit test runs. This is probably possible to clean up, but primarily wanted to get it working first, and there are subtle differences throughout. Note that there is a new set of pip caches, because the docs install sphinx and numpydoc. These can be brought together better in the future.

One thing that may not be obvious: if Sphinx exits with an error, this is supposed to fail immediately. The GHA docs claim that, by explicitly specifying bash as the shell, pipefail is set and the job will fail if any command fails.

I chose to not try and build the docs if the unit tests fail. This is because unit test failures can indicate a failure to build (esp. irbem) and so the docs could also fail for the same reason.

EDIT: #653 is merged and pybats docs fixed, so ready to review.

## PR Checklist

### DIFF
--- a/spacepy/igrf.py
+++ b/spacepy/igrf.py
@@ -131,6 +131,12 @@ class IGRF():
 
         Dipole moments (`dict`).
     """
+    dipole = {}
+    """Characteristics of dipole (`dict`)."""
+
+    moment = {}
+    """Dipole moments (`dict`)."""
+
     def __init__(self):
         self.__status = {'coeffs': False,
                          'init': False,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,16 +172,103 @@ jobs:
         run: (! file irbem-lib-*/*.f irbem-lib-*/source/*.f | grep -q Unicode )
         working-directory: ${{ github.workspace }}/spacepy/irbempy/
 
+  docs:
+    name: Build docs
+    runs-on: ${{ matrix.os }}
+    needs: [test]
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - python-version: "3.6"
+            os: ubuntu-20.04
+            numpy-version: "~=1.15.1"
+            piplist: "python-dateutil~=2.1.0 scipy~=1.0.0 matplotlib~=3.1.0 h5py~=2.10.0 astropy>=1.0,<1.1 sphinx~=1.8.0 numpydoc"
+            cflags: "-Wno-error=format-security"
+            cdf-version: "3.5.1"
+            cdf-munged-version: "35_1"
+            dep-strategy: "oldest"
+          - python-version: "3.10"
+            os: ubuntu-20.04
+            numpy-version: ">=1.21.0"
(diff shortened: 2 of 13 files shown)


spacecataz (approved):

spacepy/igrf.py
```diff
@@ -131,6 +131,12 @@ class IGRF():
 
         Dipole moments (`dict`).
     """
+    dipole = {}
```
drsteve, line 134: If we're doing this here, would it make sense to change how the dicts are populated later? Or at the very least remove line 223 (which just makes the dict again)

drsteve (commented):
I don't like the aesthetics of the class level declarations with associated docstring (e.g, 134 & 138 in `igrf.py` and 127 & 130 in `poppy.py`). However, if it makes docs cleaner and easier then I guess I'll just suffer some discomfort. Just one short question and I'm otherwise good on this.

jtniehof:
> I don't like the aesthetics of the class level declarations with associated docstring

Without an object grab onto, the autosummary blows its top, and it doesn't instantiate classes so it won't get instance attributes. So it's unfortunately unavoidable.

Once I pull in the dbprocessing patches for Sphinx, that can go away in most cases. You probably won't like those patches though :) It won't work for the vectorized functions in pycdf though, since numpy requires the vectorization to take place after the method is bound.

spacepy/igrf.py
jtniehof: This line makes a class member. Line 223 makes an instance member (which then masks this class member). If line 223 were left out, every instance of the class would share the same `dipole` dict.

It would be possible to make the class members `None`, but making them a dict helps document that this is, well, a dict.

drsteve: Ah, yes, good catch. I still dislike having the class member floating just for the sake of docs, but point taken. And if we have to have this there I'd rather it was a dict so the type is described properly.

drsteve:
> > I don't like the aesthetics of the class level declarations with associated docstring
> 
> Without an object grab onto, the autosummary blows its top, and it doesn't instantiate classes so it won't get instance attributes. So it's unfortunately unavoidable.

It's still ugly and mildly obfuscated in purpose. But like I said, I can deal with it if it means better/easier docs.

> 
> Once I pull in the dbprocessing patches for Sphinx, that can go away in most cases. You probably won't like those patches though :) It won't work for the vectorized functions in pycdf though, since numpy requires the vectorization to take place after the method is bound.

Great, I look forward it...

drsteve (approved):

PR #661, opened 2022-10-11
Plot support for SpaceData

jtniehof:
This PR provides methods to plot data in SpaceData assuming they are ISTP-compliant. There are also several helper methods to support this which I intend to continue to flesh out in the future.

This is not going to make a perfect plot of everything yet (for instance, the EPI-Lo data in the example, which I will check on; likely a simpleSpectrogram issue). And we can always add thousands of knobs to tweak. I'd like to try and focus this PR on the core functionality and then we can open up enhancements to build on it.

## PR Checklist

### DIFF
--- a/Doc/source/release_notes.rst
+++ b/Doc/source/release_notes.rst
@@ -42,6 +42,11 @@ New features
 now support reading from gzipped input files; filenames ending with
 ``.gz`` are assumed to be gzipped.
 
+`.datamodel` now has explicit support for operations using
+ISTP-compliant metadata; see `~.datamodel.dmarray` and
+`~.datamodel.SpaceData` for details. In particular note the plotting
+support with `~.datamodel.ISTPContainer.plot`.
+
 :mod:`.pycdf` :meth:`~.pycdf.Library.set_backward` now returns the prior
 state of backward compatibility mode.
 
--- a/Doc/source/release_notes.rst
+++ b/Doc/source/release_notes.rst
@@ -47,6 +47,9 @@ ISTP-compliant metadata; see `~.datamodel.dmarray` and
 `~.datamodel.SpaceData` for details. In particular note the plotting
 support with `~.datamodel.ISTPContainer.plot`.
 
+`~.plot.spectrogram.simpleSpectrogram()` supports plotting zeros as
+valid data.
+
 :mod:`.pycdf` :meth:`~.pycdf.Library.set_backward` now returns the prior
 state of backward compatibility mode.
 
--- a/Doc/source/release_notes.rst
+++ b/Doc/source/release_notes.rst
@@ -47,8 +47,8 @@ ISTP-compliant metadata; see `~.datamodel.dmarray` and
 `~.datamodel.SpaceData` for details. In particular note the plotting
 support with `~.datamodel.ISTPContainer.plot`.
 
-`~.plot.spectrogram.simpleSpectrogram()` supports plotting zeros as
-valid data.
+`~.plot.spectrogram.simpleSpectrogram()` supports treating zeros as
+valid data on log plots.
 
 :mod:`.pycdf` :meth:`~.pycdf.Library.set_backward` now returns the prior
 state of backward compatibility mode.
--- a/spacepy/datamodel.py
+++ b/spacepy/datamodel.py
@@ -227,8 +227,8 @@ class MetaMixin(object):
 class ISTPArray:
     """Mixin class for array using ISTP metadata.
 
-    Additional methods for array types, such as `dmarray`, assuming
-    that the attributes of the array use the
+    Array types like `dmarray` provide all these methods if attributes
+    of the array use the
     `ISTP metadata standard <https://spdf.gsfc.nasa.gov/sp_use_of_cdf.html>`_
 
     Note that some operations that may seem to relate to an array (e.g.
@@ -301,9 +301,9 @@ class ISTPArray:
 class ISTPContainer(collections.abc.Mapping):
     """Mixin class for containers using ISTP metadata.
 
-    Additional methods for container types, such as `SpaceData`, assuming
-    that the attributes of the container and the arrays it contains use
-    the `ISTP metadata standard <https://spdf.gsfc.nasa.gov/sp_use_of_cdf.html>`_
+    Container types like `SpaceData` provide all these methods if attributes
+    of the container and the arrays it contains use the
+    `ISTP metadata standard <https://spdf.gsfc.nasa.gov/sp_use_of_cdf.html>`_
 
     .. autosummary::
         ~ISTPContainer.lineplot
@@ -419,6 +419,26 @@ class ISTPContainer(collections.abc.Mapping):
         -------
         fig : `matplotlib.figure.Figure`
             The figure on which the variables were plotted
+
+        See Also
+        --------
+        lineplot : to line plot a single variable
+        spectrogram : to make a spectrogram of a single variable
+
+        Examples
+        --------
+
+        >>> import spacepy.pycdf
+        # https://rbsp-ect.newmexicoconsortium.org/data_pub/rbspa/ECT/level2/
+        >>> with spacepy.pycdf.CDF('rbspa_ect-elec-L2_20140115_v2.1.0.cdf') as f:
+        ...     data = f.copy()
+        >>> fig = data.plot(['FESA', 'Position'])
+        >>> fig.show  # if needed
+        # https://spp-isois.sr.unh.edu/data_public/ISOIS/level2/
+        >>> with spacepy.pycdf.CDF('psp_isois_l2-summary_20201130_v13.cdf') as f:
+        ...     data = f.copy()
+        >>> fig = data.plot(['A_H_Rate_TS', 'H_CountRate_ChanP_SP'])
+        >>> fig.show  # if needed
         """
         if fig is None:
             import matplotlib.pyplot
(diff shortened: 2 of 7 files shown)


jtniehof:
I've pushed some simpleSpectrogram tests and fixes here, but at this point I'm not very confident in simpleSpectrogram. I'm going to do some more work on it in this PR and will mark draft until that's done.

Doc/source/release_notes.rst
```diff
@@ -42,6 +42,14 @@ New features
 now support reading from gzipped input files; filenames ending with
 ``.gz`` are assumed to be gzipped.
 
+`.datamodel` now has explicit support for operations using
+ISTP-compliant metadata; see `~.datamodel.dmarray` and
+`~.datamodel.SpaceData` for details. In particular note the plotting
+support with `~.datamodel.ISTPContainer.plot`.
+
+`~.plot.spectrogram.simpleSpectrogram()` supports plotting zeros as
+valid data.
```
drsteve, line 51: When log-scaling. If log-scaling is off zeros are always valid...

spacepy/datamodel.py
```diff
@@ -223,10 +224,373 @@ def meta(self):
...
+        See Also
+        --------
+        .pycdf.istp.nanfill : an in-place variant
+
+        Notes
+        -----
+        .. versionadded:: 0.5.0
+
+        Comparisons with ``FILLVAL`` are done using `~numpy.isclose` and
+        so may NaN values thare are near, but not identical, to fill.
```
drsteve, line 269: I'm having trouble parsing this... is there a typo confusing me?

drsteve (commented):
On my first pass, including playing with functionality, this looks good. I'm going to have to take another look at the code for the mixin just because there's a lot of thinking required.

Doc/source/release_notes.rst
jtniehof: Right.

spacepy/datamodel.py
jtniehof: It's using NaN as a verb...changed to "may replace values, hopefully that clarifies?

jtniehof:
The mixin approach does, unfortunately, obscure things a bit. The goal is to eventually have CDF and Var inherit from the relevant mixins and get the same functionality. One thing I'd be interested in feedback on is if it makes sense to link the mixin classes from the main class docstring (as is done right now) or just document all those methods directly in the main class--it's just a difference in how the autosummary is set up, it doesn't require putting the methods in the main class.

The other thing is that they're sort of abstract base classes, in that they're not intended to be instantiated directly (and won't work as such) but the Python ABC setup is oriented around requiring the subclass to implement various methods--sort of the opposite of this situation. That's why the weird ``attrs: collections.abc.Mapping``...it's our first case of using the Python 3.6 type specifiers, and it's a way to hint at the existence of that (Python) attribute. Mostly it makes pylint happy (and so can go away). I can't find back the StackOverflow question, but basically the options for doing this are the type specifier or making a property that's also decorated as an abstract method. That seems a lot of machinery for syntactic sugar.

drsteve (changes requested):
In addition to my very minor inline review comments I have some things I wanted to note, but they probably don't need to be in this PR. So I'm good once the inline comments are addressed.

I tried to be a bad user, by applying this to non-ISTP datasets and the error messages aren't fantastic. I'm happy for that to get put on the list for subsequent updates but wanted to note it here.

I also tried to be a good user and grabbed an Arase data file, which turned out poorly and has a very cryptic error message...
```
In [27]: with spacepy.pycdf.CDF('/home/smorley/tmp/erg_lepi_l2_omniflux_20170907_v03_00.cdf') as fh:
    ...:     foo = fh.copy()
    ...:

In [28]: foo.plot(['FPDO'])
Input In [28], in <cell line: 1>()
----> 1 foo.plot(['FPDO'])

File ~/miniconda3/envs/default310/lib/python3.10/site-packages/spacepy/datamodel.py:486, in ISTPContainer.plot(self, vnames, fig)
    484             ax.legend(h, l, loc='best')
    485     else:
--> 486         self.spectrogram(k, target=ax)
    487 return fig

File ~/miniconda3/envs/default310/lib/python3.10/site-packages/spacepy/datamodel.py:519, in ISTPContainer.spectrogram(self, vname, target)
    517 fig, ax = spacepy.plot.utils.set_target(target)
    518 x = self[v.attrs['DEPEND_0']]
--> 519 data = v.replace_invalid()
    520 x = self[v.attrs['DEPEND_0']]
    521 y = self[v.attrs['DEPEND_1']]

File ~/miniconda3/envs/default310/lib/python3.10/site-packages/spacepy/datamodel.py:274, in ISTPArray.replace_invalid(self)
    272 idx = numpy.zeros_like(data, dtype=bool)
    273 if self.attrs.get('FILLVAL') is not None:
--> 274     idx |= numpy.isclose(data, self.attrs['FILLVAL'])
    275 if self.attrs.get('VALIDMIN') is not None:
    276     idx |= data < self.attrs['VALIDMIN']

File <__array_function__ internals>:180, in isclose(*args, **kwargs)

File ~/miniconda3/envs/default310/lib/python3.10/site-packages/numpy/core/numeric.py:2355, in isclose(a, b, rtol, atol, equal_nan)
   2347 # Make sure y is an inexact type to avoid bad behavior on abs(MIN_INT).
   2348 # This will cause casting of x later. Also, make sure to allow subclasses
   2349 # (e.g., for numpy.ma).
   (...)
   2352 #       timedelta works if `atol` is an integer or also a timedelta.
   2353 #       Although, the default tolerances are unlikely to be useful
   2354 if y.dtype.kind != "m":
-> 2355     dt = multiarray.result_type(y, 1.)
   2356     y = asanyarray(y, dtype=dt)
   2358 xfin = isfinite(x)

File <__array_function__ internals>:180, in result_type(*args, **kwargs)

TypeError: The DType <class 'numpy._FloatAbstractDType'> could not be promoted by <class 'numpy.dtype[str_]'>. This means that no common DType exists for the given inputs. For example they cannot be stored in a single array unless the dtype is `object`. The full list of DTypes is: (<class 'numpy.dtype[str_]'>, <class 'numpy._FloatAbstractDType'>)
```
I believe the issue here is that the `FILLVAL`, `VALIDMIN` and `VALIDMAX` are stored as strings in the Arase data files. While that's not strictly ISTP compliant, I suspect we should be handling it. I'd also support having this in a subsequent update, but wanted to get it on the list.

spacepy/datamodel.py
```diff
@@ -223,10 +224,373 @@ def meta(self):
...
+            return True
+        # Reasonable dividing line is probably 4 stacked line plots
+        return self.shape[-1] < 5
+
+
+class ISTPContainer(collections.abc.Mapping):
+    """Mixin class for containers using ISTP metadata.
+
+    Container types like `SpaceData` provide all these methods if attributes
+    of the container and the arrays it contains use the
```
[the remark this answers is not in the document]
jtniehof: Docstring updated to indicate it's always there, just won't _work_ without ISTP metadata. I have thought about if we want to have some means for the metadata to indicate it's ISTP, and/or run an ISTP check and enable methods accordingly, but it also makes some sense for things to stagger along where they can if it's only sort-of-ISTP.

spacepy/datamodel.py
```diff
@@ -223,10 +224,373 @@ def meta(self):
...
+        -----
+        .. versionadded:: 0.5.0
+
+        Examples
+        --------
+
+        >>> import spacepy.pycdf
+        # https://rbsp-ect.newmexicoconsortium.org/data_pub/rbspa/ECT/level2/
+        >>> with spacepy.pycdf.CDF('rbspa_ect-elec-L2_20140115_v2.1.0.cdf') as f:
+        ...     data = f.copy()
```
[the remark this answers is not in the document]
jtniehof: Changed the first example.

jtniehof:
> In addition to my very minor inline review comments I have some things I wanted to note, but they probably don't need to be in this PR. So I'm good once the inline comments are addressed.

Fixing those required a new commit, and at that point I suspected the CI might not work because of the issue fixed in #666...so there was a big rebase, but only the final commit is actually new.

> I tried to be a bad user, by applying this to non-ISTP datasets and the error messages aren't fantastic.

I'm torn on this, because it does get a bit into trying to read the tealeaves and/or a hideous amount of error checking. If it's ISTP-compliant, it should work. If not ISTP-compliant, results are undefined. That might be a more helpful approach if the ISTP-checking bits of pycdf worked on a SpaceData (which is definitely a thing I'd like for the future). And I do think there are some cases of "almost compliant" where we can probably figure things out--another thing for the future is do more work through helper methods rather than directly manipulating the attributes, which gives scope for smarter.

The short answer is I'm afraid we might have to figure on a case-by-case basis where it makes sense to try and have a good error message, where it makes sense to try and plunge ahead, and where it makes sense to halt and catch fire. If we try to write preemptive error messages for all cases, we'd be here forever.

> I believe the issue here is that the `FILLVAL`, `VALIDMIN` and `VALIDMAX` are stored as strings in the Arase data files. While that's not strictly ISTP compliant, I suspect we should be handling it.

That's pretty seriously noncompliant ("must match the data type of the variable") but we're living in a world where those files exist.

I'd be a fan of working down current major data sets and trying them one at a time, opening issues as necessary.

drsteve (approved):

drsteve:
I think we have a growing laundry list of things to augment this, but they don't need to stand in the way of immediate progress. I'll merge now.

Issue #679, opened 2023-02-26
Installing spacepy errors


dstansby:
I've just tried installing `spacepy` in a fresh Python 3.11 conda environment, and it fails. See below for the output. It's trying to install `spacepy` from a source distribution, and for some reason as part of the build process is trying to compile numpy, which is where the error occurs.

### Minimal example to reproduce issue:

```
python -m pip install spacepy
```

### Error message/Traceback:

```
% python -m pip install spacepy
Collecting spacepy
  Using cached spacepy-0.4.1.zip (18.2 MB)
  Installing build dependencies ... error
  error: subprocess-exited-with-error
  
  × pip subprocess to install build dependencies did not run successfully.
  │ exit code: 1
  ╰─> [1312 lines of output]
      Ignoring numpy: markers 'python_version == "3.5"' don't match your environment
      Ignoring numpy: markers 'python_version == "3.6"' don't match your environment
      Ignoring numpy: markers 'python_version == "3.7" and (platform_system != "Darwin" or platform_machine != "arm64")' don't match your environment
      Ignoring numpy: markers 'python_version == "3.7" and platform_system == "Darwin" and platform_machine == "arm64"' don't match your environment
      Ignoring numpy: markers 'python_version == "3.8" and (platform_system != "Darwin" or platform_machine != "arm64")' don't match your environment
      Ignoring numpy: markers 'python_version == "3.8" and platform_system == "Darwin" and platform_machine == "arm64"' don't match your environment
      Ignoring numpy: markers 'python_version == "3.9" and (platform_system != "Darwin" or platform_machine != "arm64")' don't match your environment
      Ignoring numpy: markers 'python_version == "3.9" and platform_system == "Darwin" and platform_machine == "arm64"' don't match your environment
      Collecting setuptools
        Using cached setuptools-67.4.0-py3-none-any.whl (1.1 MB)
      Collecting wheel
        Using cached wheel-0.38.4-py3-none-any.whl (36 kB)
      Collecting numpy~=1.21.0
        Using cached numpy-1.21.1.zip (10.3 MB)
        Installing build dependencies: started
        Installing build dependencies: finished with status 'done'
        Getting requirements to build wheel: started
        Getting requirements to build wheel: finished with status 'done'
        Preparing metadata (pyproject.toml): started
        Preparing metadata (pyproject.toml): finished with status 'done'
      Building wheels for collected packages: numpy
        Building wheel for numpy (pyproject.toml): started
[...]

PR #680, opened 2023-02-26
Use oldest-supported-numpy in build reqs

dstansby:
Fixes https://github.com/spacepy/spacepy/issues/679. As described at https://github.com/scipy/oldest-supported-numpy, this should avoid the numpy versions having to be continually updated in the requirements, and will keep them up to date.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,11 +1,2 @@
 [build-system]
-requires = ["setuptools", "wheel",
-    "numpy ~= 1.12.0 ; python_version == '3.6'",
-    "numpy ~= 1.15.0 ; python_version == '3.7' and (platform_system != 'Darwin' or platform_machine != 'arm64')",
-    "numpy ~= 1.20.0 ; python_version == '3.7' and platform_system == 'Darwin' and platform_machine == 'arm64'",
-    "numpy ~= 1.17.0 ; python_version == '3.8' and (platform_system != 'Darwin' or platform_machine != 'arm64')",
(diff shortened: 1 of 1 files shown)


jtniehof:
I looked into oldest-supported-numpy for #639 and #649 since, indeed, doing this by hand is not much fun. Unfortunately I seem to have lost most of my notes on why it didn't work out. I remember part of the issue is that it's set to require the oldest numpy _with a binary wheel_ and that broke things on some older numpy, including those that had binaries in conda. It also doesn't distinguish between Apple Silicon and other ARM so it broke Raspberry Pi support.

It doesn't seem like there's any way to avoid having to disable build isolation in some circumstances, it's just a question of which circumstances.

jtniehof:
I had a shower moment (actually cooking breakfast, but still) and realized that we can probably use no-build-isolation while making the binary wheels (where full control is the most important) and then use oldest-supported-numpy for the rest. I'm not sure of the set of people who won't be using binary wheels (once I have the Mac binaries) but also get thrown by no-build-isolation, but that might be the right solution. I will have to double-check just how bad the situation is on Pi.

dstansby:
Going to close this since it doesn't seem like it's going to be merged.

jtniehof:
I was actually just looking back at this in the context of some challenges in #723 (and was leaving it open until I was done with that work). For now there are enough moving pieces that it's still working out best to do the versioning by hand, but I'm going to keep oldest-supported-numpy in my back pocket--e.g. it might appear as a backstop for unreleased Python versions.

This discussion's been useful; thanks.

Issue #669, opened 2022-11-08
Typo in datamodel string typing


drsteve:
`atsype` should be `astype`
https://github.com/spacepy/spacepy/blob/0ec652f963e21001905740cda4ffae6871ad4672/spacepy/datamodel.py#L1580

PR #681, opened 2023-02-27
Port installer to setuptools (abandon distutils)

jtniehof:
This PR reworks the installer into a modern setuptools framework rather than the existing distutils-based setup.py. Some functionality in the distutils namespace is still used; this is provided by setuptools (pypa/setuptools#2806). This also includes, of course, dropping the deprecated numpy distutils. Some other results/implications:

- The required versions of pip, setuptools, and build are now explicitly documented (and testsed)
- Support for Intel and PG compilers is dropped...gfortran only
- A lot of the installer docs are updated since it's all pip, all the way
- Built docs are no longer included in the source distribution (rst source is); it just seems a rare use case and this lets me simplify setup.py
- The installer no longer clears out old files if installing on top of old spacepy...again, this simplifies setup.py, and pip should(?) do the uninstall first instead of mixing stuff together.

Closes #688. Also fixes the current CI failure, which is due to numpy 1.18 not building under the latest cython.

I still have a personal punchlist for installer tasks, but this is getting the foundation laid for hopefully smaller task-focused updates later.

## PR Checklist

- [X] Pull request has descriptive title
- [X] Pull request gives overview of changes
- [X] New code has inline comments where necessary
- [X] (N/A) Any new modules, functions or classes have docstrings consistent with SpacePy style
- [X] Added an entry to release notes if fixing a significant bug or providing a new feature
- [X] (N/A) New features and bug fixes should have unit tests
- [X] Relevant issues are linked to (e.g. `See issue #` or `Closes #`)

PR #684, opened 2023-04-04
Datamodel ISTP updates; toDataFrame

jtniehof:
This PR contains updates to the ISTP handling in datamodel, basically followup from #661.

- adds support for parsing units strings
- provides output to pandas DataFrame and input from
- provides output to Astropy Quantity and input from

"Input from" is fairly sparse because both objects are far less rich than ISTP metadata. There may be some way to attach more metadata to be able to round-trip ISTP, but that's a tale for another day.

I also found that the test for #669 broke on newer h5py, so that's been updated. Not sure why that isn't showing up in CI?

## PR Checklist


jtniehof:
There are two classes of problem with the CI, both pertaining to the "oldest supported" tests.

The first is that pandas requires dateutil 2.5 and right now we only require 2.1. I hate bumping our requirements just to make *another* dependency happy, and one that we don't actually use but only have for compatibility shims. But 2.5 came out in 2016, so in practice I don't think this is a big issue and isn't worth wasting time on more convolute workarounds.

The second is that on 3.10, the earliest version of Pandas that works (1.0) has a build-requires hard-coded to a version of numpy that doesn't work on 3.10 (1.17.3). We could disable pip build isolation (and maybe dependency handling) in the CI. That is actually somewhat tempting. Or for the 3.10 "earliest" tests, we could roll forward to the oldest version of pandas that actually installs without parlor tricks; older versions of pandas are still being tested with older versions of Python anyhow.

jtniehof:
Current CI failures came after rebasing to incorporate the merge of #681 to main; probably has to do with either newer versions of stuff appearing since the last CI run or me messing up the conflicts in the rebase (both this and #681 started pulling in cython explicitly and there's probably issues with version of cython required for numpy vs. for pandas).

spacecataz (approved):
Everything looks sharp.

PR #685, opened 2023-04-11
Check for mingw32 libs before adding to path

jtniehof:
Our Windows binary wheels ship the mingw32 runtime so that people don't need it installed to run. This involves adding our shipped runtime to the DLL path. Unfortunately this addition was hard-coded, and the runtime _isn't_ there if you build from source on Windows. The result is an error of the form `FileNotFoundError: [WinError 2] The system cannot find the file specified: 'C:\\Users\\user\\python\\spacepy\\spacepy\\mingw'`.

This PR makes the DLL path change contingent on the directory actually existing.

I need to muck about a bit to make sure I reproduce the problem and also that this fix works, so draft for now.

## PR Checklist

### DIFF
--- a/spacepy/__init__.py
+++ b/spacepy/__init__.py
@@ -82,16 +82,13 @@ __all__ = ["seapy", "toolbox", "poppy", "coordinates", "time", "omni",
            "irbempy", "empiricals", "radbelt", "data_assimilation", "pycdf",
            "datamanager", "datamodel", "ae9ap9"]
 
-# on windows, make sure the mingw runtime libs are findable
-if sys.platform == 'win32':
-    minglibs = os.path.join(os.path.dirname(__file__), 'mingw')
-    if 'PATH' in os.environ:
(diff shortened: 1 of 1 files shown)


jtniehof:
Hmm, I can't reproduce this. If I have mingw32 installed, this works, since we actually do copy the libraries on an install from source as well. If I don't have mingw32 installed, the installer blows up with "Can't locate runtime libraries."

About the only way I can get this to happen is to `import spacepy` from within the _source_ tree. At that point it's not going to work anyhow, so it's probably not worth doing anything about.

I don't know. Maybe it's okay to merge this anyhow, it cleans things up a little bit and _maybe_ doesn't hurt? But in most cases where this situation happens, things are so far down the road of badness that HACF seems appropriate.

spacecataz:
This is worth keeping for the code cleanup, as discussed offline.

spacecataz (approved):

PR #708, opened 2023-08-14
Update dictree to allow returning not printing the result

jtniehof:
From discussion in #704, this updates dictree to allow returning (intead of printing) the result. The default remains printing. We could possibly change that (via a deprecation cycle) if we glue on appropriate ipython pretty-print methods. Then ipython would Just Do The Right Thing.

Also extends dictree test coverage and does some refactoring.

## PR Checklist

### DIFF
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -600,7 +600,8 @@ def human_sort( l ):
     return l
 
 
-def dictree(in_dict, verbose=False, spaces=None, levels=True, attrs=False, **kwargs):
+def dictree(in_dict, verbose=False, spaces=None, levels=True, attrs=False,
+            print_out=True, **kwargs):
     """
     pretty print a dictionary tree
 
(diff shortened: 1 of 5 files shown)


spacepy/toolbox/__init__.py
```diff
@@ -600,22 +600,28 @@ def human_sort( l ):
...
+            print_out=True, **kwargs):
     """
     pretty print a dictionary tree
 
     Parameters
     ----------
     in_dict : dict
         a complex dictionary (with substructures)
-    verbose : boolean (optional)
+    verbose : bool, default True
```
drsteve, line 612: verbosity default is False

drsteve (changes requested):
Apart from the docstring note I have no comments on the commits.
However, I will note that I usually call dictree as the `tree` method on `SpaceData` objects. Setting `print_out` to False doesn't work when calling the object method because it's just returning None every time. We should probably update that so it returns the return value from the `dictree` call?

spacepy/toolbox/__init__.py
jtniehof: Well that's embarrassing. I'm guessing a copy/paste failure. Fixed in rebase.

drsteve (approved):

jtniehof:
> Setting `print_out` to False doesn't work when calling the object method because it's just returning None every time. We should probably update that so it returns the return value from the `dictree` call?

Good catch, fixed in new commit (at tip of rebased branch). I also did a quick grep to see if there were any other calls to dictree in the spacepy code; didn't find any.

While I was at it, I grabbed the updated parameters description from `dictree` and put it on `tree`. Do we want to mention that the update also affects `tree` in the release notes? It seems excessive...

drsteve:
Agreed. Seems excessive. I'm going to go ahead and merge.

PR #730, opened 2024-02-01
Update test_toolbox to fix failing CI: import inf from numpy not scipy

drsteve:
It appears that inf was removed from the scipy base namespace at some point, so this is now failing to import. Numpy still provides this and the objects are the same in older versions of both libraries.

CI is currently failing because of this, so the change (which is backward compatible) should enable the testing to run on all supported versions.

- [N/A] New code has inline comments where necessary
- [N/A] Added an entry to release notes if fixing a significant bug or providing a new feature
- [N/A] New features and bug fixes should have unit tests

### DIFF
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -2353,7 +2353,7 @@ def intsolve(func, value, start=None, stop=None, maxit=1000):
     **Note:** Assumes func is everywhere positive, otherwise solution may
            be multi-valued.
     """
-    from scipy import inf
+    from numpy import inf
     from scipy.integrate import quad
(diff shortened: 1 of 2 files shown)


drsteve:
Also impacts `intsolve` and `dist_to_list` in toolbox.

jtniehof (approved):
Those `from scipy import inf` date back to 2010 and 2011. I can't find when it went into numpy but it was in by 1.13 (the changelog notes differences in handling of comparisons with inf) and we're requiring 1.15, so that confirms what the tests seem to indicate.

I did a quick check and didn't find any other places, so this seems complete.

PR #733, opened 2024-03-04
Mac/Linux binary wheels; include CDF in wheels; other installer updates

### LINKED ISSUE #122: Improve NASA CDF installation experience; make pycdf.download_library useful
The NASA CDF library is required for many (most?) users of SpacePy, but it's never been the easiest to install. It's also one of the FAQs for SpacePy.

I'd really like to investigate some ways of making it easier to install CDF with SpacePy.
For example:
- on install try to download and build CDF
   - easy on posix systems; harder on Windows (but doable, esp. in conda)
   - note that the Windows installer from NASA requires admin, so we can't rely on that
- build CDF and include it with the releases so that there's a functional CDF provided
   - this has the down-side of not having up-to-date CDF in case of leap-second updates, etc.
- `spacepy.pycdf.download_library` exists, but isn't usable without *already having CDF installed*
   - `spacepy.pycdf` won't import without CDF installed, so download_library isn't usable unless CDF is already present
   - `download_library` has very limited platform support anyway
- if `download_library`  can be made useful across platforms, then it can also serve as an update function to help the user stay up-to-date

jtniehof:
This PR includes several installer updates, including the support and instructions for Mac and Linux wheels and CDF library in all wheels (closes #122). Documentation has been updated to note that the binary wheels are preferred. The commit messages are the best detailed description of the changes.

There are also several little structural updates based on changes to the Python packaging ecosystem, such as different instructions for test pypi and using metadata 2.1 in the wheels.

The end state of this PR has been uploaded to [test PyPI](https://test.pypi.org/project/spacepy/0.5.0b1/). [Docs are here](https://github.com/spacepy/spacepy/files/14516183/spacepy-0.5.0b1-doc.zip).

To install this version:
```sh
pip install -i https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple/ --only-binary spacepy "spacepy>0.0d0"
```

The version number is weird but that's how to coerce installing the latest development version without installing dev versions of the dependencies.

To run the tests, _in the tests directory_ run:
```sh
python test_all.py
```

The astropy tests will fail if astropy is not installed; that's expected.

All the wheels have been tested in clean conda envs (i.e. the first thing done in the env was `pip install spacepy`, then the unit tests run). On Windows the whole OS was rolled back to before the build before making the clean conda envs.

## PR Checklist


drsteve (commented):
Looks fine to me from a read-through, but I'd be happier if someone else (paging @spacecataz) could give it a second set of eyes. The changes do look pretty clear.

The one question I have is that it looks like the Mac binary wheels are built with CDF now? But I don't see that reflected in `install_mac.rst`.

jtniehof:
> The one question I have is that it looks like the Mac binary wheels are built with CDF now? But I don't see that reflected in `install_mac.rst`.

Correct, I'll update all the install docs all at once when the Linux stuff is done, since there will be a lot of changes pushing people towards prebuilt binaries first. You're right that's inconsistent between Mac and Windows here.

Also, once I have the Linux wheels working, I'll do a full build on testpypi so we can really beat on them. We may find some functional issues in this code at that point, but it seemed worth splitting up the reviews a bit so it's less of an all-at-once issue.

jtniehof:
Rather than wait for this and then do another PR, I updated this with all the installer updates. This should basically be it. The README is out of date; I always forget to check that. So I can update that here or in the final release PR; whichever. Suggestions for how much detail to go into are welcome, e.g. CDF library given that it's now in the binaries.

drsteve (approved):
So I'll approve, but I have one question/suggestion/comment.
The install docs, etc. have been updated to say CDF is packaged with the binary wheels. It's probably useful to also say that if the user needs to update CDF without updating SpacePy, they can just install CDF as usual and we should autodetect that version (assuming proper install). Or at least something that says we can still autodetect non-bundled versions, because I can foresee getting questions about it next time CDF update. (Or whatever text is more useful and accurate that that, but you get my point)

If that seems valuable, I can approve and merge once it's in. Otherwise, I can merge anyway.

The test PyPI binary linux wheels work in the random, non-comprehensive environments I checked.

jtniehof:
> It's probably useful to also say that if the user needs to update CDF without updating SpacePy, they can just install CDF as usual and we should autodetect that version.

Good catch. I had the full rigmarole in the CDF docs, and a bit in the Windows docs, but not elsewhere. Just pushed three commits:

- Update the INSTALL file to basically say just use pip; this hadn't been updated in years.
- Updated the `__license__` to note the GCC runtime on Mac and Linux
- Updated the dependencies, install_mac, and install_linux to note you can BYOCDF

The other two I basically caught while I was noodling around on this....

Doc/source/install_linux.rst
```diff
@@ -64,8 +64,9 @@ To get the dependencies for building documentation::
 
 CDF
 ===
-If building SpacePy from source, you will need to install the NASA CDF
-library. This is not required if installing from a binary wheel.
+The NASA CDF library is included if installing from a binary wheel. If
+building SpacePy from source or wish to use a different version of the
```
drsteve, line 68: It looks like you either thought words and didn't type them, or rewrote something and lost words and a comma. Can we change to: "if building SpacePy from source, or if you wish to use a different version of the CDF library," (see also the equivalent spot in `install_mac.rst`

jtniehof: Thanks, I just added a subject and pushed.

drsteve (approved):

Issue #313, opened 2020-05-11
test coverage: toolbox


jtniehof:
Down to 70%, including the integration tests. Lots of little exceptional cases to check, bigger chunks:

- quaternionRotateVector is untested
- update isn't tested and should probably go in the integration test; it calls a lot of the other stuff that isn't tested.
- assemble with sortKey specified

jtniehof:
Self-assigning for update because I've been messing with it; the rest is up for grabs.

drsteve:
I should write the quaternion rotation tests...

jtniehof:
quaternion rotations moved to coordinates in #369, where I also added the rotation tests. @drsteve , feel free to self-assign here; if you think there should be more rotation tests, that would now be in #302.

jtniehof:
- line 42: remove the dance around this import, just do the import
- tCommon valueerrors (may be hard to hit)
- loadpickle: we can't test the out of memory, but I think we only hit "if gzip" when gzip is true
- assemble: test with sortkey, remove superfluous else
- consider breaking out update and testing subpieces
- indsFromXrange
- boothisto: write a test of the plotting, look at test_plot_utils for the setup, this test will be in a new class

PR #751, opened 2024-06-04
Improve Toolbox Test Coverage

jamesw732:
This pull request slightly improves the test coverage of `spacepy.toolbox`. Removed some unreachable or redundant lines, and added a test for `assemble` with a `sortkey`. See issue #313.
## PR Checklist

### DIFF
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -386,8 +386,6 @@ def loadpickle(fln):
                 return pickle.load(fh, encoding='latin1')
         except pickle.UnpicklingError: #maybe it's a gzip?
             gzip = True
-        else:
-            gzip = False
     if gzip:
         try:
             import zlib
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -520,9 +520,6 @@ def assemble(fln_pattern, outfln, sortkey='ticks', verbose=True):
             ax = np.where(dim==TAIcount)[0]
             if len(ax) == 1: # then match with length of TAI
                 dcomb[key] = dcomb[key][idx] # resort
-    else:
-        # do nothing
-        pass
 
     if verbose: print('\n writing: ', outfln)
     savepickle(outfln, dcomb)
--- a/spacepy/toolbox/__init__.py
+++ b/spacepy/toolbox/__init__.py
@@ -378,27 +378,27 @@ def loadpickle(fln):
     >>> d = loadpickle('test.pbin')
     """
     if not os.path.exists(fln) and os.path.exists(fln + '.gz'):
-        gzip = True
+        # It's a gzip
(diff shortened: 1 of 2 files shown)


spacepy/toolbox/__init__.py
```diff
@@ -392,8 +386,6 @@ def loadpickle(fln):
                 return pickle.load(fh, encoding='latin1')
         except pickle.UnpicklingError: #maybe it's a gzip?
             gzip = True
-        else:
-            gzip = False
     if gzip:
```
jtniehof, line 389: I think the idea was we don't even need this "if"...gzip is guaranteed to be true if we reach this point.

- the .gz exists and fln doesn't, we reach this point and gzip is True
- the fln exists OR the fln doesn't: we try to load the pickle. If it succeeds, return out. If it fails, gzip is set True. (Or some other exception is raised, e.g. file not existing....)

So we should just rip out the `if gzip` and dedent that block.

tests/test_toolbox.py
```diff
@@ -100,6 +100,18 @@ def test_assemble(self):
...
         self.assertEqual(expected, result)
 
+    def test_assemble_sorted(self):
+        tb.savepickle(os.path.join(self.tempdir, 'test_pickle_1.pkl'), self.D1)
+        tb.savepickle(os.path.join(self.tempdir, 'test_pickle_2.pkl'), self.D2)
+        tb.savepickle(os.path.join(self.tempdir, 'test_pickle_3.pkl'), self.D3)
+        # Manually sort expected by 'names'
+        expected = {'names': sorted(self.all['names']),
+                    'TAI': [val for _, val in sorted(zip(self.all['names'], self.all['TAI']))]}
+        result = tb.assemble(os.path.join(self.tempdir, 'test_pickle_[1-3].pkl'), os.path.join(self.tempdir, 'test_all.pkl'), sortkey="names", verbose=False)
```
jtniehof, line 110: This line is 157 characters long and should be split. We're not firm about 80 on SpacePy anymore (need to write an explicit style) but over 100 is too much.

jtniehof (changes requested):
Two small changes and I think we're good.

spacepy/toolbox/__init__.py
jamesw732: Ah, I see. I think we can actually cut out the gzip logic completely - we don't use the `gzip` boolean outside of line 389, and as you explained there's no way for `gzip` to be false at that point (or ever). The occurrences of `gzip = True` are pretty illustrative of the intention of the code, I'll put some comments in so that's not lost.

jtniehof (approved):
Looks good. One note: for things in the checklist that aren't applicable, we normally mark them off and put (N/A) so it's clear the person writing the PR specifically thought about it:

- [X] (N/A) Change the oil

PR #753, opened 2024-06-12
Increased coverage on test_toolbox and integration_toolbox

jongeorge1999:
Wrote tests for get_url() to increase the testing coverage on test_toolbox and integration_toolbox

### DIFF
--- a/tests/integration_toolbox.py
+++ b/tests/integration_toolbox.py
@@ -6,6 +6,7 @@ Copyright 2019 SpacePy contributors
 """
 
 import http.server
+import http.client
 requesthandlerclass = http.server.SimpleHTTPRequestHandler
 import os
 import shutil
@@ -13,6 +14,7 @@ import socket
 import tempfile
 import threading
 import unittest
+from urllib.error import HTTPError
 
 import spacepy_testing
 import spacepy.toolbox
@@ -156,6 +158,46 @@ class WebGettingIntegration(unittest.TestCase):
             conn.close()
         self.assertEqual(b'This is a test\n', data)
 
+    def testGetUrl404(self):
+        """Call get_url for a file that doesn't exist"""
+        with self.assertRaises(HTTPError):
+            data = spacepy.toolbox.get_url(
+                'http://localhost:{}/doesnotexist.txt'.format(self.port))
+    
+    def testGetUrlNoConnNoKeepAlive(self):
+        """Call get_url without specifying connection or keepalive"""
+        ctype = http.client.HTTPConnection
+        connection = ctype("localhost:{}".format(self.port))
+        with self.assertRaises(ValueError):
+            spacepy.toolbox.get_url('http://localhost:{}/'.format(self.port), conn=connection, keepalive=False)
+    
+    def testGetUrlReadableWritable(self):
+        """Call get_url to trigger readable and writable"""
+        ctype = http.client.HTTPConnection
+        connection = ctype("localhost:{}".format(self.port))
+        connection.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM) 
+        _, connection = spacepy.toolbox.get_url('http://localhost:{}/'.format(self.port), conn=connection, keepalive=True)
+        self.assertEqual(connection.sock, None)
+    
+    def testGetUrlValueErrorOn404WithKeepAlive(self):
+        """Call get_url with keepalive and a bad link"""
+        ctype = http.client.HTTPConnection
+        connection = ctype("localhost:{}".format(self.port))
+        with self.assertRaises(RuntimeError):
+            spacepy.toolbox.get_url('http://localhost:{}/doesnotexist.txt'.format(self.port), keepalive=True)
+    
+    def testGetUrlCacheNoOutfileNoKeepAlive(self):
+        """Call get_url with Cache, no outfile, and no keepalive"""
+        ctype = http.client.HTTPConnection
+        connection = ctype("localhost:{}".format(self.port))
+        with self.assertRaises(ValueError):
+            spacepy.toolbox.get_url('http://localhost:{}/'.format(self.port), cached=True, outfile=None, keepalive=False)
+    
+    def testGetUrlNoKeepaliveBadUrl(self):
+        ctype = http.client.HTTPConnection
+        connection = ctype("localhost:{}".format(self.port))
+        with self.assertRaises(RuntimeError):
+            spacepy.toolbox.get_url('http://localhost:{}/doesnotexist.txt'.format(self.port),  keepalive=False)
 
 if __name__ == "__main__":
     unittest.main()
--- a/tests/test_toolbox.py
+++ b/tests/test_toolbox.py
@@ -8,6 +8,7 @@ Copyright 2010-2014 Los Alamos National Security, LLC.
 """
 
 import builtins
+import http.client
 import time, datetime
 import glob, os, sys
 import io


tests/integration_toolbox.py
```diff
@@ -156,6 +158,46 @@ def testGetUrlToFileCachedKeepaliveChanged(self):
...
+    def testGetUrl404(self):
+        """Call get_url for a file that doesn't exist"""
+        with self.assertRaises(HTTPError):
+            data = spacepy.toolbox.get_url(
+                'http://localhost:{}/doesnotexist.txt'.format(self.port))
+    
+    def testGetUrlNoConnNoKeepAlive(self):
+        """Call get_url without specifying connection or keepalive"""
+        ctype = http.client.HTTPConnection
+        connection = ctype("localhost:{}".format(self.port))
```
jtniehof, line 170: There's no point in the assignment since it's not used again (there were some cases where we were doing this with a different assignment for Python 2 and 3 where the class name changed, but not relevant here)--so this can just be one line, in all the cases.

The connection should also be closed at the end, which basically means putting everything from the point that the connection is made to the end within a try/finally block. True for all these tests.

tests/integration_toolbox.py
```diff
@@ -156,6 +158,46 @@ def testGetUrlToFileCachedKeepaliveChanged(self):
...
         self.assertEqual(b'This is a test\n', data)
 
+    def testGetUrl404(self):
+        """Call get_url for a file that doesn't exist"""
+        with self.assertRaises(HTTPError):
+            data = spacepy.toolbox.get_url(
+                'http://localhost:{}/doesnotexist.txt'.format(self.port))
+    
+    def testGetUrlNoConnNoKeepAlive(self):
+        """Call get_url without specifying connection or keepalive"""
```
jtniehof, line 168: This says without specifying a connection, but it specifies a connection.

tests/integration_toolbox.py
```diff
@@ -156,6 +158,46 @@ def testGetUrlToFileCachedKeepaliveChanged(self):
...
+        ctype = http.client.HTTPConnection
+        connection = ctype("localhost:{}".format(self.port))
+        with self.assertRaises(RuntimeError):
+            spacepy.toolbox.get_url('http://localhost:{}/doesnotexist.txt'.format(self.port), keepalive=True)
+    
+    def testGetUrlCacheNoOutfileNoKeepAlive(self):
+        """Call get_url with Cache, no outfile, and no keepalive"""
+        ctype = http.client.HTTPConnection
+        connection = ctype("localhost:{}".format(self.port))
+        with self.assertRaises(ValueError):
```
jtniehof, line 193: The ValueErrors at least should actually check the message in the exception, see e.g. testSparsePrevInsert in test_pycdf. I think most of those RuntimeErrors are also checkable.

tests/integration_toolbox.py
```diff
@@ -156,6 +158,46 @@ def testGetUrlToFileCachedKeepaliveChanged(self):
...
+        connection = ctype("localhost:{}".format(self.port))
+        connection.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM) 
+        _, connection = spacepy.toolbox.get_url('http://localhost:{}/'.format(self.port), conn=connection, keepalive=True)
+        self.assertEqual(connection.sock, None)
+    
+    def testGetUrlValueErrorOn404WithKeepAlive(self):
+        """Call get_url with keepalive and a bad link"""
+        ctype = http.client.HTTPConnection
+        connection = ctype("localhost:{}".format(self.port))
+        with self.assertRaises(RuntimeError):
```
jtniehof, line 186: Are we actually getting a RuntimeError in the case where a connection is specified (here) but an HTTPError where it isn't (testGetURL404), or is this a case where we're getting lucky because HTTPError is a subclass of RuntimeError? We should capture the most specific exception that we're expecting, and test the message (as mentioned elsewhere).

jtniehof (changes requested):
In addition to the specifics, please go through the checklist. If there's something that's not applicable, check it off and mark N/A:
 - [X] (N/A) useless thing

This shows you've actually evaluated the item in the checklist. For example, #313 is clearly related and should be mentioned in the description (but without the magic closing words, as it won't close that).

tests/integration_toolbox.py
jongeorge1999: This test was duplicating the functionality of another test, so it was removed

tests/integration_toolbox.py
jongeorge1999: I made all of the new test cases check the message of the exceptions

tests/integration_toolbox.py
jongeorge1999: This was accidental, and it is now fixed, and the docstring is updated

tests/integration_toolbox.py
jongeorge1999: I'm not completely sure what you meant by this, so,I got rid of the connection variable and called the ctype() in the get_url function arguments

jtniehof: The connection assignment is useful (particularly since you have to close it later). The ctype one isn't...just instantiate http.client.HTTPConnection directly.

tests/integration_toolbox.py
```diff
@@ -155,7 +157,45 @@ def testGetUrlToFileCachedKeepaliveChanged(self):
         finally:
             conn.close()
         self.assertEqual(b'This is a test\n', data)
-
+    
+    def testGetUrlReadableWritable(self):
+        """Call get_url to trigger readable and writable"""
+        try:
+            connection = http.client.HTTPConnection("localhost:{}".format(self.port))
+            connection.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
jtniehof, line 165: What's the reason for explicitly assigning the socket and checking that it's None after? I'm confused by this pattern--HTTPConnection should automatically do all the setup that's necessary.

tests/test_toolbox.py
```diff
@@ -8,6 +8,7 @@
 """
 
 import builtins
+import http.client
```
jtniehof, line 11: This import isn't used

jtniehof (changes requested):

tests/integration_toolbox.py
jongeorge1999: This is because I wanted to hit the branch where it sets the socket to None and then returns, so I was just testing that the value of socket is correct from that branch.

tests/test_toolbox.py
jongeorge1999: This is fixed for the next commit

tests/integration_toolbox.py
jtniehof: Hmm, reading that now. What needs to happen is to have an established connection here, and stale data (otherwise that isn't really hit). So instead of assigning to the socket, calling `connection.request('GET', 'localhost:{}/')` before `get_url` should do the trick--since it's starting the connection but not actually reading the data off it.

jongeorge1999: This is causing an error from socketserver.py: `ConnectionAbortedError: [WinError 10053] An established connection was aborted by the software in your host machine` I am going to do a bit of investigation with this one

jtniehof: If it winds up being a rabbit hole, we can leave that branch untested. Although it's probably worth doing a test where we just call get_url back-to-back twice so the connection is reused (it just won't do the clearing old data off the socket).

jongeorge1999: Didn't seem to work, the weird part is that it doesn't seem like a socket is ever assigned, printing out the socket, after creating the connection, and both get_url() calls result in "None" all three times.

jtniehof:
Brings us up to 72% between unit and integration test scripts. I would like to test the user-agent thing but it looks like more trouble than it's really worth.

jtniehof (approved):
Looks good!

PR #755, opened 2024-06-12
Explicit checks for type comparisons to CDF_TIME_TT2000 in istp

### LINKED ISSUE #752: pycdf.istp.FileChecks.all fails : TypeError: '<' not supported between instances of 'float' and 'datetime.datetime'
Hello,

I have an issue when I run istp check for the CDF file solo_L1_rpw-tds-surv-stat-cdag_20230402_V01.cdf (available in https://rpw.lesia.obspm.fr/roc/data/pub/shared/outgoing/solo_L1_rpw-tds-surv-stat-cdag_20230402_V01.cdf).

I'm not sure to know where the problem comes from. 
At a first glance, the CDF file looks well formed, but maybe not.

Thanks for help.

```
cdf = 'solo_L1_rpw-tds-surv-stat-cdag_20230402_V01.cdf'
with pycdf.CDF(str(rpw_cdf)) as cdf:
    result = pycdf.istp.FileChecks.all(cdf)
```

### Error message/Traceback:
```
  File "/obs/xbonnin/software/DataPool/SOLO/RPW/CDF/Tools/check_rpw_cdf.py", line 1054, in main
    result = pycdf.istp.FileChecks.all(cdf)
  File "/home/xbonnin/.virtualenvs/mypy/lib/python3.8/site-packages/spacepy/pycdf/istp.py", line 731, in all
    for e in VariableChecks.all(f[v], catch=catch)))
  File "/home/xbonnin/.virtualenvs/mypy/lib/python3.8/site-packages/spacepy/pycdf/istp.py", line 129, in all
    errors.extend(f(v))
  File "/home/xbonnin/.virtualenvs/mypy/lib/python3.8/site-packages/spacepy/pycdf/istp.py", line 587, in validscale
    return cls._validhelper(v, False)
  File "/home/xbonnin/.virtualenvs/mypy/lib/python3.8/site-packages/spacepy/pycdf/istp.py", line 505, in _validhelper
    if numpy.any((minval > attrval)) or numpy.any((maxval < attrval)):
TypeError: '<' not supported between instances of 'float' and 'datetime.datetime'
```

Run on Debian 12, Python 3.11, numpy 1.24.4

### Version of SpacePy
0.6

jongeorge1999:
Made an explicit check, and an assertion for the validchecks on istp. Check that the system produces an error when a non-timetype is compared against a timetype.

Closes #752 

- [x] (N/A) Any new modules, functions or classes have docstrings consistent with SpacePy style
- [x] (N/A)Added an entry to release notes if fixing a significant bug or providing a new feature

### DIFF
--- a/spacepy/pycdf/istp.py
+++ b/spacepy/pycdf/istp.py
@@ -492,6 +492,15 @@ class VariableChecks(object):
                     continue
                 if firstdim: #Add pseudo-record dim
                     attrval = numpy.reshape(attrval, (1, -1))
+            if (v.attrs.type(which) in spacepy.pycdf.lib.timetypes) != (v.type() in spacepy.pycdf.lib.timetypes):
+                errs.append(
+                    '{} type {} not comparable to variable type {}.'.format(
+                        which,
+                        spacepy.pycdf.lib.cdftypenames[v.attrs.type(which)],
+                        spacepy.pycdf.lib.cdftypenames[v.type()]
+                    )
+                )
+                continue
             # min, max, variable data all same dtype
             if not numpy.can_cast(numpy.asanyarray(attrval),
                                   numpy.asanyarray(minval).dtype):
--- a/spacepy/pycdf/istp.py
+++ b/spacepy/pycdf/istp.py
@@ -469,8 +469,8 @@ class VariableChecks(object):
         for which in (whichmin, whichmax):
             if not which in v.attrs:
                 continue
-            atype = v.attrs.type(which) # attribute type
-            vtype = v.type() # variable type
+            atype = v.attrs.type(which)
+            vtype = v.type()
             if atype != vtype:
                 errs.append(
                     '{} type {} does not match variable type {}.'.format(
--- a/spacepy/pycdf/istp.py
+++ b/spacepy/pycdf/istp.py
@@ -469,12 +469,14 @@ class VariableChecks(object):
         for which in (whichmin, whichmax):
             if not which in v.attrs:
                 continue
-            if v.attrs.type(which) != v.type():
+            atype = v.attrs.type(which) # attribute type
+            vtype = v.type() # variable type
+            if atype != vtype:
                 errs.append(
                     '{} type {} does not match variable type {}.'.format(
                         which,
-                        spacepy.pycdf.lib.cdftypenames[v.attrs.type(which)],
-                        spacepy.pycdf.lib.cdftypenames[v.type()]))
+                        spacepy.pycdf.lib.cdftypenames[atype],
+                        spacepy.pycdf.lib.cdftypenames[vtype]))
             attrval = v.attrs[which]
             multidim = bool(numpy.shape(attrval)) #multi-dimensional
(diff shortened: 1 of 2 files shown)


spacepy/pycdf/istp.py
```diff
@@ -492,6 +492,15 @@ def _validhelper(cls, v, rng=True):
                     continue
                 if firstdim: #Add pseudo-record dim
                     attrval = numpy.reshape(attrval, (1, -1))
+            if (v.attrs.type(which) in spacepy.pycdf.lib.timetypes) != (v.type() in spacepy.pycdf.lib.timetypes):
```
jtniehof, line 495: Can we just add this as an "or" to the below if statement and call it good, since the error message is the same and all?

tests/test_pycdf_istp.py
```diff
@@ -523,6 +524,14 @@ def testValidScale(self):
              'SCALEMIN type CDF_INT2 does not match variable type CDF_BYTE.'
              ],
              errs)
+        tt2000.attrs['SCALEMIN'] = 200.42e45 #tt2000 testing
```
jtniehof, line 527: probably better to use tt2000.attrs.new() so you can specify the type explicitly instead of having to use a very very large number.

tests/test_pycdf_istp.py
```diff
@@ -523,6 +524,14 @@ def testValidScale(self):
...
+        errs = spacepy.pycdf.istp.VariableChecks.validscale(tt2000)
+        errs.sort()
+        self.assertEqual(
+            ['SCALEMIN type CDF_DOUBLE does not match variable type CDF_TIME_TT2000.',
+             'SCALEMIN type CDF_DOUBLE not comparable to variable type CDF_TIME_TT2000.',
```
jtniehof, line 532: Aha, this is weird! I hadn't realized that the code would actually register a validation failure at this point, but keep going...since the exception then hid that. Nice, this is definitely the behavior we want.

spacepy/pycdf/istp.py
jtniehof, line 495: We're also using v.attrs.type(which) and v.type() a lot, starting back on line 472. Should we just assign to a variable at the point where we decide to execute the loop? (i.e. first line after the `continue`). Call 'em attrtype and vtype or something?

spacepy/pycdf/istp.py
```diff
@@ -469,12 +469,14 @@ def _validhelper(cls, v, rng=True):
         for which in (whichmin, whichmax):
             if not which in v.attrs:
                 continue
-            if v.attrs.type(which) != v.type():
+            atype = v.attrs.type(which) # attribute type
+            vtype = v.type() # variable type
```
jtniehof, line 473: The comments are probably superfluous (and per PEP 8, same-line comments should have two spaces before the #). At least I'm hoping those names are clear enough that the comments are superfluous....

jtniehof (approved):
Alright, let's stick a fork in this one now :)

Issue #411, opened 2020-08-21
Soft dependency chart does not render well


balarsen:
Going way back to e1932247eddac1683f76f02da11e09d4a1bfe4f4 the soft dependency chart does not render well. 

Maybe it needs lines, or less text or a taller format or...

![Annotation 2020-08-21 144831](https://user-images.githubusercontent.com/508397/90933813-6faaa580-e3bd-11ea-8ed9-36678daeae76.png)

It is like 3 full pages tall and can't tell the columns near the bottom, 

related to #8 and #334

jtniehof:
This looks like a Sphinx style change, all tables no longer have lines in 3.4.0. 1.6.7 does have lines. We'll have to go find that bit of style and port it forward.

drsteve:
Likely related to:
https://review.opendev.org/c/openstack/openstackdocstheme/+/661344

jtniehof:
There's also some weirdness where I'm not seeing any space between a parameter's name and its type; we should look into it while fixing the theme.

jtniehof:
Oh boy this looks like it's gotten worse. There's so much white space in there that I'm starting to think this presentation doesn't make much sense anymore anyhow.

I'm still back and forth on just treating scipy and matplotlib like numpy, i.e. expect to have them. But people have complained about scipy being a "heavy" dependency...

PR #759, opened 2024-06-24
Convert sphinx documentation to full autodocs

jtniehof:
This PR reworks the documentation to rely entirely on autosummary rather than having to separately list all child documentation. That also means a lot more documentation is building than before, so there are a lot of fixes for newly-revealed misformatting.

This does start a split in our documentation between API documentation and user documentation. Most of our .rsts were just the auto directives. pycdf.rst had a lot of other content, so that content now stands alone separate from the API documentation. pybats had very long docstrings including some headings that are not numpydoc compliant. Those were pulled out into a new .rst file. This is just the start of the process--plenty of other modules (e.g. time) have a lot in the docstring that can be pulled into separate rst files.

There's a lot more than needs to happen to really have the documentation up to snuff, this is purely the *structural* change. We will also need to e.g. look at changes to templates. So for example we are not addressing #411 yet.

## PR Checklist


jongeorge1999:
In the dbprocessing, it seems like they do it the same way, where there is rst files linked to each page of the html. Is the intention here to only list out new modules in the top level __init__.py, and it will go through all modules/classes/functions and documents them in a directory tree?

jtniehof:
> In the dbprocessing, it seems like they do it the same way, where there is rst files linked to each page of the html.

Nope, [this](https://github.com/spacepy/dbprocessing/blob/main/docs/developer/api.rst) is the only rst file for all the [API documentation](https://spacepy.github.io/dbprocessing/developer/api.html). In the case of dbp we're putting the toctree entry (sort of the "controlling reference" which forces the documentation to actually be built) in the rst, and the `__init__.py` just has another set of references to all the submodules for convenience. (I cannot recall why I did the toctree in the rst and not in the docstring).

Of the list of modules in the [module reference](https://spacepy.github.io/#spacepy-module-reference) section of [index.rst](https://github.com/spacepy/spacepy/blob/main/Doc/source/index.rst), pycdf is the only one (plus a few notes on irbempy) where the .rst isn't simply a bunch of stuff that the autosummary will automatically make. So set this section up just like in dbprocessing: make it an autosummary table that specifies the toctree, and leave the `__init__.py` as the set of references without the toctree entry. We'll "rescue" the pycdf documentation later--once we have this set up, the API documentation and "howto" documentation will [become distinct](https://diataxis.fr/).

jongeorge1999:
Everything is working as it should it seems like(except pycdf which is a special case), from here there is legacy autosummary code in the code classes that I should probably go through and remove, it doesn't harm anything, but it really has no reason to be there.

jtniehof:
@spacecataz , even though we aren't ready for review yet, you might want to take a look at the pybats changes to start getting your head around the idea and see if the split is credible.

jongeorge1999:
> @spacecataz , even though we aren't ready for review yet, you might want to take a look at the pybats changes to start getting your head around the idea and see if the split is credible.

As a followup to this: currently the skeleton of how the documentation is going to be abstracted is there, and as it stands now, I am working on moving some of the chattier docstrings over to the user documentation section, while leaving the portions necessary to using spacepy as a programming tool in the API sections.

Doc/Makefile
```diff
@@ -49,7 +49,7 @@ PDFs := $(foreach dir, $(IMAGEDIRS), $(patsubst %.svg,%.pdf,$(wildcard $(SOURCED
 images: $(PDFs)
 
 clean:
-	-rm -rf $(BUILDDIR)/* source/autosummary/*.rst source/autosummary/autosummary/*.rst $(PDFs)
+	-rm -rf $(BUILDDIR)/* source/autosummary build $(PDFs)
```
jtniehof, line 52: This is removing the entire build directory (and assuming its name). I'm okay with removing the entire autosummary directory instead of individual .rst, but let's not remove ``build`` (which may or may not be the build directory).

Doc/source/coordinates.rst
```diff
@@ -1,32 +1,99 @@
+#############################################################################################
+spacepy.coordinates - Implementation of Coords class functions for coordinate transformations
```
jtniehof, line 2: Let's not include the "spacepy" in these titles, since they will be going into indices and tables of contents and such.

Doc/source/coordinates.rst
```diff
@@ -1,32 +1,99 @@
...
 
-##############################################
-coordinates - module for coordinate transforms
-##############################################
 
-.. automodule:: spacepy.coordinates
+See also the `full API documentation <spacepy.coordinates>`.
 
-.. currentmodule:: spacepy.coordinates
+NOTES
```
jtniehof, line 12: No need for all caps, and I don't think "notes" is a very informative heading. It can probably just be removed.

Doc/source/coordinates.rst
```diff
@@ -1,32 +1,99 @@
...
+    ``ECI2000``, ``ECIMOD``, ``ECITOD``, and ``CDMAG``. With the exception
+    of ``ECIMOD``, these can be used with the existing IRBEM backend, and
+    will be converted to their closest equivalents.
+
+    .. versionchanged:: 0.4.0
+
+    The default backend for coordinate transformations was changed from IRBEM
+    to the CTrans-based SpacePy backend.
+
+differences between representations
```
jtniehof, line 60: Leading cap here

spacepy/coordinates.py
```diff
@@ -2,97 +2,12 @@
...
-imported as ``spc``, to set the SpacePy backend as the default and set
-``itol`` to 5 seconds:
-
-    >>> spc.DEFAULTS.set_values(use_irbem=False, itol=5)
-
 Authors: Steven Morley and Josef Koller
 Institution: Los ALamos National Laboratory
 Contact: smorley@lanl.gov
 
+For additional documentation :doc:`../coordinates`
```
jtniehof, line 9: Does this not work as ``:doc:`coordinates`?`` If not, probably should be `/coordinates` to be explicitly relative to the root. Or maybe there should be an explicit reference reference anchor added there and this can be `:ref:` If this is rendering with the title then it's probably okay to stay as `:doc:` with the absolute ref.

Doc/source/coordinates.rst
```diff
@@ -1,32 +1,99 @@
...
+and AstroPy. However, for use cases where the required precision is of order
+1 percent the output can be considered equivalent.
+
+Setting options for coordinate transformation
+---------------------------------------------
+The backend for coordinate transformations can be provided at
+instantiation of a :class:`Coords` object using a keyword
+argument. However, for convenience and flexibility the options can be
+set at the module level. Configurable options include the backend used
+(:mod:`~spacepy.irbempy` or SpacePy's :mod:`~spacepy.ctrans`) and the
```
jtniehof, line 86: A lot of these probably do not need the `:class:` etc.; we have it set to auto-search. Similarly you may be able to get away with just `irbempy`. Do the minimal markup that works, here and elsewhere, so it's a bit more readable if looking at source.

Doc/source/datamodel.rst
```diff
@@ -1,45 +1,169 @@
...
+    :local:
+
+See also the `full API documentation <spacepy.datamodel>`.
+
+Documentation
+=============
+
+This contains the following classes:
+ * :py:class:`dmarray` - numpy arrays that support .attrs for information about the data
+ * :py:class:`SpaceData` - base class that extends dict, to be extended by others
```
jtniehof, line 20: We almost certainly don't need the py domain specified here.

Doc/source/pybats.rst
```diff
@@ -1,7 +1,347 @@
...
+For more information on the SWMF, please visit the
+`Center for Space Environment Modeling <http://csem.engin.umich.edu>`_.
+
+See also the `full API documentation <spacepy.pybats>`.
+
+PyBats Module
+^^^^^^^^^^^^^
+
+pybats
+------
```
jtniehof, line 20: Follow the [Sphinx conventions](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#rst-sections) for header ordering. No need to rework all the existing stuff in this order, but where we're adding new headers, try to do that.

Each of these submodule-specific sections should link to that submodule, and the submodule should contain a `:ref:` link back to these sections.

Doc/source/time.rst
```diff
@@ -1,27 +1,154 @@
...
-#########################################################################
+.. contents::
+   :depth: 2
+   :local:
 
-.. automodule:: spacepy.time
+See also the `full API documentation <spacepy.time>`.
 
-.. rubric:: Classes
+Notes
```
jtniehof, line 14: Again don't really need a header here.

spacepy/ctrans/__init__.py
```diff
@@ -136,7 +107,7 @@ class Ellipsoid(dm.SpaceData):
     """Ellipsoid definition class for geodetic coordinates
 
     Other Parameters
-    ----------------
+    ================
```
jtniehof, line 110: Can we revert all these without it being a complete disaster? [These should all be -,](https://numpydoc.readthedocs.io/en/latest/format.html) and we don't need to change everything existing, but we should change in the wrong direction.z

spacepy/lib.py
```diff
@@ -91,8 +91,10 @@ def load_lib():
     Normally this will be in the directory where spacepy is installed,
     under libspacepy.
 
-    @return: the open library
-    @rtype: ctypes.CDLL or ctypes.WinDLL
+    Returns
+    =======
+    library : ctypes.DLL or ctypes.winDLL
+        the open library
```
jtniehof, line 97: Does this render properly (linking to the types) with the "or"? At any rate this isn't true anymore, it's all ctypes.CDLL (note not DLL).

spacepy/lib.py
```diff
@@ -119,11 +121,15 @@ def load_lib():
...
-    @param lib: library where functions specified in L{call_dict} live.
-    @type lib: ctypes.WinDLL or ctypes.CDLL
+    Parameters
+    ==========
+    call_dict : dict
+        call dictionary. Keyed by function name;
+        values are [return type, argtype0, argtype 1...]
+    
+    lib : ctypes.WinDLL or ctypes.CDLL
+        library where functions specified in L{call_dict} live.
```
jtniehof, line 131: Just CDLL. Also Sphinx doesn't support argument linking (epydoc's one advantage) so this should just be:
```
``call_dict``
```
...we just render parameter references as code

spacepy/poppy.py
```diff
@@ -420,7 +396,7 @@ def plot(self, figsize=None, dpi=80, asympt=True, show=True, norm=True,
                 plt.ylabel(
                     'Normalized Association Number n(u, h={0}) / n({1}, h={0})'.format(
                     self.winhalf,
-                    r'$\mathrm{u\rightarrow\infty}$'
+                    r'$\\mathrm{u\rightarrow\infty}$'
```
jtniehof, line 399: This is a raw string so the backslash doesn't need to be escaped.

spacepy/pybats/__init__.py
```diff
@@ -169,20 +70,16 @@ def parse_filename_time(filename):
 
     Parameters
     ==========
-    filename : string
+    filename - string
```
jtniehof, line 73: Needs to be a colon for numpydoc.

spacepy/pybats/__init__.py
```diff
@@ -169,20 +70,16 @@ def parse_filename_time(filename):
...
     Returns
     =======
-    i_iter : integer, list of integers, or None
+    i_iter - integer, list of integers, or None
         The iteration at which the file was written, if found.
-    runtime : float, list of floats, or None
+    runtime - float, list of floats, or None
         The run time (in seconds) at which the file was written, if found.
-    time : datetime.datetime, list of datetimes, or None
+    time - datetime.datetime, list of datetimes, or None
```
jtniehof, line 82: Again, :

spacepy/pybats/__init__.py
```diff
@@ -323,19 +220,19 @@ def add_planet(ax, rad=1.0, ang=0.0, add_night=True, zorder=1000,
...
        Set the rotation of the day-night terminator from the y-axis, in degrees
        Defaults to zero (terminator is aligned with Y-axis.)
-    add_night : boolean
+    add_night - boolean
        Add night hemisphere.  Defaults to **True**
-    zorder : int
+    zorder - int
        Set the matplotlib zorder of the patch to set how other plot
        elements order with the inner boundary patch. Defaults to 1000,
        nightside patch is given zorder of *zorder+5*.
```
jtniehof, line 238: I'm not going to point all these out, please go through the diffs and find them. Then rebase into the correct spot so that they never happened in history. Or you can rebase --interactive and label the commits where these were introduced as "edit". When the rebase reaches that commit, edit this out, do `commit --amend`, and then continue the rebase.

spacepy/pybats/__init__.py
```diff
@@ -422,25 +319,48 @@ def _read_idl_ascii(pbdat, header='units', start_loc=0, keep_case=True):
...
+        3D magnetosphere files, this preserves the 3D block structure,
+        as required for the BATSRUS interpolator in the `Kamodo
+        Heliophysics model readers package
+        <https://github.com/nasa/kamodo>`_. Before 0.5.0, binary
+        unstructured data were sorted in an attempt to put nearby
+        positions close to each other in the data arrays. This sorting
+        was nondeterministic and has been removed; see
+        :meth:`~spacepy.pybats.bats.Bats2d.extract` and
+        :class:`~spacepy.pybats.qotree.QTree` for processing
+        adjacent cells. (ASCII data were never sorted.)
```
jtniehof, line 363: Why was this moved here? It's specific to IdlFile

spacepy/pybats/bats.py
```diff
@@ -308,7 +285,7 @@ class Extraction(PbData):
     :func:`~spacepy.pybats.bats.Bats2d.extract`)
 
     Parameters
-    ----------
+    ==========
```
jtniehof, line 288: Not going to point out the rest of these, either.

spacepy/pybats/rim.py
```diff
@@ -101,7 +84,7 @@ def tex_label(varname):
         'sigmah':r'$\sigma_{Hall}$',
         'sigmap':r'$\sigma_{Peder}$',
         'jr':r'$J_{radial}$',
-        'mA/m^2':r'$\mu A/m^{2}$',
+        'mA/m^2':r'$\\mu A/m^{2}$',
```
jtniehof, line 87: Raw string, shouldn't need double backslash.

spacepy/pybats/rim.py
```diff
@@ -246,16 +229,20 @@ def calc_j(self):
         Calculate total horizontal current as related values.  Each will be
         stored into *self* using the typical key-value approach.  Calculations
         are done for both the northern and southern hemisphere with the
-        appropriate prefixes ('n_' and 's_') applied to each key.
+        appropriate prefixes ('n' and 's') applied to each key.
```
jtniehof, line 232: Unless you've checked in the code and there's really no underscore, this is probably wrong. If it's trying to render as italic make it code:
```
``n_s``
```

spacepy/pybats/rim.py
```diff
@@ -289,14 +276,14 @@ def calc_I(self):
         Iup:   The total upward current over one hemisphere.
         Idown: The total downward current over one hemisphere.
 
-        Values are stored in the object with the prefix 'n_' or 's_'
+        Values are stored in the object with the prefix 'n' or 's'
```
jtniehof, line 279: Same

spacepy/pybats/rim.py
```diff
@@ -289,14 +276,14 @@ def calc_I(self):
...
         to indicate the hemisphere and may be accessed via self['n_I'], etc.
 
         Values are stored in units of mega Amps.
 
         Parameters
         ==========
-
+        N/A
```
jtniehof, line 286: Just leave the section out if it's not used.

spacepy/pybats/rim.py
```diff
@@ -387,7 +374,7 @@ def add_cont(self, var, target=None, n=50, maxz=False, lines=False,
             that is strictly positive and "Seismic" for diverging data.
             Alternatively, legacy Ridley Ionosphere Model color maps can be
             loaded using "l_wr" (white red) or "l_bwr" (blue-white-red),
-            where the "l_" prefix indicates legacy and not Matplotlib color maps.
+            where the "l" prefix indicates legacy and not Matplotlib color maps.
```
jtniehof, line 377: Again, the underscore probably matters. Please check through for this pattern so I don't have to highlight them all.

spacepy/pycdf/__init__.py
```diff
@@ -103,93 +105,19 @@ class Library(object):
     Calling the C library directly requires knowledge of
     `ctypes`.
 
-    Instantiating this object loads the C library, see :doc:`/pycdf` docs
+    Instantiating this object loads the C library, see :doc:`/autosummary/spacepy.pycdf` docs
```
jtniehof, line 108: This should either refer to the module or to the new rst. I think probably the new rst. Never directly refer to the autosummary docs, always to objects in them.

spacepy/pycdf/__init__.py
```diff
@@ -5919,7 +5699,7 @@ def type(self, name, new_type=None):
         @return: CDF variable type, see `pycdf.const`
         @rtype: int
         @note: If changing types, old and new must be equivalent, see CDF
-               User's Guide section 2.5.5 pg. 57
+        User's Guide section 2.5.5 pg. 57
```
jtniehof, line 5702: This should probably be either de-epydoced or just left alone.

jtniehof (commented):
Plenty of things to go back and check. I'm not going to go through the history and look for rebasing opportunities right now, because there's a lot of stuff that needs to be undone. It probably is going to make most sense to use the "edit" command of git rebase to take care of all those first, then we can look at more cleanup possibilities later.

spacepy/coordinates.py
jongeorge1999: Yeah, I tried `:ref:` at first and it didn't seem to work, `:doc:` worked and renders with the title correctly

jtniehof: ref doesn't work unless you [give it a reference label](https://www.sphinx-doc.org/en/master/usage/referencing.html#role-ref).

Doc/source/coordinates.rst
jongeorge1999: Oddly enough, either way I do it does not result in a hyperlink being generated.

spacepy/pybats/rim.py
jongeorge1999: The problem with these was that sphinx does not like things ending in an underscore. I tried a few ways to leave the underscore there, but sphinx always gave back warnings. I figured the 'n' is still technically a prefix, which is why my fix was pretty trivial.

Doc/source/pybats.rst
jongeorge1999: Yeah, when I was going through originally I was looking at the sphinx conventions, it didn't seem to work the way it should, this weirdness with the ^^^^^ and the ------- was due to sphinx not rendering the sections correctly any other way.

Doc/source/time.rst
jongeorge1999: I did this just for the table of contents, I just called it Notes because I didn't know what else to title it.

spacepy/ctrans/__init__.py
jongeorge1999: This one may be a little harder, I think the easiest and cleanest way to do it would be to change it back in a commit, I have a commit for all these changes anyways, so I can just lump it into there. Hopefully because I'm reverting it back to what it was before it was changed, that the diff will recognize that, and it will remove it from the diff completely.

spacepy/lib.py
jongeorge1999: I don't actually see a documentation section for lib, maybe its a keyword that sphinx isn't picking up on? I checked 0.7.0 docs and 0.6 docs and neither have documentation for lib

spacepy/pycdf/__init__.py
jongeorge1999: the new rst bring you to the additional user documentation, I am trying to get to the API documentation which is autogenerated, I think going to the module would be the best best.

Doc/source/pybats.rst
jtniehof: It's going to depend a bit on what other .rst files are doing. Stick to the sphinx ordering even if the ToC looks nested wrong) and we'll double back later to make them consistent across all the docs.

Doc/source/time.rst
jtniehof: The paragraph right under the ToC doesn't need an entry in the ToC. It's okay.

spacepy/ctrans/__init__.py
jtniehof: The github diff will look clean, but the history is still hard to crawl through. I think you're doing the rebasing at this point so I think we're good to go.

spacepy/lib.py
jtniehof: It's not in our list of modules in index.rst, but it is linked if you click on the main top-level spacepy link in the list of modules. I think I prefer it that way, honestly.

spacepy/pybats/rim.py
jtniehof: Yeah, but that doesn't tell the user what the prefix _really is_. Wrapping in double backticks should do the trick--I tested to confirm it in this situation.

spacepy/pycdf/__init__.py
jtniehof: Yep, if you want the module, `~spacepy.pycdf` should do the trick.

Doc/source/pybats.rst
jongeorge1999: Okay, makes sense to stick to conventions, I can do that.

spacepy/ctrans/__init__.py
jongeorge1999: Yes I got it down, grinding out the changes now

spacepy/pybats/__init__.py
jongeorge1999: I just saw a wicked long sectioned docstring in the file, and I added it to the additional user documentation. I can move it back if it makes more sense, but I have it currently under the IdlFile section of the pybats user doc

jtniehof: "Notes" is a legit numpydoc section heading, so it can stay in the docstring.

It was in the IdlFile class. As of the fetch I just did (and was the case when I mae this comment) it was in the docstring of `_read_idl_ascii`, which it's completely irrelevant to.

Doc/Makefile
jongeorge1999: This one was an easy fix, I just took out the "build"

Doc/source/coordinates.rst
jongeorge1999: This is just a reminder to myself to add nitpicky when this is the last comment left, I will commit the current module stuff now, and amend later with the conf.py edits.

Doc/source/coordinates.rst
jongeorge1999: Also an easy fix, this one is resolved

Doc/source/coordinates.rst
jongeorge1999: This one is resolved

Doc/source/coordinates.rst
jongeorge1999: This one is resolved

spacepy/coordinates.py
jongeorge1999: I just left it as :doc:`/coordinates` which worked fine, and I think simpler than adding reference labels, although possibly not as powerful.

Doc/source/datamodel.rst
jongeorge1999: This was resolved

spacepy/ctrans/__init__.py
jongeorge1999: I think I hit all of these, I think that this is resolved.

spacepy/pybats/rim.py
jongeorge1999: resolved

spacepy/pybats/rim.py
jongeorge1999: Resolved

spacepy/pybats/rim.py
jongeorge1999: resolved

Doc/source/time.rst
jongeorge1999: Resolved

spacepy/lib.py
jongeorge1999: Resolved

spacepy/lib.py
jongeorge1999: resolved

spacepy/poppy.py
jongeorge1999: resolved

spacepy/pybats/rim.py
jongeorge1999: resolved

spacepy/pybats/rim.py
jongeorge1999: resolved

spacepy/pybats/bats.py
jongeorge1999: resolved

spacepy/pycdf/__init__.py
jongeorge1999: I reverted this, I will get rid of the epydocs in a later PR, I don't want to lost focus of what this PR is actually for.

jongeorge1999: Ended up un-epydoccing this, it is resolved.

spacepy/pybats/__init__.py
jongeorge1999: Resolved

spacepy/pybats/__init__.py
jongeorge1999: Resolved

spacepy/pybats/__init__.py
jongeorge1999: resolved

spacepy/pycdf/__init__.py
jongeorge1999: Gotcha, It should indeed link to the user docs, will fix this now

spacepy/pybats/__init__.py
jongeorge1999: I think what happened was I pasted it in the wrong spot on accident, I fixed this now, so it should be resolved

spacepy/lib.py
```diff
@@ -119,11 +124,15 @@ def load_lib():
...
-    @type call_dict: dict
-    @param lib: library where functions specified in L{call_dict} live.
-    @type lib: ctypes.WinDLL or ctypes.CDLL
+    Parameters
+    ==========
+    call_dict : dict
+        call dictionary. Keyed by function name;
+        values are [return type, argtype0, argtype 1...]
+    
+    lib : ctypes.WinDLL or ctypes.CDLL
```
jtniehof, line 133: While we're fixing this, like I mentioned before, it's just ctypes.CDLL, no WinDLL anymore.

spacepy/pybats/rim.py
```diff
@@ -289,15 +271,11 @@ def calc_I(self):
         Iup:   The total upward current over one hemisphere.
         Idown: The total downward current over one hemisphere.
 
-        Values are stored in the object with the prefix 'n_' or 's_'
+        Values are stored in the object with the prefix ``'n_'`` or ``'s_'``
```
jtniehof, line 274: There's a bonus single quote here. Just double backticks, no single straight quotes.

spacepy/pybats/rim.py
```diff
@@ -387,7 +365,7 @@ def add_cont(self, var, target=None, n=50, maxz=False, lines=False,
             that is strictly positive and "Seismic" for diverging data.
             Alternatively, legacy Ridley Ionosphere Model color maps can be
             loaded using "l_wr" (white red) or "l_bwr" (blue-white-red),
-            where the "l_" prefix indicates legacy and not Matplotlib color maps.
+            where the ``"l_"`` prefix indicates legacy and not Matplotlib color maps.
```
jtniehof, line 368: Ugh, this gets nasty...can we just make them all monospace with no quotes for consistency here? Or given that we have the others up there, just leave the `l_` in normal type in double quotes, but escape the underscore? `"l\_"`

spacepy/pybats/rim.py
```diff
@@ -606,7 +584,7 @@ def add_oval_line(self, time, *args, **kwargs):
         not an existing axes object, a new axes object is created and customized
         to be an ionosphere polar plot.
 
-        Extra arguments/kwargs are sent to :meth:`matplotlib.pyplot.plot`.
+        Extra arguments/kwargs are sent to ``matplotlib.pyplot.plot()``.
```
jtniehof, line 587: This shoud be a valid cross-reference. Maybe try without the `:meth:`. If it needs a specifier, I think it's a `:func:`.

spacepy/pycdf/__init__.py
```diff
@@ -5913,14 +5707,23 @@ def __len__(self):
...
+
+        Returns
+        -------
+        CDF variable type : int
+
+        Notes
+        -----
+        If changing types, old and new must be equivalent, see CDF
+        User's guide section 2.5.5 pg. 57
+
```
jtniehof, line 5726: Don't need a bonus blank line here.

jtniehof (commented):
Few things that still need cleanup. I also have rebased against current `main` since #769 is merged, so you should be good to go running on irbempy doc fixes.

spacepy/pycdf/__init__.py
jongeorge1999: Resolved

spacepy/lib.py
jtniehof: Poke on this one again :)

drsteve (commented):
A few thoughts as I look through this...
Overall, I think this has been pulled out really nicely and will make documentation maintenance much easier moving forward.

Some thoughts and questions which may, or may not, be in scope for the PR:
- At the bottom of `spacepy/Doc/build/html/autosummary/spacepy.irbempy.html#module-spacepy.irbempy` there's an entry for the irbemlib `.so` file, which is really messy. Can that be excluded (or otherwise handled better)?
- Was there a plan behind the order of items in the "Additional User Documentation" section in `index.rst`? I'd suggest either alphabetical or order of "most fundamental to SpacePy". 
- In the `Parameters` entry of every function/method, there's no space between the parameter name and the type. I know it's not new, but is it a style setting or a formatting fix?
- Is there a reason why some modules have had preamble pulled out for the "Additional User Documentation" section but not others? E.g., the OMNI preamble is short, but was separated. The PoPPy preamble is longer and has not.
- The items under "Additional User Documentation" are inconsistent in naming. "OMNI" has no additional text associated in the title, while "spacepy.time - ..." is the only module including spacepy in the title. Four entries have leading capitals, two do not.
- `spacepy/Doc/build/html/autosummary/spacepy.toolbox.html#module-spacepy.toolbox` doesn't have a divider above the final entry (the exception), where everything else is delimited by a divider.
- As previous, but for `spacepy/Doc/build/html/autosummary/spacepy.datamodel.html#module-spacepy.datamodel`
- I've never liked the phrasing at the top of the datamodel docs claiming that datamodel is "meant to mirror the functionality of the data model output from pycdf, though implemented slightly differently" (yes, it's been there a very long time and I know I could/should have changed it). The intent was always to mirror the concepts in the HDF5 datamodel, which is compatible with/similar to the way things are represented in pycdf (as noted at the top of the actual docs, rather than the preamble I'm complaining about).

I think this also highlights the many places we need to fix documentation style and content (e.g., all of the pybats submodules).

I also haven't spotted anything that's missing or broken, so call this a general concurrence with questions. I can approve after we decide whether all of my thoughts are out of scope.

jtniehof:
@jongeorge1999 go ahead and take a crack at this...

>     * At the bottom of `spacepy/Doc/build/html/autosummary/spacepy.irbempy.html#module-spacepy.irbempy` there's an entry for the irbemlib `.so` file, which is really messy. Can that be excluded (or otherwise handled better)?

I don't think we can exclude the `= repr` portion, which is indeed ugly. We could extend the actual docstring so it's a little more clear what it is. We could also make it private, and then it won't be rendered in the docs. My preference would be either make it private or beef up the docstring a bit (so it's explained it's a handle to the C library, etc.) It also looks like it might be possible to just remove the docstring, and then it won't render. Should we try that short term?

(While looking at this, I noticed `lib` isn't linked from the list in `index.rst` and that should be fixed).

>     * Was there a plan behind the order of items in the "Additional User Documentation" section in `index.rst`? I'd suggest either alphabetical or order of "most fundamental to SpacePy".

Arbitrary; let's go for alphabetical.

>     * In the `Parameters` entry of every function/method, there's no space between the parameter name and the type. I know it's not new, but is it a style setting or a formatting fix?

We're going to look at the CSS and layout.html once this is merged, plus lots of other docs stuff.

>     * Is there a reason why some modules have had preamble pulled out for the "Additional User Documentation" section but not others? E.g., the OMNI preamble is short, but was separated. The PoPPy preamble is longer and has not.

Primarily the decision was "does this break numpydoc", e.g. having arbitrary section headings. We can refine this. Long-term we want to split API vs. "how-to"; I was pointed at [Diataxis](https://diataxis.fr/) as one model. We can break more stuff out, based on whatever suggestion you like.

>     * The items under "Additional User Documentation" are inconsistent in naming. "OMNI" has no additional text associated in the title, while "spacepy.time - ..." is the only module including spacepy in the title. Four entries have leading capitals, two do not.

Yeah, these should be smoothed.

>     * `spacepy/Doc/build/html/autosummary/spacepy.toolbox.html#module-spacepy.toolbox` doesn't have a divider above the final entry (the exception), where everything else is delimited by a divider.

That's what the autodoc exception template gives us. We could explicitly put an `<hr>` into the module template so there's one line above all the exceptions, or edit the exception template to put an `<hr>` above each. Should we defer this to visual / CSS enhancements?

>     * As previous, but for `spacepy/Doc/build/html/autosummary/spacepy.datamodel.html#module-spacepy.datamodel`

Ditto.

>     * I've never liked the phrasing at the top of the datamodel docs claiming that datamodel is "meant to mirror the functionality of the data model output from pycdf, though implemented slightly differently" (yes, it's been there a very long time and I know I could/should have changed it). The intent was always to mirror the concepts in the HDF5 datamodel, which is compatible with/similar to the way things are represented in pycdf (as noted at the top of the actual docs, rather than the preamble I'm complaining about).

We can defer, or put modified wording in here, whatever words you like.

drsteve (approved):
I think this covers everything that's in easy scope. As notes to move forward, I think some of the items discussed but unaddressed are:

> In the `Parameters` entry of every function/method, there's no space between the parameter name and the type.
> _...look at the CSS and layout.html once this is merged, plus lots of other docs stuff._

> Some modules have had preamble pulled out for the "Additional User Documentation" section but not others? E.g., the OMNI preamble is short, but was separated. The PoPPy preamble is longer and has not.
> _Primarily the decision was "does this break numpydoc", e.g. having arbitrary section headings. We can refine this. Long-term we want to split API vs. "how-to"; I was pointed at [Diataxis](https://diataxis.fr/) as one model. We can break more stuff out, based on whatever suggestion you like._

We also need to expand the guide-style intros to all of these modules.

> The items under "Additional User Documentation" are inconsistent in capitalization and formality of naming.

> No divider above exceptions, where everything else is delimited by a divider.
> _That's what the autodoc exception template gives us. We could explicitly put an `<hr>` into the module template so there's one line above all the exceptions, or edit the exception template to put an `<hr>` above each. Should we defer this to visual / CSS enhancements?_

Docs updates should be much easier now the big structural changes have been made, so I'm going to merge and note that we still have a lot to do here.

PR #760, opened 2024-06-25
Continuous integration fix for users with scipy>=1.14.0

jongeorge1999:
Scipy 1.14 updated the integration.simps() function to integration.simpson(), this is a fix to maintain continuous integration with users who have scipy<1.14

- [x] (N/A) Any new modules, functions or classes have docstrings consistent with SpacePy style
- [x] (N/A) Added an entry to release notes if fixing a significant bug or providing a new feature
- [x] (N/A) New features and bug fixes should have unit tests
- [x] (N/A) Relevant issues are linked to (e.g. `See issue #` or `Closes #`)

### DIFF
--- a/spacepy/empiricals.py
+++ b/spacepy/empiricals.py
@@ -672,7 +672,7 @@ def omniFromDirectionalFlux(fluxarr, alphas, norm=True):
        fac = 2*np.pi
        denomina = 1
     alphrad = np.deg2rad(alphas)
-    numera = integ.simpson(fluxarr*np.sin(alphrad), alphrad)
+    numera = integ.simpson(y=fluxarr*np.sin(alphrad), x=alphrad)
     omniflux = fac*numera/denomina
     return omniflux
 
--- a/spacepy/empiricals.py
+++ b/spacepy/empiricals.py
@@ -17,6 +17,8 @@ import warnings
 from functools import partial
 import numpy as np
 import scipy.integrate as integ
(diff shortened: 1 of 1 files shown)


jtniehof:
The relevance here is that 1.14 removed simps, so needed to change to the new name.

spacepy/empiricals.py
```diff
@@ -670,7 +672,7 @@ def omniFromDirectionalFlux(fluxarr, alphas, norm=True):
        fac = 2*np.pi
        denomina = 1
     alphrad = np.deg2rad(alphas)
-    numera = integ.simps(fluxarr*np.sin(alphrad), alphrad)
+    numera = integ.simpson(fluxarr*np.sin(alphrad), alphrad)
```
jtniehof, line 675: Looks like [x must be a kwarg](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.simpson.html) but [it's been named x even in simps](https://docs.scipy.org/doc/scipy-0.9.0/reference/generated/scipy.integrate.simps.html) so should be fine to just make it a kwarg here and it'll still work on the old code.

jtniehof (changes requested):
Whoops, doesn't even fix the problem right now :)

jtniehof:
Actually can you change the title, the fix is for >=1.14.0 (and check yer checkboxes)

jtniehof (approved):
Looks good.

PR #749, opened 2024-05-29
Remove f2py and use ctypes for irbem

jtniehof:
This PR replaces the f2py handling of irbem with building a shared library directly and using ctypes. Closes #725. Closes #679 because that relates to using numpy in the build, which is no longer the case. Closes #113 as irrelevant.

Instead of building irbem into an extension module with f2py, setup.py now builds irbem as a shared library. It's loaded using ctypes, like with libspacepy. This does mean the irbempy code has various typing bits scattered throughout now to prepare for the ctypes call into irbem. "irbempylib" (the exposed interface to the irbem extension module, imported as oplib) is now gone. Performance is comparable, and installation is about 15% faster.

Unit tests were added for the LCDS functions. These do take a little while to run (which is probably why they didn't exist yet). Expected values are taken from running the old code.

This removes f2py and numpy from the build process. The resulting wheels then don't depend on numpy or Python version, so we've updated the wheel metadata to indicate that. That also simplifies many build/test scripts since we don't need numpy in place first, or to always build against the oldest version. The CI has been updated to once again test with numpy >= 1.26 and python 3.12, which didn't build properly before.

In the process of simplifying the wheel build, we found it was fairly easy to build wheels for Raspberry Pi, so that's been added.

`irbempy/__init__.py` used to be basically an `import *` of `irbempy/irbempy.py`. The contents of `irbempy.py` were moved into `__init__.py`--among other things, this simplifies ongoing documentation work. However, that makes the diff on github unreadable; here is a [side-by-side diff](http://htmlpreview.github.io/?https://gist.githubusercontent.com/jtniehof/36753ef2cc493656e6b5bec0c6e8dd12/raw/192052a070f1d271e12c26943aa966fb26c317ed/irbem_diff.html).

TODO list is now complete:

- [x] Converting all oplib references
- [x] Checking other platforms (Windows, Mac x86, Mac ARM, Pi)
- [x] Verifying performance
- [x] Removing extension library
- [x] Verifying manylinux wheels and relevant library inclusions
- [x] Attempting to get wheels that are arch dependent but not Python version dependent
- [x] On manylinux wheel build, make dist directory explicitly
- [x] Many linux wheels: try (failed) to build CDF library only not all tools: unfortunately the CDF library makefiles make it difficult at best to build just the library

## PR Checklist

- [X] Pull request has descriptive title
- [X] Pull request gives overview of changes
- [X] New code has inline comments where necessary
- [X] Any new modules, functions or classes have docstrings consistent with SpacePy style
[...]

PR #767, opened 2024-07-09
Bump Pandas minimum version for Python 3.9 in CI and test_all.sh

### LINKED ISSUE #765: CI takes forever on Python 3.9, "old" dependencies.
The "old" dependencies strategy for Python 3.9 uses Pandas 0.24. This isn't available as a wheel for 3.9, so it builds from source. This first tries to build a wheel, which fails, and then for some reason setup.py install works (as a fallback). But that means the wheel isn't cached, so this whole build process runs every time the CI does. This takes ~14 minutes.

Probably the solution is to bump the earliest pandas version tested on 3.9. May be similar to numpy/numpy#17937 which describes a change to the Python API. Our tests for 3.8 use 0.24, and 3.10 use 1.2

jamesw732:
This pull request resolves #765, drastically improving the time it takes to install dependencies for Python 3.9 on the CI. The fix was to simply increase the Pandas version for the "oldest" dependency strategy to >=1.1.3, which is the earliest version of Pandas that supports binary wheels for Python 3.9 ([source](https://pandas.pydata.org/docs/whatsnew/v1.1.3.html)).

## PR Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,7 @@ jobs:
             os: ubuntu-22.04
             build-deps: "pip~=20.0.0 setuptools~=44.1.0 wheel~=0.34.2 cython<3.0"
             numpy-version: "~=1.18.0"
-            piplist: "python-dateutil~=2.5.0 scipy>=1.5.0,<1.6.0 matplotlib~=3.1.0 h5py~=2.10.0 astropy>=2.0,<2.1 pandas~=0.25.0"
+            piplist: "python-dateutil~=2.6.1 scipy>=1.5.0,<1.6.0 matplotlib~=3.1.0 h5py~=2.10.0 astropy>=2.0,<2.1 pandas~=0.25.0"
             cflags: "-Wno-error=format-security"
             cdf-version: "3.5.1"
             cdf-munged-version: "35_1"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,7 @@ jobs:
             os: ubuntu-22.04
             build-deps: "pip~=20.0.0 setuptools~=44.1.0 wheel~=0.34.2 cython<3.0"
             numpy-version: "~=1.18.0"
-            piplist: "python-dateutil~=2.5.0 scipy>=1.5.0,<1.6.0 matplotlib~=3.1.0 h5py~=2.10.0 astropy>=2.0,<2.1 pandas~=0.24.0"
+            piplist: "python-dateutil~=2.5.0 scipy>=1.5.0,<1.6.0 matplotlib~=3.1.0 h5py~=2.10.0 astropy>=2.0,<2.1 pandas~=0.25.0"
             cflags: "-Wno-error=format-security"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -87,7 +87,7 @@ jobs:
             os: ubuntu-22.04
             build-deps: "pip~=20.0.0 setuptools~=44.1.0 wheel~=0.34.2 cython<3.0"
             numpy-version: "~=1.18.0"
-            piplist: "python-dateutil~=2.5.0 scipy>=1.5.0,<1.6.0 matplotlib~=3.1.0 h5py~=2.10.0 astropy>=2.0,<2.1 pandas~=0.24.0"
+            piplist: "python-dateutil~=2.5.0 scipy>=1.5.0,<1.6.0 matplotlib~=3.1.0 h5py~=2.10.0 astropy>=2.0,<2.1 pandas>=1.1.3,<1.2.0"
```
jtniehof, line 90: This can just be `pandas~=1.1.3`

jtniehof (changes requested):
We were set up to build pandas from source, so we don't need to rely on a version with binary wheels. We just need one that builds cleanly from source, instead of doing the two-step of "try to build a wheel, fail, run setup.py install". Is there an earlier version than 1.1.3 where that's the case?

jamesw732:
> We were set up to build pandas from source, so we don't need to rely on a version with binary wheels. We just need one that builds cleanly from source, instead of doing the two-step of "try to build a wheel, fail, run setup.py install". Is there an earlier version than 1.1.3 where that's the case?

I see, I did not check that. The way I deduced 1.1.3 was by running `pip install pandas==ver` on an otherwise pretty blank conda python 3.9 environment for each version later than 0.24.0, and 1.1.3 was the first that successfully installed. I guess I should just commit/amend and push a bunch of different pandas versions to get the CI to run on each? That might be quicker than figuring out how to install old pandas versions from source locally.

jtniehof:
> I see, I did not check that. The way I deduced 1.1.3 was by running `pip install pandas==ver` on an otherwise pretty blank conda python 3.9 environment for each version later than 0.24.0, and 1.1.3 was the first that successfully installed. I guess I should just commit/amend and push a bunch of different pandas versions to get the CI to run on each? That might be quicker than figuring out how to install old pandas versions from source locally.

Nope, installing from source locally is pretty easy, just look at the CI. Start with a blank conda env, pip install "cython<3.0", and then go running along your pip install pandas versions. Three possible outcomes:

- build wheel fail, setup.py install works -- current state, bump minor version
- completely fails -- hmm, probably something wrong
- build wheel succeeds and installs --win!

It's pip doing the wheel build and then running setup.py when it fails, nothing you're doing.

jtniehof:
Pretty sure #749 will fix the 3.12 failure, since that relates to having to build an old numpy to work around f2py. We should probably just wait for that.

jtniehof (approved):
This only takes two minutes off the build directly, but the pandas wheel builds successfully. So next time the cache expires it should cache the pandas wheel and we'll get a huge speedup.

Issue #768, opened 2024-07-10
Refactor find_LCDS and find_LCDS_K


jtniehof:
Depends on #749.

find_LCDS and find_LCDS_K have several instances of a pattern:

1. unpack prep_irbem values
2. call find_magequator
3. filter bad output

This can probably be made a helper function, either a nested function for each or a separate one that's shared. (First eyeball glance indicates this section is the same between them).

drsteve:
Agreed, but I think this is more broad than just `find_magequator`. The same pattern is there for other irbem calls too...

As a side note, we should think about deprecating `find_LCDS` as it's basically an implementation of an incorrect method that had previous use in the literature. It was there basically to provide a way to illustrate the differences.

_Short explanation: L* is usually calculated as a function of pitch angle at a given location, but this is used to calculate K which is invariant around the drift shell. Different K at a given location will have a different L*. For an LCDS search you want to keep K invariant so that you're comparing like with like as you do your bisection to find the last valid L* (aka LCDS). The constant-alpha approach in `find_LCDS` means each search has a different K._

jtniehof:
Mostly just identifying low hanging fruit.

At a minimum we should updte that docstring to warn people off. Should we put in specific citations for where the inaccurate one was used? I can see preserving it for reproduction more or less indefinitely.

I can't seem to find my copy of Roederer but I presume that would be reasonable cite for a fuller explanation.

drsteve:
Most specific explanation for LCDS in particular is in Albert et al. (2018; https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2018JA025991). But yeah, Roederer is good for the underlying issues.

jamesw732:
I realized I didn't make #769 close this when it was supposed to. Deprecating `find_LCDS` seems to me like a separate issue, so I'll call this done and close it.

Issue #750, opened 2024-05-30
find_Bmirror ignores pitch angles


jamesw732:
`irbempy.find_Bmirror` ignores all elements of its argument `alpha` (pitch angles) aside from the first. As long as the first element is untouched, adding, removing, or changing elements of a given `alpha` produces the same output as the original `alpha`. Furthermore, this results in output arrays that are smaller than expected; all output sizes are as if `alpha` was a singleton list.

This appears to be caused by `alpha` being passed into `oplib.find_mirror_point1` as an array-like, when a float is expected. I'm not certain what happens to `alpha` afterwards, but I think it's implicitly casted to a float by taking the first element and truncating the rest.

**Edit**: Since changing irbempy to use ctypes instead of f2py, this bug was intentionally preserved by passing only the first element of `alpha` into `irbemlib.find_mirror_point1`.

### Minimal example to reproduce issue:
See unit tests added in [3a7498a](https://github.com/spacepy/spacepy/pull/749/commits/3a7498a5a52cac22c68c5bd769f75b9040017b9a) and [ae6e97d](https://github.com/spacepy/spacepy/pull/749/commits/ae6e97da84a469bbaed51ad809126890df470215).

### OS, Python version, and dependency version information:
```
Windows-10-10.0.22631-SP0
sys.version_info(major=3, minor=10, micro=14, releaselevel='final', serial=0)
numpy=1.26.4
scipy=1.12.0
matplotlib=3.8.4
```

### Version of SpacePy

Built/installed from source, commit [ae6e97d](https://github.com/spacepy/spacepy/pull/749/commits/ae6e97da84a469bbaed51ad809126890df470215).

PR #769, opened 2024-07-15
irbempy: Improve test coverage

### LINKED ISSUE #307: test coverage: irbempy
irbempy init is 100% coverage, but the real work in irbempy is 45% covered. 

- find_LCDS is untested, as is find_LCDS_K. I wonder if we could use lgmpy to determine "truth" and compare.
- get_AEP8 has a lot of exceptional cases not handled, but prints out. Maybe these should be warnings or exceptions.
- _get_Lstar when nAlpha == 0
- get_Lm --again, can use lgmpy, as that's pretty heavily tested for this.
- get_Lstar

Hard to test:
- updateTS07Coeffs downloads data and could probably be hit with an integration test

jamesw732:
This pull request improves the test coverage of the irbempy submodule, up to 88% from about 80%. Also, some minor bug fixes and code improvements along the way. Closes #307 and (slightly off-topic) #768.

The bulk of the improvements came from testing cases where the internal irbemlib call returned close to badval, and the further handling as np.nan. I also added some regression tests for untested edge cases and in some cases, complete functions (`find_Lm`). All the new tests are regression tests since I do not have a first principles understanding of the code.

What's still missing is some harder-to-hit edge cases and the hard-to-test header code.

Unfortunately, the edge cases in `find_LCDS` and `find_LCDS_K` mean those functions have to be run several times, causing`test_irbempy.py` to run very slowly - about 5 minutes on Windows, 87 seconds on WSL (you read that right). Given talks about deprecating `find_LCDS` in #768, it might be safe to just remove or skip its tests for the sake of improving the runtime of the tests.

Originally, I intended to include a fix for #750 in this PR but decided that it is too off-topic and might require too much targeted review, so it makes sense to just put that in another PR that depends on this one.

## PR Checklist

### DIFF
--- a/spacepy/irbempy/__init__.py
+++ b/spacepy/irbempy/__init__.py
@@ -420,9 +420,9 @@ def find_Bmirror(ticks, loci, alpha, extMag='T01STORM', options=[1, 0, 0, 0, 0],
 
         # take out all the odd 'bad values' and turn them into NaN
         if np.isclose(blocal, badval):
-            blocal = np.nan
+            blocal[()] = np.nan
         if np.isclose(bmirr, badval):
-            bmirr = np.nan
+            bmirr[()] = np.nan
         GEOcoord[np.where(np.isclose(GEOcoord, badval))] = np.nan
 
         results['Blocal'][i] = blocal
--- a/spacepy/irbempy/__init__.py
+++ b/spacepy/irbempy/__init__.py
@@ -1091,9 +1091,8 @@ def AlphaOfK(ticks, loci, K, extMag='T01STORM', options=[0, 0, 3, 0, 0], omnival
             # print('i= {0}, found alpha = {1}'.format(i, pa0))
             continue
 
-        Done, count = False, 0
         pa_upper, pa_lower, pa_test = pa0, 0, 30
-        while not Done:
+        for _ in range(21):
             # print('Testing PA={0}'.format(pa_test))
             # Now get K for initial alpha at this location...
             LS1 = get_Lstar(ticks[i], pos1, pa_test, extMag=extMag, options=options, omnivals=omnivals)
@@ -1116,9 +1115,6 @@ def AlphaOfK(ticks, loci, K, extMag='T01STORM', options=[0, 0, 3, 0, 0], omnival
                 # print('Reset upper bound. U,L = {0},{1}'.format(pa_upper, pa_lower))
             pa_test = pa_lower+np.abs((pa_upper-pa_lower))/2.0
             # print('Change alpha to {0}'.format(pa_test))
-            count += 1
-            if count > 20:
-                break
     return outvals
 
 
--- a/spacepy/irbempy/__init__.py
+++ b/spacepy/irbempy/__init__.py
@@ -775,9 +775,38 @@ def find_LCDS_K(ticks, K, extMag='T01STORM', options=[1, 1, 3, 0, 0], omnivals=N
     mlt = np.deg2rad(mlt)
 
     options = (int4 * 5)(*options)
+    nTtoG = 1.0e-5
 
+    pa = np.nan
     bmin = np.empty((), np.float64)
     GEOcoord = np.empty((3,), np.float64)
+
+    def inner_lcds_k_call(d):
+        badval = d['badval']
+        kext = d['kext']
+        sysaxes = d['sysaxes']
+        iyearsat = d['iyearsat']
+        idoysat = d['idoysat']
+        secs = d['utsat']
+        xin1 = d['xin1']
+        xin2 = d['xin2']
+        xin3 = d['xin3']
+        magin = d['magin']
+
+        irbemlib.find_magequator1(
+            int4(kext), options, int4(sysaxes), int4(iyearsat[0]),
+            int4(idoysat[0]), real8(secs[0]), real8(xin1[0]),
+            real8(xin2[0]), real8(xin3[0]),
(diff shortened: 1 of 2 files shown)


jamesw732:
Since the CI still fails on main and this is otherwise ready for review, I figure it doesn't hurt to mark it as so.

jtniehof:
Thanks James. Looks like the only problem is the known one fixed in #759. I hope to get a better look Monday, but you can also look at rebasing possibilities if there's some back-and-forth and blind alleys in history.

spacepy/irbempy/__init__.py
```diff
@@ -894,7 +850,7 @@ def find_LCDS_K(ticks, K, extMag='T01STORM', options=[1, 1, 3, 0, 0], omnivals=N
                 continue
 
             # print('L* at inner bracket: {0}'.format(LS1['Lstar']))
-            LCDS, LCDS_PA = LS1['Lstar'][0], pa
+            LCDS, LCDS_PA = LS1['Lstar'][0, 0], pa[0]
```
jtniehof, line 853: This feels inconsistent because pa is treated as a scalar above, in the assignment to results. Does AlphaOfK return a single value or an array of length 1? Should probably document it.

tests/test_irbempy.py
```diff
@@ -108,6 +108,20 @@ def test_prep_irbem(self):
             numpy.testing.assert_almost_equal(expected[key],
                                               actual[key],
                                               decimal=5)
+    def test_prep_irbem_sph_coords(self):
+        """Test prep_irbem with spherical coordinate input"""
+        loci = spacepy.coordinates.Coords([[3,0,0],[2,0,0]], 'GDZ', 'sph', use_irbem=True)
```
jtniehof, line 113: Can we try something with nontrivial angles here?

tests/test_irbempy.py
```diff
@@ -314,6 +396,52 @@ def test_find_footpoint(self):
...
+    def test_get_Bfield_badval(self):
+        """test get_Bfield hitting badval"""
+        loci_noncredible = spacepy.coordinates.Coords([[-1e4, 1e4, 0], [2, 0, 0]], 'GEO', 'car', use_irbem=True)
+        actual = ib.get_Bfield(self.ticks, loci_noncredible, omnivals=self.omnivals)
+        self.assertTrue(np.isnan(actual["Blocal"][0]))
+        self.assertTrue(np.isnan(actual["Bvec"][0]).all())
+
+    def test_find_Bmirror_badval(self):
+        """test find_Bmirror hitting badval"""
+        loci_noncredible = spacepy.coordinates.Coords([[-1e4, 1e4, 0], [2, 0, 0]], 'GEO', 'car', use_irbem=True)
```
jtniehof, line 412: Can we try this where the location is credible, but the pitch angle is in the loss cone? Here (if I'm reading this right) the PA 0 case is also the "way out in the Universe" case.

tests/test_irbempy.py
```diff
@@ -347,6 +475,19 @@ def test_coord_trans(self):
             [ 1.91462214,  0.06992421,  0.57387514]])
         numpy.testing.assert_almost_equal(expected, ib.coord_trans(self.loci, 'GSM', 'car'))
 
+    def test_coord_trans_swap(self):
+        """test coord_trans with invalid sysaxesout requiring a carsph swap"""
+        # re-assign ticks so this test can be run in isolation
+        self.loci.ticks = self.ticks
```
jtniehof, line 481: I'm not sure what this means. It's assigning ticks to the loci, which didn't have it before, but how is that isolating? (The setUp is run for every test, so one test shouldn't affect another)

jtniehof (commented):
Few suggestions on a quick skim

spacepy/irbempy/__init__.py
jamesw732: Aye it's always an array. This was causing a previously unhit numpy deprecation warning (and somehow an error in the CI... that was a frustrating one until I read more into the deprecation warning).

tests/test_irbempy.py
jamesw732: Yes, I'm kind of surprised I didn't notice that was missing from the prep_irbem test.

tests/test_irbempy.py
jamesw732: Yes, I think the reason I used a non-credible location was because I couldn't make Blocal hit badval otherwise. It seems appropriate to me to split this test up into two cases: noncredible location with a reasonable pitch angle, and credible location with zero pitch angle.

tests/test_irbempy.py
jamesw732: Somehow, I was not aware of that last part. I'll remove that comment.

spacepy/irbempy/__init__.py
jtniehof: But every place else pa is treated as a scalar: `results['AlphaEq'][idxt, idxk] = pa`. It's going to be `nan` if `GEOcoord[0]` isn't finite (which is a bit weird but I guess the assumption is that all three dimensions will be bad for a bad coordinate).

Do we treat this as "irbempy needs more work and this is out of scope for this PR"?

jtniehof (approved):
I'm going to merge this, but it leaves two open questions:
1. What further irbempy cleanup do we need to do, and do we need to make notes of all of them or just figure we'll find them back?
2. What should we put in release notes? find_Bmirror always returning the same locus sounds fairly severe but nobody's tripped over it yet...do we just say "several fixes for unusual cases in irbempy" or something? @drsteve ?

PR #770, opened 2024-07-19
irbempy: fix find_Bmirror ignoring pitch angles

### LINKED ISSUE #750: find_Bmirror ignores pitch angles

jamesw732:
Depends on #769.

This pull request closes #750. It fixes an issue with find_Bmirror where input pitch angles aside from the first are ignored completely, and handled as if there was only one.

Review will be required to verify the correctness of this. Initially, I imagined the fix would be to give each output a new 'pitch angle' dimension, however both 'loci' and 'Blocal' do not depend on pitch angle at all. 'Bmirr' is the only output field that was changed, and it is now in dimensions nTAI x nalpha. It is also unclear to me whether we'd want 'Bmirr' to always be 2-dimensional (even if alpha contains only one pitch angle), or if we'd want the number of dimensions to depend on whether there's more than one pitch angle.

## PR Checklist

### DIFF
--- a/Doc/source/release_notes.rst
+++ b/Doc/source/release_notes.rst
@@ -42,6 +42,11 @@ The ``spacepy.irbempy.irbempylib`` module has been removed. This was
 the old internal interface to the IRBEM library and was not intended
 for public use.
 
+Major bugfixes
+**************
+`~spacepy.irbempy.find_Bmirror` now correctly returns one ``Bmirr`` per input
+pitch angle instead of ignoring all but the first.
+
 Other changes
 *************
 Operations on `~spacepy.datamodel.dmarray` which return a scalar value
--- a/spacepy/irbempy/__init__.py
+++ b/spacepy/irbempy/__init__.py
@@ -365,6 +365,11 @@ def find_Bmirror(ticks, loci, alpha, extMag='T01STORM', options=[1, 0, 0, 0, 0],
     results : dictionary
         containing keys: Blocal, Bmirr, GEOcar
 
+    Notes
+    -----
+    The resulting Bmirr is a one-dimensional array in time if alpha is a singleton, otherwise
+    it is two-dimensional
+
     Examples
     --------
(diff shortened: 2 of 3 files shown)


jamesw732:
Is this fix significant enough to warrant a release notes entry?

jtniehof (changes requested):
Rebased to main after merging #769

I'd say in the interests of minimal change, keep the return 1D if there is only one pitch angle. So I think that means rebasing away your final commit.

This is probably worth a release note, and can you also update the docstring to indicate the dimensionality of the returned value?

jtniehof (approved):
Looks good; failure is because of fixed pending in #759

PR #559, opened 2021-10-08
Robust .spacepy directory handling

jtniehof:
This PR has a lot of "quality of life" improvements around handling of the `.spacepy` directory. Closes #282.

 * If the user specifies the `$SPACEPY` environment variable, all directories up to the specified one are created, assuming the user knows what they're asking for.
 * If the `.spacepy` directory exists but doesn't have the rc file or the data directory, those are made.
 * If the `spacepy.rc` file is empty, it is replaced with the defaults. This is similar to previous behavior, where if it was invalid for other reasons it would be replaced.
 * The `tai-utc` file is given permissions from user's umask, rather than the permissions in the repository. This is basically what would happen on `toolbox.update()` anyhow; they were different in the past due to differences between `copy` and `open`.
 * Refactoring to separate out these behaviors, and testing to verify them.

Because of #520, we already gracefully handle a missing `tai-utc.dat` file (#435). So I think with this PR, you can knock out just about any arbitrary piece of the `.spacepy` directory and things will still work.

One note: there's a pattern in here that looks like:

1. Check if something exists
2. Make it if it doesn't
3. Trap the error that happens if trying to make something that already exists

This is to handle a race condition if something else makes the directory or file between 1 and 2.

## PR Checklist

- [X] Pull request has descriptive title
- [X] Pull request gives overview of changes
- [X] New code has inline comments where necessary
- [X] Any new modules, functions or classes have docstrings consistent with SpacePy style
- [X] Added an entry to CHANGELOG if fixing a major bug or providing a major new feature
- [X] New features and bug fixes should have unit tests
- [X] Relevant issues are linked to (e.g. `See issue #` or `Closes #`)

PR #775, opened 2024-07-30
test_base: Improve test coverage

### LINKED ISSUE #300: test coverage: __init__.py
Our top-level init has 66% coverage. Some of this is hard to hit, but probably the best low-hanging fruit is working on more of the weird cases for the config file. Hitting the web browser with an integration test probably also wouldn't hurt.

jamesw732:
This pull request improves the test coverage of `spacepy/__init__.py`, closes #300. The coverage of this file has been improved since this issue was opened (in #376 and #559), up to 85%. This pull request improves this to 89%.

While writing tests, I noticed a bug in `_write_defaults` that produced erroneous output when `spacepy.rc` has more than one section. I rewrote the latter half of `_write_defaults` with a different approach to fix this. Thinking back on it now, I think a more minimal change was possible, although the new code is significantly shorter than the old code.

Also, `_read_config` contained a smaller bug, where if the initial generic configparser read produced an error, the existing `config` would always be used instead of the defaults. Then, when we try to cast `config` with `str2bool`, it fails because it's expecting the values to be strings, but they are not
This was fixed by moving `config = {}` into the `if not successful` block, as this step ensured that the new defaults would be written to `config`.

What's untested are `help`, some Python version/platform workarounds, and OSErrors at the bottom. The OSErrors seem to be avoiding race conditions, which I think are impossible to test, especially in Python.

## PR Checklist

### DIFF
--- a/spacepy/__init__.py
+++ b/spacepy/__init__.py
@@ -268,7 +268,7 @@ def _write_defaults(rcfile, defaults, section='spacepy'):
     try:
         startpos = f.tell()
         rclines = f.readlines()
-        writeme = [k for k in defaults.keys()]
+        writeme = list(defaults)
         #Places where sections start
         secidx = [i for i in range(len(rclines))
                   if re.match(r"^\[[^\[\]]*\]\n$", rclines[i])]
@@ -284,7 +284,6 @@ def _write_defaults(rcfile, defaults, section='spacepy'):
             else:
                 nextsec = None
             #Read only this section for lines matching default values
-            present = []
             for l in rclines[thissec:nextsec]:
                 #For each key, does the line match, commented or not?
                 for k in writeme:
@@ -292,23 +291,15 @@ def _write_defaults(rcfile, defaults, section='spacepy'):
                                 l):
                         writeme.remove(k)
                         break
-        f.seek(startpos) #Back to start of file
-        #Scan to just after section header. Text mode, can only seek() to tell()
-        if thissec != -1:
-            while f.readline() != '[{section}]\n'.format(section=section):
-                pass
-        if sys.platform == 'win32':
-            #Tickle the file for read/write switch
-            #https://stackoverflow.com/questions/11176724/python-file-operations
-            f.seek(0, 2)
-        #Write default values for anything not read
-        for k in sorted(writeme):
-            f.write(("#SpacePy {ver} default {key}: {value}\n"
-                     "#{key}: {value}\n").format(
-                         key=k, value=defaults[k], ver=__version__))
-        #And write all the remaining lines from the section header to end
-        if writeme:
-            f.writelines(rclines[thissec+1:])
+        #This is what we'll actually write to the file
+        writeme_verb = (f"#SpacePy {__version__} default {k}: {defaults[k]}\n"
+                        f"#{k}: {defaults[k]}\n"
+                        for k in sorted(writeme))
+        #Add writeme to end of this section
+        rclines[:nextsec] += writeme_verb
+        #Write the new contents
+        f.seek(0)
+        f.writelines(rclines)
     finally:
         f.close()
 
--- a/spacepy/__init__.py
+++ b/spacepy/__init__.py
@@ -266,7 +266,6 @@ def _write_defaults(rcfile, defaults, section='spacepy'):
     """Write configuration defaults out to a file if they aren't there"""
     f = open(rcfile, 'r+t') #Avoid race condition, open for read and write
     try:
-        startpos = f.tell()
         rclines = f.readlines()
         writeme = list(defaults)
         #Places where sections start
--- a/spacepy/__init__.py
+++ b/spacepy/__init__.py
@@ -291,11 +291,11 @@ def _write_defaults(rcfile, defaults, section='spacepy'):
                         writeme.remove(k)
                         break
         #This is what we'll actually write to the file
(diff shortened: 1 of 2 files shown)


jtniehof:
BTW need to fix the magic word, "closes".

I'm not thinking an integration test makes much sense for `help` now. We could get bizarre and mock up webbrowser.open to check the passed in URL, but in the end I think that would just be for test purity and not do anything good.

Because you can pass a filename into _write_defaults, it should be _realtively_ easy just to make a pseudo-config file in a tempdir, call _write_defaults, then read back in the result to check.

I think the things that I noted for testing were in _read_config, so all you need to do there is make a config file with garbage and call _read_config. You'll need to back up and restore config after the fact, though. (Can also make a config file with a valid format but no spacepy section).

Similarly you can play with _write_defaults by defining a config file with additional sections and that will trigger the seeking code.

This starts to sound like it's worth making a class just for these tests.

jamesw732:
For `_write_defaults` I'm able to create a config file with two sections and write to it, however the output is not as I expect, possibly because I don't understand the function very well. I'm essentially expecting the contents of the `defaults` argument to be inserted into `rcfile` starting immediately after `section`, but the following just adds all the defaults to the end of the file. Is this intended, or should they be inserted between the two sections like my intuition suggests?
```
import tempfile
import os
import shutil
import spacepy

td = tempfile.mkdtemp()
configfile = os.path.join(td, 'spacepy.rc')
try:
    with open(configfile, 'w') as cf:
        cf.write('[spacepy]\n')
        cf.write('[test section]\n')
    spacepy._write_defaults(configfile, {"test entry": "value"})
    with open(configfile, 'r') as cf:
        lines = cf.readlines()
        print(lines)
finally:
    shutil.rmtree(td)
```
```
['[spacepy]\n', '[test section]\n', '#SpacePy UNRELEASED default test entry: value\n', '#test entry: value\n']
```

jtniehof:
> For `_write_defaults` I'm able to create a config file with two sections and write to it, however the output is not as I expect, possibly because I don't understand the function very well.

Yeah, that looks like a bug. If the spacepy section isn't completely empty (say one line) does it work? That might be a hint at the problem.

jamesw732:
> If the spacepy section isn't completely empty (say one line) does it work?

This produces an interesting result. With input file looking like:
```
[spacepy]
some sample text

[test section]
```

I get
```
['[spacepy]\n', 'some sample text\n', '[test section]\n', '#SpacePy UNRELEASED default test entry: value\n', '#test entry: value\n', 'some sample text\n', '[test section]\n']
```
For some reason the section is repeated. This is some sort of lead, I guess.

jamesw732:
I had to mostly overhaul `_write_defaults`, I think the old approach was not very compatible with multiple sections due to how writing to python file streams works. My new approach was to just keep track of the lines in a list and do one big write to the file at the end. This might not be reasonable if the config files can be massive, but I don't believe that is the case for us.

spacepy/__init__.py
```diff
@@ -268,7 +268,7 @@ def _write_defaults(rcfile, defaults, section='spacepy'):
     try:
         startpos = f.tell()
```
jtniehof, line 269: This variable is no longer used since the whole file is rewritten verbatim.

spacepy/__init__.py
```diff
@@ -284,31 +284,22 @@ def _write_defaults(rcfile, defaults, section='spacepy'):
...
-            f.write(("#SpacePy {ver} default {key}: {value}\n"
-                     "#{key}: {value}\n").format(
-                         key=k, value=defaults[k], ver=__version__))
-        #And write all the remaining lines from the section header to end
-        if writeme:
-            f.writelines(rclines[thissec+1:])
+        #This is what we'll actually write to the file
+        writeme_verb = (f"#SpacePy {__version__} default {k}: {defaults[k]}\n"
+                        f"#{k}: {defaults[k]}\n"
+                        for k in sorted(writeme))
```
jtniehof, line 297: This had me confused for a bit; I think it would be clearer as a list comprehension instead of a generator expression, since ultimately it's added into a list.

spacepy/__init__.py
```diff
@@ -284,31 +284,22 @@ def _write_defaults(rcfile, defaults, section='spacepy'):
...
+        #Add writeme to end of this section
+        rclines[:nextsec] += writeme_verb
```
jtniehof, line 299: I am rather amazed this works. I think it's probably clearer to do a slice assignment:
```python
rclines[nextsec:nextsec] = writeme_verb
```

tests/test_base.py
```diff
@@ -152,6 +165,17 @@ def testDotflnRelative(self):
...
             os.chdir(wd)
 
+    def testDotflnTilde(self):
+        """Checks DOT_FLN with no available environment variables"""
+        if 'SPACEPY' in os.environ:
+            del os.environ['SPACEPY']
+        if 'HOME' in os.environ:
+            del os.environ['HOME']
+        if 'HOMEDRIVE' in os.environ:
+            del os.environ['HOMEDRIVE']
```
jtniehof, line 175: This makes me a little nervous about affecting other tests down the line. Can we save/restore these in a try/finally block?

jtniehof (changes requested):
Looks good; just a few little cleanup suggestions.

jamesw732:
The CI failures are from something unrelated, I assume a non-backwards-compatible dependency update. I'll re-request review.

jtniehof (approved):
Excellent!

PR #784, opened 2024-09-04
Remove Python 3.6 support

jtniehof:
The latest version of the GitHub runners, even on Ubuntu 20.04, [removes Python 3.6 support](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2004-Readme.md#python). This [broke CI](https://github.com/spacepy/spacepy/actions/runs/10697303743). Since we can no longer test it in CI, this PR removes SpacePy support for Python 3.6 and AstroPy 1.0.

Once 3.7 was our oldest version, the documentation CI was grabbing a version of numpydoc that would install but not work on 3.7. So this PR also establishes a minimum version of numpydoc (0.8.0).

Several of our actions were using old versions that had a deprecated node.js, so this PR also updates those runners (as long as we're messing with CI).

## PR Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,7 +139,7 @@ jobs:
             dep-strategy: "newest"
     steps:
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
     - name: Get Week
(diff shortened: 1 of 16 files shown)


jtniehof:
Now suddenly [the CI is working again](https://github.com/spacepy/spacepy/actions/runs/10716066172). Python 3.6 is still not listed in the runner.

actions/setup-python#544 is old but states "If you need Python 3.6, it’s still available on ubuntu-20.04".

Even if 3.6 is currently working, it's the only test we need to run on 20.04, and the days are clearly numbered. So we might as well proceed now.

drsteve (approved):

PR #790, opened 2024-10-14
Support for Python 3.13

jtniehof:
This PR makes updates for Python 3.13 support

- Changes the CI and build/test scripts to work against 3.13. There's some more general cleanup of the build/test scripts that was really enabled by #749 but it just got annoying not to do it here.
- Updates the CI badge in the readme per [new directions](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/monitoring-workflows/adding-a-workflow-status-badge). The old badge referring to the workflow name instead of filename was showing Passing despite the [CI failing](https://github.com/spacepy/spacepy/actions?query=workflow%3ACI) from Sphinx changes (fixed in this PR).
- Fix some invalid docstring formatting that only showed up once testing against the newer sphinx. This was actually breaking CI but wasn't apparent due to above.
- Fixes tests for the deprecation decorator test due to python/cpython#81283
- Fixes a numpy 1.25 deprecation in test_coordinates. This wasn't raising an error since it was associated with numpy.testing, not spacepy code.
- Works around the deprecation of pkg_resources in irbempy. irbempy uses importlib.resources on 3.9+, but 3.8 is still getting updates to setuptools which deprecated pkg_resources. The workaround tries to use the older version of importlib.resources where it can, only punting to pkg_resources where it fails (on zip / egg import). We picked up an undeclared runtime dependency on pkg_resources (setuptools) in 5a920b9af7bc0d0142f51bee8ad9a1ae523039e9.

## PR Checklist

- [X] (N/A?) Added an entry to release notes if fixing a significant bug or providing a new feature

No release note at this point since there were no changes to the spacepy code required to work with 3.13.


tests/test_coordinates.py
```diff
@@ -84,9 +84,9 @@ def test_GDZ_in_kilometers(self):
         got = cc_km.convert('GDZ', 'sph')
         expected = cc_Re.convert('GDZ', 'sph')
         # same valued output in km regardless of units of input
-        np.testing.assert_approx_equal(expected.radi, got.radi, significant=6)
-        np.testing.assert_approx_equal(expected.lati, got.lati, significant=6)
-        np.testing.assert_approx_equal(expected.long, got.long, significant=6)
+        np.testing.assert_almost_equal(expected.radi, got.radi, decimal=6)
```
drsteve, line 87: Just to note that the numpy docs discourage use of this function, which might be cause to look into switching comparison functions across the whole test suite (I don't think we're using their recommended functions anywhere)

drsteve (approved):
I'm good with all of this, noting that we may want to review how we do comparisons to a specified precision in future.

tests/test_coordinates.py
jtniehof: Yep, they also discourage `assert_approx_equal` so in some sense it's the closest equivalent and this is the smallest workable change :) I think in most cases that we're using it's okay, but worth considering a full scrub in the future.

PR #798, opened 2024-11-18
pycdf: support reading/setting last update of leapsecond table

### LINKED ISSUE #792: pycdf: access LeapSecondLastUpdated from CDF GDR?
Is it possible to return the LeapSecondLastUpdated value for a CDF using pycdf? i.e., the value from section 2.3 of the IDF (https://spdf.gsfc.nasa.gov/pub/software/cdf/doc/cdf390/cdf39ifd.pdf)

I don't see it in the pycdf docs, couldn't find any code for it when searching the repo.

jtniehof:
CDF ships with a leapsecond table and, when a CDF is created, writes the date of the final leapsecond in that table into the file. This PR adds support for reading and writing that value.

Closes #792

Test with:
```sh
pip install git+https://github.com/spacepy/spacepy.git@refs/pull/798/merge
```

## PR Checklist


jtniehof:
There's a _leapsgood function in spacepy.time which might be useful for figuring out the implications of an old leapsecond table. We can promote that to public API if it would be useful.

spacepy/pycdf/const.py
```diff
@@ -19,7 +19,7 @@
 # Regeneration process:
 # Edit a copy of cdf.h
 # M-x replace-string (CDFstatus) with nothing
-# h2py cdf.h
+# h2py cdf.h (probably need to use Python 2 to capture the Longs)
```
drsteve, line 22: Does this update need "h2py.py", and if so, where is this coming from now. My understanding is that it was dropped from the CPython repo. (The Python 2 comment is definitely making me question this...)

spacepy/pycdf/const.py
```diff
@@ -262,7 +268,15 @@
 IS_A_NETCDF =  -2228
 TT2000_TIME_ERROR = -2229
 UNABLE_TO_PROCESS_CDF = -2230
-ZLIB_DECOMPRESSION_ERROR = - 2231
+ZLIB_COMPRESS_ERROR = -2231
```
drsteve, line 271: `ZLIB_DECOMPRESSION_ERROR` has been removed, but had the same value that's now assigned to `ZLIB_COMPRESS_ERROR`. What version range of CDF is the new value valid for, and could this cause any misreporting in anything we support?

jtniehof: Ultimately we don't parse the error codes, they get passed to the CDF library's parser. At some point they changed from ZLIB_DECOMPRESSION_ERROR to two codes, ZLIB_COMPRESS_ERROR and ZLIB_UNCOMPRESS_ERROR. This change is present in 3.5, which is our oldest supported version.

We don't use this value for anything. I suppose we could include both names (both set to -2231) in case someone is using the old one, but given our CDF version support, I feel safe just using the current name.

drsteve: I'm happy to have this as-is if it's from 3.5 since that's a decade old at this point.

drsteve (approved):

PR #802, opened 2025-01-16
Update to use IGRF14

drsteve:
This PR takes all the necessary steps to bring in IGRF-14. This will fully address #801 

IGRF is provided (as a minimal implementation) via SpacePy in the `igrf` module, for the SpacePy backend coordinate conversions.
IGRF is also part of the IRBEM library, used for the IRBEM backend coordinate conversions and access to IRBEM functionality like `spacepy.irbempy.get_Lm`

This PR:
- updates the IRBEM library
- adds IGRF14 as bundled IGRF coefficient file and sets as default
- updates SV (secular variation) tests impacted by the change in IGRF/DGRF coefficients
- removes obsolete and unused IRBEM `manual` directory 
- updates release notes to record the update to IGRF14
- adds documentation to help the user change the IGRF model used by the SpacePy backend, if desired

### PR Checklist
- [N/A] New code has inline comments where necessary


drsteve:
Note that IRBEM has changed how they bundle docs, and now use sphinx.
We have the old `manual` directory, which I don't think we use, but is also several docs revisions old and hasn't existed for a while.
Do we want to:
1. remove the `manual` directory and just not bundle the (unused) docs, or
2. update to the current IRBEM docs in that location

My inclination is to just remove the manual directory. That will save our future git history and we don't use the docs anyway. (Calling @jtniehof, @spacecataz for comment)

Doc/source/coordinates.rst
```diff
@@ -98,4 +98,16 @@ different accuracies.  For example, assuming this module has been
...
+    >>> spc.DEFAULTS.set_values(use_irbem=False, itol=5)
+
+Note also that the magnetospheric systems listed above require the
+specification of a centered dipole axis. This is specified using the
+first three coefficients of the IGRF, evaluated at the specified time
+(specifically, the g[1][0], g[1][1], and h[1][1] coefficients).
+For the SpacePy backend, this uses the bundled IGRF coefficients file
+in the standard format. The coefficients from a different version of
+the model can be used by placing them in the ``.spacepy`` directory
+with the name ``igrfcoeffs.txt``. This would be used, for example, if
```
jtniehof, line 110: Based on the code, isn't this the data subdirectory of the .spacepy directory? (And should the reference to the .spacepy directory link configuration.rst, where it's documented?)

jtniehof (changes requested):
I'll probably take a closer look later, generally looks good (the irbem diffs are always fun).

Do we want to keep shipping igrf13coeffs.txt and just ignore it, or should we remove old versions as we go? It's not like we're shipping 12 just so people can copy the file elsewhere.

Doc/source/coordinates.rst
drsteve: Yes, and yes. And you might need to remind me how to format the link in docs...

jtniehof:
```rst
the model can be used by placing them in the ``data`` subdirectory of the :doc:`.spacepy directory <configuration>`
```

jtniehof (changes requested):
Two documentation changes: release notes as below, and then also note this change in coordinates.rst, starting on line 62below the other changes:

```

    .. versionchanged:: 0.8.0

    The IGRF model for the CTrans-based SpacePy backend was updated to IGRF14;
    the IRBEM backend was also updated to a version using IGRF14.
```

Technically this is in igrf but our pattern has been to document this sort of thing in coordinates, which I think helps discourage direct use of igrf.

jtniehof (commented):
All good. I cleaned up the old igrf13 file.

PR #811, opened 2025-06-02
Changing some integers to 64 bit so numpy >=2.0 can handle large files.

### LINKED ISSUE #810: MagGridFile function fails with newer versions of NumPy
When trying to read larger mag grid files, the bats.MagGridFile function yields the following error:

```
  File "/home/user/numpytesting.py", line 5, in <module>
    mags = bats.MagGridFile('testgrid.outs')
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/py_env/lib/python3.12/site-packages/spacepy/pybats/bats.py", line 3135, in __init__
    super(MagGridFile, self).__init__(header=None, *args, **kwargs)
  File "/home/user/py_env/lib/python3.12/site-packages/spacepy/pybats/__init__.py", line 1074, in __init__
    self._scan_bin_frames()
  File "/home/user/py_env/lib/python3.12/site-packages/spacepy/pybats/__init__.py", line 1111, in _scan_bin_frames
    info = _scan_bin_header(f, self._endchar,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/py_env/lib/python3.12/site-packages/spacepy/pybats/__init__.py", line 649, in _scan_bin_header
    info['end'] = info['start'] + head_size +  \
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OverflowError: Python integer 2147739312 out of bounds for int32
```

While troubleshooting, I found that this error appears when using **NumPy 2.0.0** or newer, while 1.26 and 1.25 work fine. It occurs with file sizes of **2 GB** or larger.

I found this size by using a single frame and a short script to test it, adding frames until I encountered this error. I can provide both the script and the file if necessary to reproduce this problem. With my specific frame, it took 3821 frames to reach the correct size and produce the error.

spacecataz:
Closes #810 

Change a few integers that default to numpy int32 to int64 to prevent overflow errors when opening large (>2Gb) files with IdlFile objects in Pybats. 

## PR Checklist

- [ x] Pull request has descriptive title
- [x ] Pull request gives overview of changes
- [ x] New code has inline comments where necessary
- [n/a ] Any new modules, functions or classes have docstrings consistent with SpacePy style
- [ n/a] Added an entry to release notes if fixing a significant bug or providing a new feature
- [ n/a] New features and bug fixes should have unit tests

### DIFF
--- a/spacepy/pybats/__init__.py
+++ b/spacepy/pybats/__init__.py
@@ -624,10 +624,10 @@ def _scan_bin_header(f, endchar, inttype, floattype):
     # Stash relevant values into the "info" dict:
     vals = readarray(f, dtype=header_fields_dtype, inttype=inttype)[0]
     for v, x in zip(['iter', 'runtime', 'ndim', 'nparam', 'nvar'], vals):
-        info[v] = x
+        info[v] = np.int64(x)
 
     # Dimensionality may be negative to indicate non-uniform grid:
-    info['ndim'] = abs(info['ndim'])
+    info['ndim'] = np.abs(info['ndim'], dtype=np.int64)
 
     # Get gridsize:
     grid = dmarray(readarray(f, inttype, inttype))
@@ -1018,7 +1018,7 @@ class IdlFile(PbData):
     keep_case : bool
         If set to True, the case of variable names will be preserved.  If
         set to False, variable names will be set to all lower case.
-    
+
     Notes
     -----
     PyBats assumes little endian byte ordering because
@@ -1041,7 +1041,7 @@ class IdlFile(PbData):
         :meth:`~spacepy.pybats.bats.Bats2d.extract` and
         :class:`~spacepy.pybats.qotree.QTree` for processing
         adjacent cells. (ASCII data were never sorted.)
-        
+
     '''
 
     def __init__(self, filename, iframe=0, header='units',
--- a/spacepy/pybats/__init__.py
+++ b/spacepy/pybats/__init__.py
@@ -627,7 +627,7 @@ def _scan_bin_header(f, endchar, inttype, floattype):
         info[v] = x.item()
 
     # Dimensionality may be negative to indicate non-uniform grid:
-    info['ndim'] = np.abs(info['ndim'])
+    info['ndim'] = abs(info['ndim'])
 
     # Get gridsize:
     grid = dmarray(readarray(f, inttype, inttype))
--- a/spacepy/pybats/__init__.py
+++ b/spacepy/pybats/__init__.py
@@ -624,10 +624,10 @@ def _scan_bin_header(f, endchar, inttype, floattype):
     # Stash relevant values into the "info" dict:
     vals = readarray(f, dtype=header_fields_dtype, inttype=inttype)[0]
     for v, x in zip(['iter', 'runtime', 'ndim', 'nparam', 'nvar'], vals):
-        info[v] = np.int64(x)
+        info[v] = x.item()
 
     # Dimensionality may be negative to indicate non-uniform grid:
-    info['ndim'] = np.abs(info['ndim'], dtype=np.int64)
+    info['ndim'] = np.abs(info['ndim'])
 
     # Get gridsize:
     grid = dmarray(readarray(f, inttype, inttype))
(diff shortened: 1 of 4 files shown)


spacecataz:
Converting to draft while we work out a related test that is quick enough/robust enough to test this.

spacepy/pybats/__init__.py
```diff
@@ -624,10 +624,10 @@ def _scan_bin_header(f, endchar, inttype, floattype):
     # Stash relevant values into the "info" dict:
     vals = readarray(f, dtype=header_fields_dtype, inttype=inttype)[0]
     for v, x in zip(['iter', 'runtime', 'ndim', 'nparam', 'nvar'], vals):
-        info[v] = x
+        info[v] = np.int64(x)
```
jtniehof, line 627: This feels a bit weird to me because these don't _all_ have to be 64 bit. In fact, none of them are really bigger than 32 bit, it's the casting rules later on that bite us.

numpy 1.x, when mixing Python int and numpy int32, would cast everything to numpy int64. 2.x casts to int32, which I guess makes more sense? Bites us here, though.

With this setup, we're still mixing Python and numpy types in calculating `info['end']`, we're just doing it in a way that's closer to how it worked on 1.x and doesn't blow up.

An alternative here would be `info[v] = x.item()` which means we do all our calculations in Python integers. We could also do the cast down in the calculation of `info['end']`.

I'm not demanding or even really asking for a change here. But I do find it a little bit confusing.

spacepy/pybats/__init__.py
```diff
@@ -624,10 +624,10 @@ def _scan_bin_header(f, endchar, inttype, floattype):
...
 
     # Dimensionality may be negative to indicate non-uniform grid:
-    info['ndim'] = abs(info['ndim'])
+    info['ndim'] = np.abs(info['ndim'], dtype=np.int64)
```
jtniehof, line 630: This doesn't need the dtype cast...if it's an int64 coming in, it will be one coming out. In fact just plain `abs` works but making it explicitly numpy seems reasonable. Casting to int64 implies it isn't one coming in.

tests/test_pybats.py
```diff
@@ -24,6 +20,10 @@
 import spacepy.pybats.ram as ram
 import spacepy.pybats.gitm as gitm
 
+import matplotlib
+matplotlib.use('Agg')
+import matplotlib.pyplot as plt
+
```
jtniehof, line 26: I like moving the mpl imports after the standard lib per PEP8. Putting them between standard lib and pybats, instead of after pybats, would 1. probably be more compliant with PEP 8 ("local" imports last) and 2. ensure that we still get the Agg backend even if pybats imports pyplot.

jtniehof (changes requested):
Two piddly little changes and one big "let's think about". I don't want to belabor this too much...the test catches the problem and the fix fixes it.

spacepy/pybats/__init__.py
spacecataz: Using Python integers makes sense and should make the process more straightforward. Changed.

spacepy/pybats/__init__.py
spacecataz: Agreed; changed.

spacepy/pybats/__init__.py
```diff
@@ -624,10 +624,10 @@ def _scan_bin_header(f, endchar, inttype, floattype):
     # Stash relevant values into the "info" dict:
     vals = readarray(f, dtype=header_fields_dtype, inttype=inttype)[0]
     for v, x in zip(['iter', 'runtime', 'ndim', 'nparam', 'nvar'], vals):
-        info[v] = x
+        info[v] = x.item()
 
     # Dimensionality may be negative to indicate non-uniform grid:
-    info['ndim'] = abs(info['ndim'])
+    info['ndim'] = np.abs(info['ndim'])
```
jtniehof, line 630: If we're making everything Python integers, this should just stay abs, not np.abs. Sorry :(

jtniehof (changes requested):
One last nitpick....

tests/test_pybats.py
spacecataz: Totally agree with the reasoning here; change reverted.

The reason I moved it is because PEP8-compliant linters really don't like the call to `matplotlib.use()` followed by more imports. This is not a problem with current versions of MPL, as `plt.use` can be called whenever; but as we're supporting older MPL versions, I didn't want to remove/move that line yet. So I compromised to try to make my linter happy.

PR #815, opened 2025-08-01
seapy - Vectorized Operations, Window Extraction, Test Stats Output 

connordimarco:
These updates improves  seapy.py. I vectorized the operations to improve speed. I changed the way that the code detects non-contiguous timesteps, and provides user feedback to detect potential problem points. I added a get_windows() function to print each of the windows that go into the superposed epoch analysis. I update the statistics functions to output the actual test statistics intstead of just printing plots. I added associated unit tests to work with the changes - I THINK these make sense. Many lines look changed because the script was run through a formatter. 

I made these changes to fit my own needs and have been running analyses with these changes for my personal use. This is my first pull request ever so if anything is incorrect please let me know. 

### DIFF
--- a/spacepy/seapy.py
+++ b/spacepy/seapy.py
@@ -104,37 +104,50 @@ class SeaBase(object):
             # Process each irregularity, focusing on actual problematic points
             irregular_count = 0
             for idx in noncontig_indices:
-                # For each irregular time difference, determine which point is problematic
-                # We need to check both points involved in the irregular difference
+                # For each irregular time difference, the problem is between idx and idx+1
+                # We need to determine which of these two points is actually problematic
 
                 # First convert all to a consistent array view for analysis
                 time_array = np.array([t.timestamp() if isinstance(t, dt.datetime) else t
                                        for t in times])
 
-                # Calculate expected positions based on surrounding points
-                if idx > 0 and idx < len(time_array) - 2:
-                    # We can check both surrounding points
-                    expected_curr = time_array[idx-1] + common_diff
-                    expected_next = time_array[idx+2] - common_diff
+                # The irregular difference is between idx and idx+1
+                actual_diff = time_array[idx+1] - time_array[idx]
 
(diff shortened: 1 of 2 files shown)


jtniehof:
I made some small changes to make CI (hopefully) pass. New tests weren't catching the warning about the input time being non-contiguous. I assumed this was still supposed to be a warning, since stdout was being checked. One test also warned about the window size changing--that warning is over-sensitive, since it's a float comparison, but I figured the smallest change was catching the warning rather than updating that comparison.

jtniehof:
Ooof, I will have to rebase out the merge commit, but that's a task for later. Lemme see if I can kick the CI....

connordimarco:
All,
Sorry for taking so long with this; I have been pulled away on other projects. 

@jtniehof Thank you for the janitorial work; git is not my strong suit and am still trying to learn. 

I've pushed two commits addressing your feedback:

force=True keyword on .sea() (e13bf5b)
Added a force keyword argument to Sea.sea(). If results already exist and force is not passed, it emits a warning and returns without recomputing, preserving backward compatibility for scripts that call .sea() twice. Passing force=True clears previous results and reruns the analysis. Test added to verify the warn-and-noop behavior.

Cadence detection via unique-counts instead of median (ec43c30)
Replaced np.median(td) with a unique-counts approach (np.unique + argmax) to find the actual most common timestep (mode) rather than the median. Went with unique-counts rather than histogram since timestep distributions are nearly discrete (sharp spikes at the true cadence) and np.histogram(bins='auto') can pick degenerate bin widths when the IQR is near zero. 

@drsteve sorry about the formatting/functional mixing, that was my mistake from
the original commit. I can rewrite the history to split them, but it'll involve
a force-push over jtniehof's commit and several cherry-pick conflicts. Would
you prefer I attempt the rewrite, or would it be easier for you to do the
history edit yourself before merging? Happy either way.

Issue #817, opened 2025-08-07
GITM Reader is out of date


abukowski21:
The reader for GITM outputs is out of date. It's missing features & is sloooowwww

This can be updated easily, I think. We have code in GITMCode/PyITM that can be reused

I'm self-assigning this so there's record that I'm working on it.

---

Here's a list of some features I think we want. Feel free to suggest and/or edit these:

- Quality-of-life upgrades:
  - [ ] Faster
  -  [ ] Read select (or all) variables; this will improve speed too
  -  [ ] convert rad -> deg automatically. 
  - [ ] Convert m -> km too?
- [ ] Separate header & file reader: this will allow users to set a path & open multiple .bin files there, or only open certain times
- [ ] Ability to decode GITM outputs' variable names. Similar to what's in Kamodo: (`[N!U+!N]` -> `N+`/`N^+`/`N_p` or something). Can grab units from Kamodo too
- [ ] Functionality to print some run information from a header file. For example we can print the grid size & limits, variable names, time range, GITM version, and output cadence from an output directory. Then the entire run will not need to be read in
- [ ] Make sure this all works for [1D,] 2D and 3D files. Perhaps include small examples for test coverage? 
- [ ] Add examples of how to read & plot

PR #819, opened 2025-08-21
Fix/robustify GITM reads

abukowski21:
# Fixing GITM reads

addresses issue #817 

Just in time for the hackathon :-)

## Major changes

- We only read the requested variables (list of integers)
- Can read multiple output files (GITM makes one output file per timestep) in one call
  - Accepts list of files or path to glob
  - Time is a dmarray, data is a dmarray too
  - pre-allocate time when parsing header, data is allocated before starting to read
- Only read what's actually necessary/requested
  - Add in `seek` to skip to line which we care about (don't read sequentially)
  - Refactor reading header. We can get dimensions, variables, etc. just once
  - All variables are listed in `self.attrs["vars"]`, with their indices, but data isn't always read

## Other things

- Tests are not super thorough. But do run! and exist
- Some things are marked `#TODO` which are functional, but low-hanging-fruit to fix
   - Kamodo's gitm reader has dicts which remap variable names & add units 
   - There can be an entry point if we just want to look at the variables contained in a header. or not.
   - Can add a check for if multiple output types are provided before going into the reading, or try and fail (current behavior)

 
Open to feedback if we want to change how anything is done. For now though it all works. 

---

- [kinda] New features and bug fixes should have unit tests

### DIFF
--- a/spacepy/pybats/gitm.py
+++ b/spacepy/pybats/gitm.py
@@ -115,17 +115,21 @@ class GitmBin(PbData):
                     # time is placed after variable names, so we skip:
                     # 52(grid etc.) + ( 40[nvars] + 8 [head/foot] ) *nVarsTotal
                     file.seek(52 + 48 * self.attrs['nVarsTotal'])
-                    (yy,mm,dd,hh,mn,ss,ms,startstop) = struct.unpack(endChar+'llllllli',file.read(32))
-                    self.attrs['time'][iFile]=datetime.datetime(yy,mm,dd,hh,mn,ss,ms*1000)
+                    (yy, mm, dd, hh, mn, ss, ms, startstop) = struct.unpack(
+                        endChar+'llllllli', file.read(32))
+                    self.attrs['time'][iFile] = datetime.datetime(
+                        yy, mm, dd, hh, mn, ss, ms*1000)
+
                 isFirstTime = False
 
                 for varname, iVar in self.attrs['var_idxs'].items():
                     # Get to the right location in file
-                    file.seek(HeaderLength+iVar*iDataLength)
+                    file.seek(HeaderLength + iVar*iDataLength)
                     # Pull out the data
-                    s=struct.unpack(endChar+'l', file.read(4))[0]
+                    s = struct.unpack(endChar+'l', file.read(4))[0]
                     self[varname][iFile, ...] = \
-                        np.array(struct.unpack(endChar+'%id'%(nPtsTotal),file.read(s))
+                        np.array(struct.unpack(endChar + '%id'%(nPtsTotal),
+                                               file.read(s))
                                  ).reshape((self.attrs['nLon'],
                                             self.attrs['nLat'],
                                             self.attrs['nAlt']), order='F')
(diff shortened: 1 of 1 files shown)


spacepy/pybats/gitm.py
```diff
@@ -8,8 +8,10 @@
 # Global imports:
 import numpy as np
 import datetime as dt
+from struct import unpack
 from spacepy.pybats import PbData
 from spacepy.datamodel import dmarray
+import glob
```
jtniehof: We can probably take advantage of the time to clean up the imports, according to [PEP8](https://peps.python.org/pep-0008/#imports).

- Separate standard library / other libraries / internal
- We know it's global imports, no need to have the comment
- I like having them alphabetized within each section, easier to tell what's going on
- Personal preference: not a fan of from imports, but we should at least be _sort of_ consistent...the fact that glob and struct are different here makes me twitchy

spacepy/pybats/gitm.py
```diff
@@ -23,103 +25,212 @@ class GitmBin(PbData):
...
+        Examples:
+            (to read all variables):
+            
+            > data = gitm.GitmBin('path/to/file/3DALL_t123_000.bin')
+
+            (to read select variables):
+            
+            > data = gitm.GitmBin('path/to/file/3DALL_*.bin', varlist=[0,1,2,34])
+
+
```
jtniehof: Can we numpydoc-ify this? The contents are great, just needs to be formatted. [example](https://github.com/spacepy/spacepy/blob/a58579f79a8b78729a1209cc399f301524cd0b1c/spacepy/pybats/bats.py#L23)

spacepy/pybats/gitm.py
```diff
@@ -23,103 +25,212 @@ class GitmBin(PbData):
...
+        # read in all the data
+        self._readbin(varlist=varlist)
+        if degrees:
+            self.calc_deg()
 
     def __repr__(self):
         return 'GITM binary output file %s' % (self.attrs['file'])
 
-    def _read(self):
+    def _readbin(self, varlist: list=None):
```
jtniehof: `varlist: list[int]=None` maybe, to be explicit?

spacepy/pybats/gitm.py
```diff
@@ -23,103 +25,212 @@ class GitmBin(PbData):
...
+                    endChar='<'
+
+                # Only read time if _read_header has not (it reads the 1st time)
+                if not isFirstTime:
+                    # Extract time, stored as 7 consecutive integers.
+                    # time is placed after variable names, so we skip:
+                    # 52(grid etc.) + ( 40[nvars] + 8 [head/foot] ) *nVarsTotal
+                    file.seek(52 + 48 * self.attrs['nVarsTotal'])
+                    (yy,mm,dd,hh,mn,ss,ms,startstop)=unpack(endChar+'llllllli',file.read(32))
+                    self.attrs['time'][iFile]=dt.datetime(yy,mm,dd,hh,mn,ss,ms * 1000)
```
jtniehof: Using this as an example (relevant elsewhere) but [PEP-8 whitespace formatting](https://peps.python.org/pep-0008/#whitespace-in-expressions-and-statements) would be helpful here...

spacepy/pybats/gitm.py
```diff
@@ -23,103 +25,212 @@ class GitmBin(PbData):
...
+                    # Pull out the data
+                    s=unpack(endChar+'l', file.read(4))[0]
+                    self[varname][iFile, ...] = \
+                        np.array(unpack(endChar+'%id'%(nPtsTotal),file.read(s))
+                                 ).reshape((self.attrs['nLon'],
+                                            self.attrs['nLat'],
+                                            self.attrs['nAlt']), order='F')
+
+        # Finally we reduce dimensionality:
+        for v in self.attrs['var_idxs'].keys():
```
jtniehof: keys() is redundant here, iterating over a mapping iterates over keys.

spacepy/pybats/gitm.py
```diff
@@ -23,103 +25,212 @@ class GitmBin(PbData):
...
+        This opens up a single GITM output ans extracts some data from the header.
+        Will, by default, only work on the first file in self.attrs['files']
+        
+        This will let us pre-allocate the arrays to hold the data
+
+        Inputs
+        ------
+            varlist (list): list of indices of variables to read. Default=None, 
+                which reads all variables
+
```
jtniehof: Again, can we numpydoc this?

jtniehof (changes requested):
Tests are great! Comments are primarily regarding formatting and documentation -- we also need a release note update. I can take a crack at those if you'd like, so I can dig through the formatting and have you double-check that I've done justice.

jtniehof:
Close/reopen to trigger CI. (EDIT: because the workflow was waiting for maintainer approval for more than 30 days...mea culpa)

spacepy/pybats/gitm.py
abukowski21: Yup! Easy. Will take care of this

PR #831, opened 2026-02-12
Add PyHC Actions for PHEP 3 compliance and environment compatibility

sapols:
Hey guys, Shawn from PyHC here! 

Can I politely ask you to be guinea pigs for these two new GitHub [Actions](https://github.com/heliophysicsPy/pyhc-actions) I developed? 😁 I'm asking all the core packages first. The actions check for (1) compliance with [PHEP 3](https://github.com/heliophysicsPy/standards/blob/main/pheps/phep-0003.md) and (2) compatibility with the [PyHC Environment](https://github.com/heliophysicsPy/pyhc-docker-environment). The idea is that they'll pass ✅ if there are no problems, but error/warn when problems arise. 

**E.g., you'll get a red ❌ if:**
- A change to your package would introduce a dependency conflict in the PyHC Environment
- You do not adopt a new Python version or core Scientific Python package version within 6 months of release

**You'll get ignorable warnings ⚠️ if:**
- You support older Python versions or core Scientific Python packages beyond the required SPEC 0 support window
- You put an upper-bound constraint on a core Scientific Python package
- An optional dependency inside an [extra] would introduce a dependency conflict in the PyHC Environment

If this goes well, I'd love to live in a future where all PyHC packages use these actions. I think it'd help keep everyone on the same page and be a huge win for interoperability!

## Summary

This PR adds two GitHub Actions workflows from [heliophysicsPy/pyhc-actions](https://github.com/heliophysicsPy/pyhc-actions):

### 1. PHEP 3 Compliance Checker
Validates that the package meets [PHEP 3](https://github.com/heliophysicsPy/standards/blob/main/pheps/phep-0003.md) requirements:
- Python version support (36-month support window)
- Core Scientific Python package version support (24-month support window)
- New version adoption within 6 months of release
- Warnings on upper-bound constraints and exact version pins (core Scientific Python packages only)

### 2. PyHC Environment Compatibility Checker
Detects dependency conflicts with the [PyHC Environment](https://github.com/heliophysicsPy/pyhc-docker-environment) to ensure the package can be installed alongside other PyHC packages:
- Errors when conflicts are detected in base package
- Warns when conflicts are detected in optional dependency extras

## Trigger Schedule
Both workflows run on:
- Push/PR to main branch
- Quarterly schedule (Jan 1, Apr 1, Jul 1, Oct 1 at 3pm UTC)
- Manual trigger via workflow_dispatch

## Notes
- I tested these actions on a personal [fork of SpacePy](https://github.com/sapols/spacepy/pull/1) and [pyhc-core](https://github.com/heliophysicsPy/pyhc-core), and that went well, so now I'm rolling them out to all core PyHC packages as the first real-world tests.
- The checks should only take a few seconds each to run, largely thanks to `uv` doing most the heavy lifting.

## Feedback
Please share any feedback about how I've implemented these! I've tested them as extensively as I can, but let me know if you experience buggy behavior. I want these to be _helpful_, not cumbersome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,3 +80,8 @@ For acknowledging SpacePy, please provide the URL to our github repository: [git
 
 ## Changes
 Changes in the released version of SpacePy are provided in the [release notes](https://spacepy.github.io/release_notes.html). For changes since the latest release, see the [repository version](https://github.com/spacepy/spacepy/blob/main/Doc/source/release_notes.rst).
+
+## The Python in Heliophysics Community
+[![PyHC Actions](https://github.com/spacepy/spacepy/actions/workflows/pyhc-actions.yml/badge.svg)](https://github.com/spacepy/spacepy/actions/workflows/pyhc-actions.yml)
+
+SpacePy is a core package of the [Python in Heliophysics Community (PyHC)](https://heliopython.org/), "promoting and facilitating the use and development of Python for Heliophysics".


jtniehof:
SunPy PR so we can see each others' discussion: sunpy/sunpy#8510

sapols:
Thanks for the feedback @jtniehof! I always appreciate your input.

The status badges are a nice idea worth looking into. I've back-burnered the idea of PyHC status badges for years but this could be a good excuse to resurface the idea. 

Linking to the repo in a comment would be easy enough. So is adjusting the cron timing (FWIW, the current timing lines up precisely with when [automated PHEP 3 emails](https://github.com/heliophysicsPy/heliophysicsPy.github.io/blob/main/.github/workflows/phep3_email_reminder.yml) get sent).

And re: limiting permissions, I'm pretty sure I already did exactly that in [PlasmaPy's PR](https://github.com/PlasmaPy/PlasmaPy/pull/3211). The workflow file failed their CI checks until I added `permissions: {}` and `persist-credentials: false`. We could all just decide to add those.

It'd be sweet if we all agreed on a canonical "right" implementation of this workflow file, then I could paste it as an example in the [repos's README](https://github.com/heliophysicsPy/pyhc-actions) for others to copy into their PyHC projects.

It's late so I'm going to bed now, and I might not return to this tomorrow. But if you don't beat me to it I'll return early next week.

jtniehof:
> So that this doesn't get overlooked, should we add a comment or a reference to https://github.com/heliophysicsPy/pyhc-actions#for-maintainers-creating-a-new-release in `DISTRIBUTE` so that we remember to do both tags? I'm pretty sure we don't do a major version tag right now, just `release-M.m.s`.

I think that's for making a new release of the action, not of a project using it. Some PyHC projects aren't using semantic versioning @sapols, can you clarify?

PR #832, opened 2026-03-10
Fix crash in parse_filename_time for filenames containing _north

### LINKED ISSUE #816: File Name Issues when reading IdlFiles
Hello!

I've run into an issue recently where the IdlFile reader will fail at reading a .outs file if "_n" is present when it calls parse_filename_time. In my case, I had a maggrid file named "mag_grid_north_e20100404-010000_20100408-221400.outs", where the "_north" would cause the error:

```
  File "/Users/arigott/Research/SWMF_helpers/unpack_outs.py", line 73, in <module>
    temp = pybats.IdlFile(f)
           ^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/spacepy/pybats/__init__.py", line 1222, in __init__
    t_info = list(parse_filename_time(filename))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/spacepy/pybats/__init__.py", line 240, in parse_filename_time
    subname = re.search(r'_n((\d+\_?)+)', filename).groups()[0]
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'groups'
```
I would assume based on the error that this issue will not be exclusive to IdlFiles and just depends on what is calling parse_filename_time.

m4cd4r4:
Fixes #816.

The check `if '_n' in filename` matches any substring starting with `_n`, including `_north`. When the regex `r'_n((\d+_?)+)'` then fails to match (because there are no digits after `_n`), calling `.groups()` on the `None` result raises an `AttributeError`.

Fixed by storing the match result and checking it before accessing groups.

- [x ] Pull request has descriptive title
- [x ] Pull request gives overview of changes

### DIFF
--- a/tests/test_pybats.py
+++ b/tests/test_pybats.py
@@ -48,15 +48,16 @@ class TestParseFileTime(unittest.TestCase):
              'z=0_mhd_2_e20140410-000000-000_20140410-000300-000.outs',
              'z=0_mhd_2_n00001500_00001889.outs',
              'log_n000010.log',
-             'log_e20140410-000000.log'
+             'log_e20140410-000000.log',
+             'mag_grid_north_e20130924-232600.out',
              ]
     iters = [None, None, 31073, 249620, None, 1500,
-             None, [1500, 1889], 10, None]
-    times = [None, None, 870, 18000, None, None, None, None, None, None]
+             None, [1500, 1889], 10, None, None]
+    times = [None, None, 870, 18000, None, None, None, None, None, None, None]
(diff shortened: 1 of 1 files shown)


m4cd4r4:
Hi @spacecataz - I've added a test to \ that includes a filename with \ in it (\). It verifies the date is still correctly parsed and no crash occurs. Let me know if you'd like the test to be more explicit about the failure case (e.g. wrapping with \ on the pre-fix code path).

